Issue #6, opened 2022-11-24
Consistency for Timestamp name


theoengland:
Timestamp name standard seems to be "time" (for instance when reading parameter data as a dataframe) but then, when I query that same data from influxdb it's called "Timestamp".

It's ok to be flexible to understand several names, but can we have the same one for sdk produced timestamps and influxdb?

PR #45, opened 2023-02-24
Docs impr, smaller renamings, raw topic fix, mem leak fix

peter-quix:
* Add version migration
* Fix RawTopic not using Get as the verb instead of Create
* Fix old naming left in new code
* DataFrame raised with 'timestamp' rather than 'time' (#6 )
* Fixed raw topic read
  * Fixed a typo (KafkaDataTime -> KafkaDateTime) in metadata
* Changed key to byte in transport layer and raw read/write
* Fixed some double dispose and memory leak


docs/versionmigration.md
```diff
@@ -0,0 +1,455 @@
...
+### OutputTopic and InputTopic classes renamed to TopicProducer and TopicConsumer
+
+This also brought several other changes to the code, see them below
+
+=== "Python before"
+
+    ``` python
+    output_topic = client.open_output_topic(TOPIC)
+    input_topic = client.open_input_topic(TOPIC)
+    # !!! Important change. input_topic used to default to a consumer group called 'Default' with Earliest offset.
```
JotaBlanco: is this good English? I didn't quite get it, but it may be me

docs/versionmigration.md
```diff
@@ -0,0 +1,455 @@
...
+    ``` python
+    stream.parameters.…
+    ```
+
+=== "Python after"
+
+    ``` python
+    stream.timeseries.…
+    ```
+
```
JotaBlanco, line 114: is this change true? I understand we are changing parameter data for time series, so where we had:
stream.parameters.write(df) now we do stream.timeseries.publish(df). But here the parameters I thought were related to the stream properties. ALso, the metadata is independent from timeseries or event data types

docs/versionmigration.md
```diff
@@ -0,0 +1,455 @@
...
+These are:
+- EventData: important to be disposed whenever manually created or received in callbacks
+- TimeseriesData: important to be disposed whenever manually created or received in callbacks
+- TimeseriesDataRaw: important to be disposed whenever manually created or received in callbacks
+- StreamPackage: important to be disposed whenever manually created or received in callbacks
+- StreamConsumer: also supports `dispose()` and automatically disposes when stream is closed.
+- StreamProducer: also supports `dispose()` and automatically disposes when stream is closed.
+- TopicConsumer: unless you're frequently subscribing to topics, this is not something you have to be too concerned about.
+- TopicProducer: unless you're frequently subscribing to topics, this is not something you have to be too concerned about.
```
JotaBlanco: these are not showing as a bullet point list:
image

src/CsharpClient/Quix.Streams.Streaming/QuixStreamingClient.cs
```diff
@@ -51,7 +51,7 @@ public class QuixStreamingClient
         /// <summary>
         /// The base API uri. Defaults to <c>https://portal-api.platform.quix.ai</c>, or environment variable <c>Quix__Portal__Api</c> if available.
         /// </summary>
-        public Uri ApiUrl = new Uri("https://portal-api.platform.quix.io");
+        public Uri ApiUrl = new Uri("https://portal-api.platform.quix.ai");
```
JotaBlanco, line 54: we should remember to change this when the platform domain change is done. Seems like the thing we would forget :S

JotaBlanco (dismissed):

docs/versionmigration.md
peter-quix: As discussed, what you're thinking of is `stream.properties` and that stays as is

docs/versionmigration.md
peter-quix: Probably just keyword overload.  Rephrased

docs/versionmigration.md
peter-quix: Checking. in readme looks fine, but not in generated docs.

src/CsharpClient/Quix.Streams.Streaming/QuixStreamingClient.cs
peter-quix: platform already injects it, only going to be a problem locally, but there will be an overlap where both will work.

docs/versionmigration.md
peter-quix: resolved
image

JotaBlanco (approved):

PR #46, opened 2023-02-27
Restructured readme and broke out optional info into supplementary topics

merlin-quix:
Updated the READMDE so that the libraries features are more prominent and understandable with code samples to illustrate how they work

Also split out lower-priority information into supplementary topics. For example:

- "What is time series data?" -> `what-is-timeseries.md` — many readers won't need this information so better to not let it clutter the main READMDE. 
- "Installing on M1/M2?" -> `mac-m1-m2-install.md` — again, this is currently "workaround" documentation (also very lengthy) and wont concern many readers (on Windows, Linux, or Intel Mac)
- "Architecture Notes" -> `arch-notes.md` — supplementary information that wont interest everyone.

Additionally, the "Prepare your Kafka Cluster" section did not provide any information on how to prepare anything, rather it's purpose seemed to be simply to communicate the message "We use Kafka".

I reworked this section to provide basic instructions for a local Kafka install to help people test the library quickly.


README.md
```diff
@@ -23,121 +23,229 @@ Quix Streams currently supports the following languages:
...
-```
+### Streaming contexts
+Streaming contexts allow you to bundle data from one data source into the same scope with supplementary metadata—thus enabling workloads to be horizontally scaled with multiple replicas.
 
-Time-series data is often plotted on a graph with the x-axis representing the time at which the data was collected and the y-axis representing the value of the data point.
+* In the following sample, the `create_stream` function is used to create a stream called _bus-123AAAV_ which gets assigned to one particular consumer and will receive messages in the correct order: 
 
-![Telemetry](./images/telemetry.png)
+    ```python
+    topic_producer = client.create_topic_producer("data")
```
tbedford: should this be `get_topic_producer`?

README.md
```diff
@@ -23,121 +23,229 @@ Quix Streams currently supports the following languages:
...
-### High-frequency data
+    stream = topic_producer.create_stream("bus-123AAAV")
+    # Message 1 sent (the stream context)
 
-High-frequency data is a type of <b>time-series data</b> that is collected at a very high rate, often at intervals of less than a second. Because high-frequency data is collected at such a high rate, it can be difficult to analyze and visualize using traditional methods. As a result, specialized techniques and tools are often used to process and analyze high-frequency data.
+    stream.properties.name = "BUS 123 AAAV"
+    # Message 2 sent (the human-readable identifier the bus)
 
-Quix streams is a library specialized in processing <b>high-frequency data</b>, although it can be used to process and analyze any kind of time-series or non-timeseries data.
+    stream.parameters \
```
tbedford: Should this be `stream.timeseries`?

I'm not sure how, but this looks like old code?

tbedford (changes requested):
Not reviewed all of it, but noticed the code in the README now seems to be for an old version of the SDK?

peter-quix:
> Not reviewed all of it, but noticed the code in the README now seems to be for an old version of the SDK?

Yeah, I think before we review this readme it would worth updating it to latest

tomas-quix:
Are you sure putting Getting started after library features?

merlin-quix:
> Are you sure putting Getting started after library features?

No - It doesn't have to be. I can reorder if you want. The "real" Quick Start is in the online docs anyway (which I am linking to)

README.md
merlin-quix: fixed these instances. Not familiar enough with all the library changes to catch everything though

README.md
merlin-quix: I don't know. I'll take your word for it though.

merlin-quix:
> > Not reviewed all of it, but noticed the code in the README now seems to be for an old version of the SDK?
> 
> Yeah, I think before we review this readme it would worth updating it to latest

Would be good to review it next to the [announcement copy](https://quix.io/blog/open-sourcing-quix-streams/?x-craft-preview=8jrofzmWCY&token=YYPO-EV_WzWuW3Erzm0aCXb5__a8ZAca) as both are designed to use the same feature lists w/ samples.

PatrickMiraP:
New arch-notes.md file is not correctly linked to the Readme. For instance where we are saying "Quix Streams is [designed to be extended](#interoperability-wrappers) to multiple programming languages.". 

Probably Contribution Guide section could be another good place to link to the new file.

SteveRosam:
Docs build including this branch is here:
https://quixdocsdev.blob.core.windows.net/pr91/sdk-intro.html

PatrickMiraP:
About removing "What is time series data" after "What is Quix Streams" section, there are some thoughts here in slack https://quixworld.slack.com/archives/C01U8DHH67J/p1677656326060989?thread_ts=1677218529.712249&cid=C01U8DHH67J

If we return it to the main readme, I will keep it in the same position. This section only makes sense to improve the context of "What is Quix Streams".

peter-quix (approved):

peter-quix (changes requested):
The Readme.md regressed in several places talking about output and inputs, etc.

peter-quix (approved):

PR #57, opened 2023-03-13
Improve performance of .net codebase

luisquix:
Optimizations:

- Paralellization of the protobufer's serialization.
- Paralellization of the DeserializingModifier's Publish method.
- Removed all lazy values.
- Updated protobuf to use the latest version so it works with Span.
- Updated all libraries.
- Updated to .NET 7.


peter-quix:
@luisquix created https://github.com/quixio/quix-streams/pull/75 based on this PR to lift the less error-prone improvements.

We can revisit the rest of the PR at a later date.

PR #58, opened 2023-03-13
[task/33358] - add data processing 101

tbedford:
## Description

This adds a new topic that provides an introduction to data processing with Quix Streams. The main objective is to cover two types of processing:

1. One message at a time
2. Stateful

The topic aims to identify issues with stateful processing around per-stream callback context, and replica instances. 

[Ticket](https://app.shortcut.com/quix/story/33358/data-processing-101)

## TODO

Still to do:

- [x] - Complete text descriptions
- [ ] - Add diagrams (a diagram has been created and will be improved by the design team, so I think this can be added later, when ready via a new PR)
- [x] - Add entry to TOC file (navbar entry) 

## Review

Notes for reviewers:

* Not sure if section "Tracking running totals across multiple streams" is correct.
* In the section "Handling system restarts and crashes" we saw this demonstrated in Quix Platform (where you had to enable "State Management" in the deployment dialog). Not sure how you would enable this if not using Platform.
* Not sure about topic name using vs data processing 101.

### Review app

See review app to be created in docs repo:

* [Page to review](https://quixdocsdev.blob.core.windows.net/pr95/sdk/using.html)

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -1,0 +1,250 @@
+# Using Quix Streams
+
+In this topic you will learn how to use Quix Streams to perform two types of data processing:
+
+1. **One message at a time processing** - Here the message received contains all required data. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically written to the output stream.
+2. **Stateful processing** - This is where you need to keep track of data between messages, such as keeping a running total of a variable. This is more complicated as state needs to be preserved between messages, and potentially between replicas where multiple replicas are deployed. In addition, state needs to be preserved in the event of the failure of a deployment - Quix Streams supports checkpointing as a way to enable this.
+
+The following sections will explore these methods of data processing in more detail.
+
+## Topics, streams, partitions, replicas, and consumer groups
+
+The main structure used for data organization in Quix is the topic. For example, the topic might be `iot-telemetry`. 
+
+
+Stream context: (also add link to stream context topic)
+You may have multiple devices, or sources, writing data into a topic, so to ensure scaling and message ordering, each source writes into its own stream. Device 1 would write to stream 1, and device 2 to stream 2 and so on.
+
+Quix Streams restricts all messages inside one stream to the same single partition. This means that inside one stream, a consumer can rely on the order of messages. A partition can contain multiple streams, but a stream is always confined to one partition.
+
+TBD: add info on replicas and consumer groups Necessary for horizontal scaling.
+
+Horizontal scaling occurs automatically, because when you deploy multiple replicas, a stream is assigned to a replica. For example, if there are three streams and three replicas, each replica will process a single stream. If you had only one replica, it would need to process all streams in that topic.
+
+## Stream data formats
+
+There are two main formats of stream data: event data and time-series data: 
+
+1. Event data, which in Quix Streams is represented with the `qx.EventData` class.
+2. Time-series, which in Quix Streams is represented with the `qx.TimeseriesData` class (and two other classes: one for Pandas data frame format, and one for raw Kafka data).
+
+Event data refers to data that is independent, whereas time-series data is a variable that changes over time. An example of event data is a financial transaction. It contains all data for the invoice, with a timestamp (the time of the transaction), but a financial transaction itself is not a variable you'd track over time. The invoice may itself contain time-series data though, such as the customer's account balance. 
+
+Time-series data is a variable that is tracked over time, such as temperature from a sensor, or the g-forces in a racing car.
+
+Time-series data has three formats in Quix Streams:
+
+1. Data (`qx.TimeseriesData`)
+2. Pandas Data Frame (`pd.DataFrame`)
+3. DataRaw (`qx.TimeseriesDataRaw`)
+
+In this topic you'll learn about the `TimeseriesData` and `pd.DataFrame` formats.
+
+## Registering a callback for stream data
+
+You can register a stream callback that is invoked when data is first received on a stream. 
+
+```python
+topic_consumer.on_stream_received = on_stream_received_handler
+```
+
+The `on_stream_received_handler` is typically written to handle a specific data format on that stream. This is explained in the next section.
+
+!!! note
+
+    This callback is invoked for each stream in a topic. This means you will have multiple instances of this callback running, if there are multiple streams. 
+    
+    If there are multiple replicas, and each replica is handling one stream, then each replica will run an instance of this callback. This has significance for how you are going to handle shared state, which is discussed later in this topic.
+
+## Registering callbacks to handle data formats
+
+Specific callbacks are registered to handle each type of stream data.
+
+The following table documents which callbacks to register, depending on the type of stream data you need to handle:
+
+| Stream data format | Callback to register |
+|----|----|
+| Event data | `stream_consumer.events.on_data_received = on_event_data_received_handler` |
+| Time-series data | `stream_consumer.timeseries.on_data_received = on_data_received_handler` |
+| Time-series raw data | `stream_consumer.timeseries.on_raw_received = on_raw_received_handler` |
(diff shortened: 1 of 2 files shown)


docs/using.md
```diff
@@ -0,0 +1,250 @@
+# Using Quix Streams
+
+In this topic you will learn how to use Quix Streams to perform two types of data processing:
+
+1. **One message at a time processing** - Here the message received contains all required data. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically written to the output stream.
```
SteveRosam: Use published instead of written
Maybe other instances

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+
+The following sections will explore these methods of data processing in more detail.
+
+## Topics, streams, partitions, replicas, and consumer groups
+
+The main structure used for data organization in Quix is the topic. For example, the topic might be `iot-telemetry`. To allow for horizontal scaling, a topic is typically divided into multiple streams. You may have multiple devices, or sources, writing data into a topic, so to ensure scaling and message ordering, each source writes into its own stream. Device 1 would write to stream 1, and device 2 to stream 2 and so on. This is the idea of [stream context](./features/streaming-context.md). 
+
+Quix Streams ensures that stream context is preserved, that is, messages inside one stream are always written to the same single partition. This means that inside one stream, a consumer can rely on the order of messages. A partition can contain multiple streams, but a stream is always confined to one partition.
+
+It is possible to group code for a topic using the idea of a consumer group. When you create the consumer you specify the group as follows:
```
SteveRosam: "It is possible to group code" - not sure if this is a typo or something is missing?
Maybe "It is possible to group messages"?

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+...
+```
+
+The key point here is that data is tracked per stream context. You keep running totals on a per-stream basis by using the stream ID, `stream_consumer.stream_id` to index a dictionary containing running totals for each stream.
+
+!!! note
+
+    A stream will only ever be processed by one replica.
+
+## Tracking running totals across multiple streams
```
SteveRosam, line 219: Yep. I like this section. The user could achieve this in multiple ways but using QuixStreams is a valid option.
We could write a `cache app` example implementation

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+    g_running_total_per_stream[stream_consumer.stream_id] += some_value
+    data.add_value("RunningTotal", g_running_total_per_stream[stream_consumer.stream_id])
+
+    topic_producer.get_or_create_stream(stream_consumer.stream_id).timeseries.publish(data)
+...
+```
+
+In this case the running total is published to its own stream in the output topic. The next service in the data processing pipeline would be able to sum all running totals across all streams in the output topic.
+
+## Handling system restarts and crashes
```
SteveRosam, line 245: I would add some explanation around the fact that any in-memory data will be lost on service crash or shutdown. This can be mitigated by using LocalFileStorage etc.

SteveRosam, line 245: In reply to your comment on the PR "Not sure how you would enable this if not using Platform"..

You don't need to enable it. It just creates a file with whatever data you push to storage.

Run a sample on your local machine and you'll see a file appear in a folder in the solutions folder.

Enabling state management on a deployment in the SaaS Platform says it enables state that is shared between replicas.
However, I don't think you have to enable it to use state management (for a single replica)

tbedford: Thanks, good to know!

docs/using.md
tbedford: Yes, thanks, that was badly worded. I've tried to clarify. I may also update again based on new information.

docs/using.md
tbedford: Hopefully I've now clarified this section a bit.

docs/using.md
tbedford: Fixed.

SteveRosam (approved):

docs/using.md
```diff
@@ -0,0 +1,282 @@
+# Using Quix Streams
+
+In this topic you will learn how to use Quix Streams to perform two types of data processing:
+
+1. **One message at a time processing** - Here the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically published to the output stream.
```
tomas-quix, line 5: Stateless processing is a more frequent term. I would use that in bold and then mention "One message at a time processing" in the description.

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+2. **Stateful processing** - This is where you need to keep track of data between messages, such as keeping a running total of a variable. This is more complicated as state needs to be preserved between messages, and potentially between replicas, where multiple replicas are deployed. In addition, state may need to be preserved in the event of the failure of a deployment - Quix Streams supports checkpointing as a way to enable this.
+
+The following sections will explore these methods of data processing in more detail.
+
+## Topics, streams, partitions, replicas, and consumer groups
+
+The main structure used for data organization in Quix is the topic. For example, the topic might be `iot-telemetry`. To allow for horizontal scaling, a topic is typically divided into multiple streams. You may have multiple devices, or sources, writing data into a topic, so to ensure scaling and message ordering, each source writes into its own stream. Device 1 would write to stream 1, and device 2 to stream 2 and so on. This is the idea of [stream context](./features/streaming-context.md).
```
tomas-quix, line 12: I would add the following paragraph: If you wish to aggregate data beyond the stream context of a source, slice streams in the pipeline based on a different key. For example, invoices are being streamed from stores where each store has one stream. We want to aggregate items per StockCode to see how much of each article was sold. We need to first split invoices into streams per StockCode and then perform aggregation.

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+There are two main formats of stream data: 
+
+1. **Event data** - in Quix Streams this is represented with the `qx.EventData` class.
+2. **Time-series** - in Quix Streams this is represented with the `qx.TimeseriesData` class (and two other classes: one for Pandas data frame format, and one for raw Kafka data).
+
+Event data refers to data that is independent, whereas time-series data is a variable that changes over time. An example of event data is a financial transaction. It contains all data for the invoice, with a timestamp (the time of the transaction), but a financial transaction itself is not a variable you'd track over time. The invoice may itself contain time-series data though, such as the customer's account balance. 
+
+Time-series data is a variable that is tracked over time, such as temperature from a sensor, or the g-forces in a racing car.
+
+Time-series data has three formats in Quix Streams:
```
tomas-quix, line 41: Time-series data has three different representations in Quix Streams to serve different use cases and developers. The underlying data that these 3 models represent are the same.

docs/using.md
```diff
@@ -0,0 +1,282 @@
...
+
+1. Data (represented by the `qx.TimeseriesData` class)
+2. Pandas Data Frame (represented by the `pd.DataFrame` class)
+3. DataRaw (represented by the `qx.TimeseriesDataRaw` class)
+
+In this topic you'll learn about the `TimeseriesData` and `pd.DataFrame` formats.
+
+## Registering a callback for stream data
+
+You can register a stream callback that is invoked when data is first received on a stream.
```
tomas-quix, line 51: The first step is to register....

tbedford: Fixed.

tbedford:
This PR was closed before the review feedback of @tomas-quix was added. I am going to raise a new PR specifically for this feedback. The review app will be the same URL.

docs/using.md
tbedford: Fixed.

docs/using.md
tbedford: I want to make sure I completely understand this before adding the paragraph.

docs/using.md
tbedford: Fixed.

PR #69, opened 2023-03-20
[Task/33358] - additions to 'using Quix Streams' topic

tbedford:
## Description

* Changes nav bar entry to 'Using Quix Streams'
* Adds paragraph on cross-stream aggregation
* Adds warnings for code that demonstrates potential pitfalls

## Review

[Page to review](https://quixdocsdev.blob.core.windows.net/pr104/client-library/using.html)

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -2,7 +2,7 @@
 
 In the following documentation you will learn how to use Quix Streams to perform two types of data processing:
 
-1. **Stateless processing** - Here one message is processed at a time, and the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically published to the output stream. 
+1. **Stateless processing** - Here one message is processed at a time, and the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate new information, which is then published to the output stream. 
 2. **Stateful processing** - This is where you need to keep track of data between messages, such as keeping a running total of a variable. This is more complicated as state needs to be preserved between messages, and potentially between replicas, where multiple replicas are deployed. In addition, state may need to be preserved in the event of the failure of a deployment - Quix Streams supports checkpointing as a way to enable this.
 
 The following sections will explore these methods of data processing in more detail.
--- a/docs/using.md
+++ b/docs/using.md
@@ -1,6 +1,6 @@
 # Using Quix Streams
 
-In this topic you will learn how to use Quix Streams to perform two types of data processing:
+In the following documentation you will learn how to use Quix Streams to perform two types of data processing:
 
 1. **Stateless processing** - Here one message is processed at a time, and the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically published to the output stream. 
 2. **Stateful processing** - This is where you need to keep track of data between messages, such as keeping a running total of a variable. This is more complicated as state needs to be preserved between messages, and potentially between replicas, where multiple replicas are deployed. In addition, state may need to be preserved in the event of the failure of a deployment - Quix Streams supports checkpointing as a way to enable this.
@@ -15,9 +15,7 @@ In some use cases you may want to aggregate data beyond the stream context of a 
 
 Quix Streams ensures that stream context is preserved, that is, messages inside one stream are always published to the same single partition. This means that inside one stream, a consumer can rely on the order of messages. A partition can contain multiple streams, but a stream is always confined to one partition.
 
-It is possible to organize the code that processes the streams in a topic using the idea of a consumer group. This indicates to the broker that you will process the topic with all available replicas.
-
-Horizontal scaling occurs automatically, because when you deploy multiple replicas, a stream is assigned to a replica. For example, if there are three streams and three replicas, each replica will process a single stream. If you had only one replica, it would need to process all streams in that topic. If you have three streams and two replicas, one replica would process two streams, and the other replica a single stream.
+It is possible to organize the code that processes the streams in a topic using the idea of a consumer group. This indicates to the broker that you will process the topic with all available replicas in the consumer group, sharing the processing of all streams in the topic. Horizontal scaling occurs automatically, because when you deploy multiple replicas in a consumer group, a stream (or group of streams) is assigned to a replica. For example, if there are three streams and three replicas, each replica will process a single stream. If you had only one replica, it would need to process all streams in that topic. If you have three streams and two replicas, one replica would process two streams, and the other replica a single stream. If you don't specify a consumer group in your code, then all streams in a topic will be processed by all replicas.
 
 When you create the consumer you specify the consumer group as follows:
 
@@ -294,7 +292,7 @@ Also, in this example, the running total is persisted in file storage, and so is
 
 ## Conclusion
 
-In this topic you have learned:
+In this documentation you have learned:
 
 * How to perform stateless "one message at a time" processing.
 * How to handle the situation where state needs to be preserved, and problems that can arise in naive code.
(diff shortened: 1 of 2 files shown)


tbedford:
@tomas-quix - would you be able to confirm I made all the changes you required please? Thanks!

docs/using.md
```diff
@@ -210,12 +219,37 @@ def callback_handler (stream_consumer: qx.StreamConsumer, data: qx.TimeseriesDat
...
+
+This ensures that the variable `my_var` is persisted, as periodically (default is 20 seconds) it is flushed to local file storage.
+
+If the system crashes (or is restarted), Kafka resumes message processing from the last committed message. This facility is built into Kafka.
+
+!!! tip
+
+    For this facility to work in Quix Platform you need to enable the State Management feature. You can enable it in the `Deployment` dialog, where you can also specify the size of storage required. When using Quix Streams with a third-party broker such as Kafka, no configuration is required, and data is stored on the local file system.
+
 ## Tracking running totals across multiple streams
```
tomas-quix, line 265: This should still clearly state it is describing the problem.

docs/using.md
```diff
@@ -210,12 +219,37 @@ def callback_handler (stream_consumer: qx.StreamConsumer, data: qx.TimeseriesDat
...
 
     A stream will only ever be processed by one replica.
 
+## Handling system restarts and crashes
+
+One issue you may run into is that in-memory data is not persisted across instance restarts, shutdowns, and instance crashes. This can be mitigated by using the Quix Streams `LocalFileStorage` facility. This will ensure that specified variables are persisted on permanent storage, and this data is preserved across restarts, shutdowns, and system crashes.
+
+The following example code demonstrates a simple use of `LocalFile Storage`:
+
+```python
```
tomas-quix, line 235: Can you here provide a full code sample as this is actually what they looking for? Also, can you make sure it is very visible this is the target code.

tbedford: I've extended the example.

docs/using.md
```diff
@@ -2,7 +2,7 @@
 
 In this topic you will learn how to use Quix Streams to perform two types of data processing:
```
JotaBlanco: I would ban the word "topic" from our vocabulary in docs/blogs/conferences when not referring to Kafka Topics, to avoid any chance of confusion. Maybe use "sections" instead?

docs/using.md
```diff
@@ -2,7 +2,7 @@
...
 
-1. **One message at a time processing** - Here the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically published to the output stream.
+1. **Stateless processing** - Here one message is processed at a time, and the message received contains all required data for processing. No state needs to be preserved between messages, or between replicas. The data from the message is used to calculate a new value, which is then typically published to the output stream.
```
JotaBlanco: The data from the message is used to calculate a new feature, typically to enrich the message, which is then published to the output stream.

docs/using.md
```diff
@@ -11,6 +11,8 @@ The following sections will explore these methods of data processing in more det
...
 
+In some use cases you may want to aggregate data beyond the stream context of a source. This can be done by writing data to new streams in the pipeline, based on a different key. For example, imagine a scenario where invoices are being streamed from stores, and where each stream is based on `StoreId`. Now, let's say you want to calculate totals of a specific item sold across all stores (streams). To do this, you can create a transform that first writes invoices into new streams based on `StockCode`, and then another transform can perform aggregation for each of these `StockCode` streams, in order to calculate how much of each item was sold.
+
 Quix Streams ensures that stream context is preserved, that is, messages inside one stream are always published to the same single partition. This means that inside one stream, a consumer can rely on the order of messages. A partition can contain multiple streams, but a stream is always confined to one partition.
 
 It is possible to organize the code that processes the streams in a topic using the idea of a consumer group. This indicates to the broker that you will process the topic with all available replicas.
```
JotaBlanco: Not sure about this way of explaining what a consumer group is :S

docs/using.md
```diff
@@ -23,10 +25,14 @@ When you create the consumer you specify the consumer group as follows:
 topic_consumer = client.get_topic_consumer(os.environ["input"], consumer_group = "empty-transformation")
 ```
 
-!!! note
+Best practice is to make sure the consumer group name matches the name of the service.
+
+!!! warning
 
     If you don't specify a consumer group, then all messages in all streams in a topic will be processed by all replicas in the microservice deployment.
```
JotaBlanco, line 30: is this right? if I have two replicas listening to two streams, don't they distribute the load accordingly?

docs/using.md
```diff
@@ -234,41 +246,57 @@ def callback_handler (stream_consumer: qx.StreamConsumer, data: qx.TimeseriesDat
...
+```
+
+In this case the running total is published to its own stream in the output topic. The next service in the data processing pipeline would be able to sum all running totals across all streams in the output topic.
 
-    For this facility to work in Quix Platform you need to enable the State Management feature. You can enable it in the `Deployment` dialog, where you can also specify the size of storage required. When using Quix Streams with a third-party broker such as Kafka, no configuration is required, and data is stored on the local file system.
+Also, in this example, the running total is persisted in file storage, and so is preserved in the event of service restarts and system crashes.
 
 ## Conclusion
 
 In this topic you have learned:
```
JotaBlanco: "topic" word again

JotaBlanco (commented):
I really liked it. I've actually learn the flush method once for all reading this! I've added some suggestions though

docs/using.md
tbedford: That's a good point. We use topic-based writing, which is why that term is used, but it could be confused with Kafka topics. I will change to something like "In this documentation..." or "In the following sections..."

docs/using.md
tbedford: Is "feature" the right word here? Perhaps "variable" or "information"? WDYT?

docs/using.md
tbedford: That was based on [this conversation](https://quixworld.slack.com/archives/C04KXLCGQ94/p1678805872511799). However, I've reworded and put in some additional explanation.

docs/using.md
tbedford: No, not if you've not specified a consumer group, because in your case both streams will be processed by both replicas. This is basically why consumer groups exist, to allow you to specify horizontal scaling. As was pointed out in the conversation linked below, for some use cases you do want all replicas to process all streams.

My understanding is based on [this conversation](https://quixworld.slack.com/archives/C04KXLCGQ94/p1678805872511799). If you don't specify a consumer group, then all replicas will process all streams.

docs/using.md
tbedford: Yes, I will change it! :)

docs/using.md
tbedford: I've gone with "information" for now.

JotaBlanco: Yep, feature is used in ML world to refer to variables that are use to predict the target variable. Variable, information, parameter, maybe even column should all be valid.

docs/using.md
JotaBlanco: Can't see the new wording :S

docs/using.md
JotaBlanco: OK, I had been understanding this wrong for some time then. I'll need to do some code refactoring :S

docs/using.md
tbedford: Probably the review app was still building. Try now here: https://quixdocsdev.blob.core.windows.net/pr104/client-library/using.html

docs/using.md
tbedford: Changed. I've also fixed a couple of other occurrences in another PR on the main docs repo. I've also updated our writing style guide and best practice guide accordingly. Good shout, thanks!

JotaBlanco (approved):
Consumer groups are much better now.

PR #74, opened 2023-03-25
Fix segfault with some linux distributions

### LINKED ISSUE #70: Segfault with ubuntu 20.04
**Tell us about the bug**
On ubuntu 20.04 with the latest python3 (3.8.10) installed, when an exception is set via the python C API in the c# codebase, segfault occurs.

Example code to produce segfault:
```
from quixstreams import QuixStreamingClient
client = QuixStreamingClient("some-invalid-token")  # this will throw an exception using python C API in c#, but segfaults
```

No Segfault, instead a python exception thrown about invalid token.

**Screenshots**

**What tools and platforms are you using (e.g. iOS, Chrome, Desktop)**
Ubuntu 20.04 (such as windows WSL, google collab)

Works fine in other distributions or when python is compiled from source.

peter-quix:
Fix for #70:
- Replace the python C API use with just interop callback

Additional improvements
- Add support to enable debug via environment variable QuixStreams__Debug
- Improve exception handling by only exposing message and implementing exception type in python also
- Fix incorrect imports (from quixstreams) now using relative import
- Fix segfault when using timedelta with events


luisquix (dismissed):
Looks good

src/PythonClient/src/quixstreams/helpers/exceptionconverter.py
```diff
@@ -0,0 +1,18 @@
+from ..exceptions.quixapiexception import QuixApiException
+from ..native.Python.InteropHelpers.InteropUtils import InteropException
+
+
+class ExceptionConverter:
+
+    @staticmethod
+    def raise_from_interop(exception: InteropException):
+        if exception.exc_type == "QuixStreams.Streaming.QuixApi.QuixApiException":
```
harisbotic, line 9: is it better if use `typeof(QuixApiException).FullName` instead of providing a string?

src/PythonClient/src/quixstreams/helpers/exceptionconverter.py
```diff
@@ -0,0 +1,18 @@
...
+            raise QuixApiException(exception.message)
+
+        if exception.exc_type == "System.NotImplementedException":
+            raise NotImplementedError(exception.message + "\n" + exception.exc_stack)
+
+        if exception.exc_type == "System.Exception":
```
harisbotic, line 15: same recommendation here, typeof(System.Exception).FullName

src/PythonClient/tests/quixstreams/integrationtests/test_integration.py
```diff
@@ -416,6 +416,59 @@ def on_event_data_handler(stream: qx.StreamConsumer, data: qx.EventData):
...
+            read_data = data
+            event.set()
+
+        topic_consumer.on_stream_received = on_stream_received
+
+        # Act
+        print("---- Start reading ----")
+        topic_consumer.subscribe()
+
+        print("---- Start Writing ----")
```
harisbotic: Possibly edit the string to Prepare Producer, because we are not writing/publishing anything in these lines

harisbotic (commented):
Read the code, learned the way we deal with exceptions. Not much I can add atm but here is some small possible improvements

src/PythonClient/src/quixstreams/helpers/exceptionconverter.py
peter-quix: This is in the non-native generated codebase, I don't have access to this type here. Only way  Icould do it is expose the native exception's type and reflect the name out of it, which is massively overkill for this purpose.

src/PythonClient/src/quixstreams/helpers/exceptionconverter.py
peter-quix: This is in the non-native generated codebase, I don't have access to this type here. Only way  Icould do it is expose the native exception's type and reflect the name out of it, which is massively overkill for this purpose.

harisbotic (dismissed):

PR #75, opened 2023-03-27
Performance improvements & buffer fix

peter-quix:
- Fix bug in TimeSeriesBuffer which caused individual row writes for single frame flushes when buffer is effectively disabled
- Update to .net7 where available
- Removed lazy in preparation of ser/des paralellization


luisquix (approved):
Approved

PR #77, opened 2023-03-28
API reference for C# and Python

### LINKED ISSUE #78: C# and Python API docs to be generated based on code documentation
I would like us to have generated documentation similar to other libraries that is automatically updated based on code docs rather than always relying on manual documentation being up-to-date. 

Whatever is acceptable for the given language, such as the Microsoft style for c#.

smadurange:
## Description

Adds API references generated from source code for both Python and C#. Also includes build scripts. Only public classes should be published.

## Review

[Review app](https://quixdocsdev.blob.core.windows.net/pr119/index.html)


peter-quix:
Could you make sure to include the script necessary to generate all this, so in future it can be kept up to date easily?

tbedford (dismissed):
Tested via review app here:

https://quixdocsdev.blob.core.windows.net/pr119/client-library-intro.html

LGTM.

We might need to add a landing page at some point, but I think it's good for now.

tbedford (approved):
I think this is a very good starting point. There are some issues around navigation, and the fact that generated files are stored in GitHub, but this can be addressed in future revisions.

peter-quix:
> I think this is a very good starting point. There are some issues around navigation, and the fact that generated files are stored in GitHub, but this can be addressed in future revisions.

This is a good point actually. I guess we don't necessarily need this in this codebase, as we have the codebase itself.
W would want it in the generated Quix docs. Maybe this should be a build step in https://github.com/quixio/quix-docs instead?

peter-quix (changes requested):
> I think this is a very good starting point. There are some issues around navigation, and the fact that generated files are stored in GitHub, but this can be addressed in future revisions.

This is a good point actually. I guess we don't necessarily need this in this codebase, as we have the codebase itself.
W would want it in the generated Quix docs. Maybe this should be a build step in https://github.com/quixio/quix-docs instead?

SteveRosam:
> > I think this is a very good starting point. There are some issues around navigation, and the fact that generated files are stored in GitHub, but this can be addressed in future revisions.
> 
> This is a good point actually. I guess we don't necessarily need this in this codebase, as we have the codebase itself. W would want it in the generated Quix docs. Maybe this should be a build step in https://github.com/quixio/quix-docs instead?

Yeah that's what I mentioned on Slack.
But having the generated files checked into GitHub does mean they stay with that version of the code. So when we do versioning we wont have the headache of somehow versioning these docs differently

smadurange:
@peter-quix I think I agree with @SteveRosam. The API reference is best versioned together with the API. Also, the doc generators need access to the built assemblies to generate the docs. So, I think the best option is to add the doc generation as a last step in the release pipeline of Quix Streams. If we don't want the API reference docs to live in this repository, we can copy them elsewhere, though I'm not sure if it adds much considering that this repository already has a directory called docs that contain docs for Quix Streams.

peter-quix (approved):
After discussion both here & in Slack, I agree that at the very least for now this is the way to do this.

PR #80, opened 2023-03-28
#79 - Handle KeyboardInterrupt signal when using app.run

### LINKED ISSUE #79: KeyboardInterrupt exception in python when using app.run
**Tell us about the bug**
When the python application making use of `app.run()` gets interrupted using CTRL-C (or OS equivalent), it raises an exception, even though it is supposed to be handled in codebase.

No exception

**Screenshots**
image

peter-quix:
Catch the keyboard interrupt signal, so it can be handled by doing nothing instead of throwing an exception in a context where try/catch is not able to deal with it.

image

### DIFF
--- a/src/PythonClient/src/quixstreams/app.py
+++ b/src/PythonClient/src/quixstreams/app.py
@@ -1,5 +1,6 @@
 import ctypes
 import traceback
+import signal
 from typing import Callable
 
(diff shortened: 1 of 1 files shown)


harisbotic (approved):
LGTM

PR #81, opened 2023-03-29
Documentation refresh

### LINKED ISSUE #82: API documentation out of date at places
Python (and at odd places c#) code - not the .md files - documentation is incorrect/outdated/grammatically wrong. Some of this resulted from 0.4.x->0.5.0 migration.

Code documentation is updated to reflect the latest version of the actual code it is documenting.

harisbotic:
Refreshing code documentation, mostly on the python end


src/CsharpClient/QuixStreams.Streaming/Configuration/SecurityOptions.cs
```diff
@@ -1,27 +1,27 @@
...
 {
     /// <summary>
-    /// Kafka security option for configuring SSL encryption with SASL authentication
+    /// A class representing security options for configuring SSL encryption with SASL authentication in Kafka.
     /// </summary>
     public class SecurityOptions
     {
         /// <summary>
-        /// The Sasl mechanism to use
+        /// The SASL mechanism to use. Defaults to ScramSha256
```
peter-quix: Can leave the "Defaults to " here, given it is set up in the constructor rather than at property level.

src/CsharpClient/QuixStreams.Streaming/ITopicProducer.cs
```diff
@@ -3,36 +3,36 @@
...
+        /// <returns>Stream producer to allow the stream to push data to the platform or null if not found.</returns>
         IStreamProducer GetStream(string streamId);
 
         /// <summary>
         /// Retrieves a stream that was previously created by this instance, if the stream is not closed, otherwise creates a new stream. 
         /// </summary>
         /// <param name="streamId">The Id of the stream you want to get or create</param>
-        /// <param name="onStreamCreated">Callback executed when a new Stream is created in the Output topic because it doesn't exist.</param>
-        /// <returns>Stream writer to allow the stream to push data to the platform.</returns>
+        /// <param name="onStreamCreated">Callback executed when a new Stream is created in the Topic consumer because it doesn't exist.</param>
```
peter-quix: Topic consumer -> topic producer

input topic is what we're consuming (reading data from, gives us data)
output topic is what we're producing to (writing to, sending data to)

src/CsharpClient/QuixStreams.Streaming/Models/StreamConsumer/StreamTimeseriesConsumer.cs
```diff
@@ -51,10 +51,10 @@ public TimeseriesBufferConsumer CreateBuffer(TimeseriesBufferConfiguration buffe
         }
 
         /// <summary>
-        /// Create a new Parameters buffer for reading data
+        /// Creates a new buffer for reading data
         /// </summary>
-        /// <param name="parametersFilter">List of parameters to filter</param>
-        /// <returns>Parameters reading buffer</returns>
+        /// <param name="parametersFilter">Zero or more parameter identifiers to filter as a whitelist. If provided, only these If provided, only these parameters will be available through this buffer</param>
```
peter-quix: Second half has few repetitions

src/CsharpClient/QuixStreams.Streaming/Models/StreamConsumer/StreamTimeseriesConsumer.cs
```diff
@@ -51,10 +51,10 @@ public TimeseriesBufferConsumer CreateBuffer(TimeseriesBufferConfiguration buffe
...
+        /// <returns><see cref="TimeseriesBufferConsumer"/> which will raise new data consumed via the OnDataReceived event</returns>
```
peter-quix: "via the OnDataReceived event" is confusing me here

src/CsharpClient/QuixStreams.Streaming/Models/TimeseriesBuffer.cs
```diff
@@ -9,7 +9,7 @@
 namespace QuixStreams.Streaming.Models
 {
     /// <summary>
-    /// Class used to read from the stream in a buffered manner
+    /// Represents a class used to consume from the stream in a buffered manner.
```
peter-quix: Original comment here may have been incorrect, as this is ultimately used for both consuming and publishing in a buffered manner

src/CsharpClient/QuixStreams.Telemetry/Kafka/TelemetryKafkaProducer.cs
```diff
@@ -24,14 +24,14 @@ public class TelemetryKafkaProducer : StreamComponent, IDisposable
...
-        /// Event raised when an exception occurs during the writing process
+        /// Event raised when an exception occurs during the publishing process
         /// </summary>
         public event EventHandler<Exception> OnWriteException;
 
         /// <summary>
         /// Initializes a new instance of <see cref="TelemetryKafkaProducer"/>
         /// </summary>
-        /// <param name="producer">The input to write the stream packages into. This is something you should share between multiple instances of this class to avoid re-initializing them.</param>
+        /// <param name="producer">The input to produce the stream packages into. This is something you should share between multiple instances of this class to avoid re-initializing them.</param>
```
peter-quix: "the input to produce ... into" sounds a bit wrong. I wonder if we can rephrase without "the input"


Bit of history, so it might shine some light why some naming is a bit weird:


In QuixStreams.Transport, the "input" and "output" was from the point of pipeline. 
Input is where stuff goes in, Output where stuff comes out. So you used to write into the input.


In QuixStreams.Telemetry and QuixStreams.Streaming the "input" and "output" was from the point of the application. This means "input" was what the application received/read, "output" was what the application published/wrote.


After the great renaming of 0.5.0, most of the "input" and "output" got weeded out to avoid confusion either way with "consumer" and "producer" doing consuming/reading and publishing/writing.

We still have the concept "input topic" and "output topic" in places, according to the QuixStreams.Streaming understanding. This means you can subscribe to an "input Topic" using the "topic consumer" and the opposite for the "output topic".

History lesson over, with this information available, are we able to better rephrase this?

src/PythonClient/src/quixstreams/builders/eventdefinitionbuilder.py
```diff
@@ -42,27 +39,31 @@ def set_level(self, level: EventLevel) -> 'EventDefinitionBuilder':
...
-        Add new event definition to the StreamPropertiesProducer. Configure it with the builder methods.
-        :param event_id: The id of the event. Must match the event id used to send data.
-        :param name: The human friendly display name of the event
-        :param description: The description of the event
-        :return: The builder
-        """
+        """Add new Event definition, to define properties like Name or Level, among others.
+
+        Args:
+            event_id: Event id. This must match the event id you use to Event values.
```
peter-quix: > This must match the event id you use to Event values.

This must match the event id you use to send event values.

or publish

src/PythonClient/src/quixstreams/models/parametervalue.py
```diff
@@ -110,15 +123,17 @@ def binary_value(self, value: Union[bytearray, bytes]):
 
     @property
     def type(self) -> ParameterValueType:
-        """Gets the type of value, which is numeric or string if set, else empty"""
+        """Gets the type of value, which is numeric, string if set, or empty."""
```
peter-quix: numeric, string, binary if set, otherwise empty.

src/PythonClient/src/quixstreams/streamconsumer.py
```diff
@@ -149,16 +168,23 @@ def _on_package_received_dispose(self):
...
+        """
         if self._streamId is None:
             self._streamId = self._interop.get_StreamId()
         return self._streamId
 
     @property
     def properties(self) -> StreamPropertiesConsumer:
         """
-        Gets the consumer for accessing timeseries related information of the stream such as parameter definitions and values
+        Gets the consumer for accessing timeseries related information of the stream such as parameter definitions and values.
```
peter-quix: this is a bad original copy paste. steram properties aren't related to timeseries, but to stream itself such as the name, metadata, time recorded etc

src/PythonClient/src/quixstreams/models/streamconsumer/streameventsconsumer.py
```diff
@@ -46,14 +48,22 @@ def _finalizerfunc(self):
     @property
     def on_data_received(self) -> Callable[['StreamConsumer', EventData], None]:
         """
-        Gets the handler for when the stream receives event. First parameter is the stream the event is received for, second is the event.
+        Gets the handler for when an events data package is received for the stream.
+
+        Returns:
+            Callable[['StreamConsumer', EventData]]:
```
peter-quix: Not sure if this is consistent with other places (files I already revieved).

> Callable[['StreamConsumer', EventData], None]

means it takes two parameters and returns type "None" (c# void), so leaving the return type off might not be the best.

src/PythonClient/src/quixstreams/models/streamproducer/streampropertiesproducer.py
```diff
@@ -13,16 +13,16 @@
 @nativedecorator
 class StreamPropertiesProducer(object):
     """
-        Provides additional context for the stream
+    Represents properties and metadata of the stream.
+    All changes to these properties are automatically populated to this class.
```
peter-quix: This would be true for consumer.

StreamPropertiesProducer 
Represents properties and metadata of the stream.
All changes to these properties are automatically published to the underlying stream.

src/PythonClient/src/quixstreams/models/streamconsumer/streampropertiesconsumer.py
```diff
@@ -45,18 +48,24 @@ def _finalizerfunc(self):
...
         return self._on_changed
 
     @on_changed.setter
     def on_changed(self, value: Callable[['StreamConsumer'], None]) -> None:
         """
-        Sets the handler for when the stream properties changed. First parameter is the stream it is invoked for.
+        Sets the handler for when the stream properties change.
+
+        Args:
+            value (Callable[['StreamConsumer']): First parameter is the stream it is invoked for.
```
peter-quix: Callable[['StreamConsumer'], None]

src/PythonClient/src/quixstreams/models/streamconsumer/streamtimeseriesconsumer.py
```diff
@@ -64,21 +66,31 @@ def _finalizerfunc(self):
     @property
     def on_data_received(self) -> Callable[['StreamConsumer', TimeseriesData], None]:
         """
-        Gets the handler for when the stream receives data. First parameter is the stream the data is received for, second is the data in TimeseriesData format.
+        Gets the handler for when data is received (without buffering).
+        This handler does not use Buffers, and data will be raised as they arrive without any processing.
+
+        Returns:
+            Callable[['StreamConsumer', TimeseriesData]: The function that handles the data received.
```
peter-quix: Callable[['StreamConsumer', TimeseriesData], None]

src/PythonClient/src/quixstreams/models/streamconsumer/streamtimeseriesconsumer.py
```diff
@@ -64,21 +66,31 @@ def _finalizerfunc(self):
...
+                The first parameter is the stream that receives the data, and the second is the data in TimeseriesData format.
         """
         return self._on_data_received
 
     @on_data_received.setter
     def on_data_received(self, value: Callable[['StreamConsumer', TimeseriesData], None]) -> None:
         """
-        Sets the handler for when the stream receives data. First parameter is the stream the data is received for, second is the data in TimeseriesData format.
+        Sets the handler for when data is received (without buffering).
+        This handler does not use Buffers, and data will be raised as they arrive without any processing.
```
peter-quix: second sentence adds nothing of value here. First or second only.
Same for getter & on_raw_received

src/PythonClient/src/quixstreams/models/streamconsumer/streamtimeseriesconsumer.py
```diff
@@ -64,21 +66,31 @@ def _finalizerfunc(self):
...
+
+        Args:
+            value: Callable[['StreamConsumer', TimeseriesData]: The function that handles the data received.
```
peter-quix: Callable[['StreamConsumer', TimeseriesData], None]

I'm going to stop leaving comment about this, just do a search for Callable[ and fix the rest

src/PythonClient/src/quixstreams/models/streamconsumer/streamtimeseriesconsumer.py
```diff
@@ -208,14 +250,18 @@ def definitions(self) -> List[ParameterDefinition]:
...
+        Creates a new buffer for consuming data according to the provided parameter_filter and buffer_configuration.
 
-        :returns: a TimeseriesBufferConsumer which will raise new data read via .on_data_released event
+        Args:
+            parameter_filter: Zero or more parameter identifiers to filter as a whitelist. If provided, only those
+                              parameters will be available through this buffer.
+            buffer_configuration: An optional TimeseriesBufferConfiguration.
+
+        Returns:
+            TimeseriesBufferConsumer: An consumer that will raise new data consumed via the on_data_received event.
```
peter-quix: buffer uses on_data_released (and other similarly named), not received
https://github.com/quixio/quix-streams/blob/6b746a0b7dc60119d87dd2a28ff21f3c246e3b97/src/PythonClient/src/quixstreams/models/timeseriesbuffer.py#L76

src/PythonClient/src/quixstreams/models/streamproducer/streameventsproducer.py
```diff
@@ -35,58 +34,62 @@ def __init__(self, net_pointer: ctypes.c_void_p):
 
     def flush(self):
         """
-        Immediately writes the event definitions from the buffer without waiting for buffer condition to fulfill (200ms timeout) #TODO verify 200ms
+        Immediately publishes the event definitions from the buffer without waiting for buffer condition to fulfill
+        (200ms timeout). NOTE: Verify 200ms timeout value.
```
peter-quix: TODO != note.

TODO is internal comment to do something, NOTE is something users should be aware while using

src/PythonClient/src/quixstreams/models/streamproducer/streameventsproducer.py
```diff
@@ -35,58 +34,62 @@ def __init__(self, net_pointer: ctypes.c_void_p):
...
         """
         self._interop.Flush()
 
     @property
     def default_tags(self) -> Dict[str, str]:
-        """Gets default tags injected to all Events Values sent by the writer."""
+        """Default Tags injected to all Event Values sent by the producer."""
```
peter-quix: "Gets the ..." it is a property getter

src/PythonClient/src/quixstreams/models/streamproducer/streameventsproducer.py
```diff
@@ -35,58 +34,62 @@ def __init__(self, net_pointer: ctypes.c_void_p):
...
+
+        Args:
+            value (str): Location string, e.g., "/Group1/SubGroup2".
         """
         self._interop.set_DefaultLocation(value)
 
     @property
     def epoch(self) -> datetime:
-        """Gets the default epoch used for event values"""
+        """Gets the default epoch used for Timestamp event values."""
```
peter-quix: This phrasing doesn't make more sense to me than the last one. 
Did you mean to say "Gets the default epoch used when sending event values using timestamp"? or something along that line.

Also let's avoid random casing in the middle of sentences

src/PythonClient/src/quixstreams/models/streamproducer/streameventsproducer.py
```diff
@@ -35,58 +34,62 @@ def __init__(self, net_pointer: ctypes.c_void_p):
...
 
         ptr = self._interop.get_Epoch()
         value = dtc.datetime_to_python(ptr)
         return value
 
     @epoch.setter
     def epoch(self, value: datetime):
-        """Sets the default epoch used for event values"""
+        """Sets the default epoch used for Timestamp event values. Datetime added on top of all the Timestamps."""
```
peter-quix: I don't understand.

"Datetime added on top of all the Timestamps."

Epoch is only used when the event (or timestamps) is added as an offset from epoch (timedelta, nanoseconds), not when added at an exact DateTime.

Example:
epoch 10, ns 300 -> 310 (Added)
epoch 10, 2021/01/06 15:59:36.3213 -> 2021/01/06 15:59:36.3213 (not added)

src/PythonClient/src/quixstreams/models/streamproducer/timeseriesbufferproducer.py
```diff
@@ -18,18 +18,20 @@
 @nativedecorator
 class TimeseriesBufferProducer(TimeseriesBuffer):
     """
-        Class used to write to StreamProducer in a buffered manner
+    A class for writing time-series data to a StreamProducer in a buffered manner.
```
peter-quix: looks like there is some leftover writing.

Also, we decided timeseries to be a single word for our usage, that is why it is not TimeSeriesData, but TimeseriesData.
Means no hypening it either

src/PythonClient/src/quixstreams/models/streamproducer/timeseriesbufferproducer.py
```diff
@@ -71,58 +94,63 @@ def add_timestamp(self, time: Union[datetime, timedelta]) -> TimeseriesDataBuild
                 return TimeseriesDataBuilder(self._interop.AddTimestamp(netdate_hptr))
             finally:
                 InteropUtils.free_hptr(netdate_hptr)  # dotnet will hold a reference to it, we no longer need it
-        if isinstance(time, timedelta):
-            nettimespan_uptr = dtc.timedelta_to_dotnet(time)
-            return TimeseriesDataBuilder(self._interop.AddTimestamp2(nettimespan_uptr))
-        raise ValueError("'time' must be either datetime or timedelta")
+            if isinstance(time, timedelta):
```
peter-quix: This indenting will break logic

src/PythonClient/src/quixstreams/models/timeseriesbuffer.py
```diff
@@ -202,16 +232,27 @@ def callback(ts: ctypes.c_void_p) -> bool:
...
         return self._custom_trigger
 
     @custom_trigger.setter
     def custom_trigger(self, value: Callable[[TimeseriesData], bool]):
         """
-            Sets the custom function which is invoked after adding a new timestamp to the buffer. If returns true, TimeseriesBuffer.on_data_received is invoked with the entire buffer content
-            Defaults to none (disabled).
+        Sets the custom trigger function, which is invoked after adding a new timestamp to the buffer.
+
+        If the custom trigger function returns True, the `TimeseriesBuffer.on_data_received`
```
peter-quix: on_data_released not on_data_received. this is a leftover after renaming.
Also, in fact on_dataframe_released and on_raw_released also get raised, depending on what user subscribed to. Saw this issue in other comments too, probably were written before the other two were available.

Possibly need to rephrase these in a way that is more abstracted like, in this scenario
"if the custom trigger function returns True, the buffer releases content and triggers relevant callbacks."

src/PythonClient/src/quixstreams/models/timeseriesbufferconfiguration.py
peter-quix, line 1: Likely most comment left on timeseriesbuffer apply here also

src/PythonClient/src/quixstreams/models/timeseriesdataraw.py
```diff
@@ -133,11 +137,14 @@ def to_dataframe(self) -> pd.DataFrame:
...
 
-        :param data_frame: The pandas DataFrame to convert to TimeseriesData
-        :param epoch: The epoch to add to each time value when converting to TimeseriesData. Defaults to 0
-        :return: Converted TimeseriesData
+        Args:
+            data_frame (pd.DataFrame): The pandas DataFrame to convert to TimeseriesData.
+            epoch (int, optional): The epoch to add to each time value when converting to TimeseriesData. Defaults to 0.
+
+        Returns:
+            TimeseriesDataRaw: Converted TimeseriesData.
```
peter-quix: TimeseriesDataRaw: Converted TimeseriesDataRaw.

src/PythonClient/src/quixstreams/models/timeseriesdataraw.py
```diff
@@ -373,9 +391,12 @@ def string_values(self) -> Dict[str, List[str]]:
     @property
     def binary_values(self) -> Dict[str, List[bytes]]:
         """
-            The binary values for parameters.
-            The key is the parameter Id the values belong to
-            The value is the binary values of the parameter. Values are matched by index to timestamps
+        The binary values for parameters.
+        The key is the parameter ID the values belong to. The value is the string values of the parameter.
```
peter-quix: binary

src/PythonClient/src/quixstreams/models/streamproducer/streameventsproducer.py
peter-quix: This phrasing doesn't look too bad, maybe can use as base
https://github.com/quixio/quix-streams/blob/6b746a0b7dc60119d87dd2a28ff21f3c246e3b97/src/PythonClient/src/quixstreams/models/timeseriesdataraw.py#L323

src/PythonClient/src/quixstreams/models/timeseriesbuffer.py
```diff
@@ -75,14 +74,23 @@ def _finalizerfunc(self):
     @property
     def on_data_released(self) -> Callable[[Union['StreamConsumer', 'StreamProducer'], TimeseriesData], None]:
         """
-        Gets the handler for when the stream receives data. First parameter is the stream the data is received for, second is the data in TimeseriesData format.
+        Gets the handler for when the stream receives data. First parameter is the stream the data is received for,
```
peter-quix: There is a bit of inconsistence with the handling of segment "the first parameter ... ".
Most other classes have it in the return statement or in the args statement, not in the property description.

peter-quix (changes requested):

src/PythonClient/src/quixstreams/models/streamproducer/timeseriesbufferproducer.py
harisbotic: I changed time-series to timeseries everywhere we had in in code. Few places only

peter-quix (approved):

PR #87, opened 2023-03-31
Change build scripts to python where possible

peter-quix:
Changed build scripts to python wherever possible to avoid having to maintain batch, bash, zsh scripts.
In a future step, we could lift shared components, but as a starter having 1 language to use is a step forward.

Added the flag support to linux/mac also (debug, nopython, nointerup etc)


src/builds/python/linux/entrypoint.py
harisbotic, line 1: ~~I would rename this file to `build.py` to match the naming convention of other platforms.~~

Nvm, I see now that its different.

src/builds/python/windows/install_dependencies.py
```diff
@@ -0,0 +1,34 @@
...
+            return True
+        else:
+            return False
+    except subprocess.CalledProcessError:
+        return False
+
+# Check for .NET 8 SDK
+if not check_installed_program_version('dotnet', r'^7\.\d+\.\d+'):
+    print('.NET 7 SDK not found or incorrect version. Please install the correct .NET 7 SDK version.')
+    sys.exit(1)
```
harisbotic: Either fix the comment or the code to look for .NET 8 instead of 7.
I would maybe just remove the version number from the comment to remove the need of maintenance

src/builds/python/windows/build_native.py
```diff
@@ -0,0 +1,101 @@
...
+        print("Cleaning interop folder...")
+        shutil.rmtree(destPlatform, ignore_errors=True)
+
+        print("Build interop projects")
+        for subdir in os.listdir(interopoutputcsharp):
+            interop_project_dir = f"{interopoutputcsharp}/{subdir}"
+            dest_platform_subdir = f"{destPlatform}/{subdir}"
+            subprocess.run(f"dotnet publish {interop_project_dir}/{subdir}.csproj /p:NativeLib=Shared /p:SelfContained=true {configuration} -r {dotnetruntime} -o {dest_platform_subdir}", shell=True, check=True)
+    else:
+        print("Not recompiling interop due to --nointerop flag")
```
harisbotic: Something felt a little bit off when I was reading this code, it's either that flag names should be improved or we mixed up the flags in these two methods.
I would use `noregen` flag for `interop_projects` method with `print("Not regenerating interop projects due to --noregen flag")` and `nointerop` flag then for `build_and_run_interop` method with `print("Not recompiling interop due to --nointerop flag")`

The same appears for all platforms (linux, mac, win)

harisbotic (changes requested):
Requesting changes here only because I think we mixed up flags in `build_interop` and `build_interop_projects` methood

src/builds/python/mac/build_native.py
```diff
@@ -0,0 +1,196 @@
...
+import sys
+import shutil
+import subprocess
+import platform
+
+def parse_arguments(args):
+    options = {
+        'nointerop': False,
+        'nopython': False,
+        'noregen': False,
```
harisbotic: Naming of these flags can be a little confusing:
- Calling `build` script in `python` folder has a flag `nopython`. A little bit weird at first
- `noregen`, not regenerating what?

Possible improvement:
- `nopython` -> `nocopy` / `compile-only`
- `noregen` -> `noprojects` / `nointeropprojects`
-  Use `skip` instead of `no` (ex. `skip-interop`, `skip-projects`...)

src/builds/python/windows/build_native.py
peter-quix: noregen = leave the generated interop layers as they are. useful for when doing manual modification to test issues.
nointerop = leave the interop as is. speeds up build when you only modify python
nopython = you can change underlying interop without losing changes made to the generated python. useful for debugging.

I'll think about better names, but the flags are currently correctly checked. Will leave some comments next to the flags

src/builds/python/mac/build_native.py
peter-quix: left a comment on the other one (https://github.com/quixio/quix-streams/pull/87#discussion_r1155476816). I'll leave that unresolved, so you can read & see the modifications.

src/builds/python/windows/build_native.py
peter-quix: applied following changes to all:
https://github.com/quixio/quix-streams/pull/87/files#diff-426e332fa6af2597d556c4d80be243f2bc864d34ccf0369d56b751bd54ea5049R9

harisbotic (dismissed):

harisbotic (approved):

PR #93, opened 2023-04-07
Use .NET 8 for all platforms

harisbotic:
In order to make Native AOT compilation work on Apple silicon (ARM) chips, we are upgrading to .net8 as previously this wasn't supported.

Some library packages are still left on version "7.0.0" for stability and no need to update frequently reasons

Tested on:
Macbook Pro M1 (ARM)
Macbook Pro 2015 (Intel) - (build script works but haven't run pytest yet)

Note: Some integration tests are still failing on python, but only few. We should investigate that before merging

Prerequisites for build.py script to work on mac are:
- ZSH instead of Bash
- Brew already installed


SteveRosam:
Some docs probably need updating to let people know this works now?
e.g. https://github.com/quixio/quix-streams/blob/main/mac-m1-m2-install.md

peter-quix:
> Some docs probably need updating to let people know this works now? e.g. https://github.com/quixio/quix-streams/blob/main/mac-m1-m2-install.md

I have created an issue to track this. For now the process is the same because using our current azure pipeline this proves to be a bit of headache as apple silicon agents are not readily available.

Issue link: https://github.com/quixio/quix-streams/issues/95

peter-quix (approved):

SteveRosam (approved):

PR #120, opened 2023-05-16
Integrate raw events

### LINKED ISSUE #118: Integrate RAW events with streaming context
The API for RAW events in the custom schema is very hidden in the current implementation. Also is not used by Quix platform so data is not surfaced in SaaS.

RAW events surfaced with existing events API.

**Describe alternatives you've considered**
Flattening of messages into existing data frame solution. Will come later.

harisbotic:
Non-quix message (raw) will now be raised as an Event by TopicConsumer

https://github.com/quixio/quix-streams/issues/118


src/CsharpClient/QuixStreams.Streaming/StreamConsumer.cs
```diff
@@ -135,6 +139,20 @@ private void OnEventDefinitionsReceived(IStreamPipeline streamPipeline, QuixStre
             this.OnEventDefinitionsChanged?.Invoke(this, obj);
         }
 
+        private void OnRawDataReceived(IStreamPipeline streamPipeline, byte[] bytes)
+        {
+            this.logger.LogTrace("StreamConsumer: OnRawDataReceived");
+            var ev = new EventDataRaw
+            {
+                Timestamp = DateTime.UtcNow.ToUnixNanoseconds() ,
```
peter-quix: We need to use kafka message timestamp in this case, must be the same value regardless of how many times we read the same event. Preference would be that the Kafka specific logic only lives where kafka specific code currently does and not introduce at new places.

We want to disentangle ourselves from kafka in future for more supported brokers.


You might be able to get the required information in OnStreamPackageReceived handler, but that would also require polluting this class with Kafka specific logic.

src/CsharpClient/QuixStreams.Telemetry/Kafka/KafkaStreamPipelineFactory.cs
```diff
@@ -18,12 +18,18 @@ public KafkaStreamPipelineFactory(IConsumer transportConsumer, Func<string, IStr
...
-        /// <param name="transportContext">The transport contect to retrieve the streamId from</param>
+        /// <param name="transportContext">The transport context to retrieve the streamId from</param>
         /// <param name="streamId">The streamId retrieved</param>
         /// <returns>Whether retrieval was successful</returns>
         protected override bool TryGetStreamId(TransportContext transportContext, out string streamId)
         {
-            streamId = Encoding.UTF8.GetString(transportContext.GetKey());
+            streamId = null;
+
+            if (transportContext.GetKey() != null)
```
peter-quix: Can reduce arrow coding here and avoid retrieving key twice.

```
var key = transportContext.GetKey();
if (key == null) return false;
streamId = Encoding.UTF8.GetString(key);
```

src/CsharpClient/QuixStreams.Transport.Kafka/KafkaHelper.cs
```diff
@@ -17,14 +16,19 @@ public static Package<byte[]> FromResult(ConsumeResult<byte[], byte[]> consumeRe
         {
             var tContext = new TransportContext(new Dictionary<string, object>
             {
-                {KnownTransportContextKeys.MessageGroupKey, consumeResult.Message.Key},
                 {KnownKafkaTransportContextKeys.Topic, consumeResult.Topic},
-                {KnownKafkaTransportContextKeys.Key, consumeResult.Message.Key},
                 {KnownKafkaTransportContextKeys.Partition, consumeResult.Partition.Value},
                 {KnownKafkaTransportContextKeys.Offset, consumeResult.Offset.Value},
-                {KnownKafkaTransportContextKeys.DateTime, consumeResult.Message.Timestamp.UtcDateTime},
+                {KnownTransportContextKeys.BrokerMessageTime, consumeResult.Message.Timestamp.UtcDateTime},
```
peter-quix, line 22: lets delete KnownKafkaTransportContextKeys.DateTime

src/CsharpClient/QuixStreams.Telemetry/Core/StreamPipeline.cs
```diff
@@ -16,11 +16,15 @@ public class StreamPipeline : IStreamPipeline, IDisposable
...
         
-
         private readonly CancellationToken cancellationToken;
         private readonly List<StreamComponent> componentsList = new List<StreamComponent>();
         private bool isClosed = false;
         
+        /// <summary>
+        /// Stream id to use when message key has not been set.
+        /// </summary>
+        public const string DefaultStreamId = "DEFAULT";
```
peter-quix: I'm thinking the value here would be better to be something like NO_STREAM_ID or similar, rather than default.

peter-quix (changes requested):

src/CsharpClient/QuixStreams.Telemetry/Core/StreamPipeline.cs
harisbotic: Makes sense to me. I changed it to 
`public const string DefaultStreamIdWhenMissing = "NO_STREAM_ID";`
Let me know if you got something better

peter-quix (dismissed):

peter-quix (dismissed):

peter-quix (approved):

PR #141, opened 2023-06-15
C# Performance improvements

peter-quix:
- Use ArraySegment at more places to avoid unnecessary byte[] initialization and copying
- Implement custom JSON serializer
- Minor improvements for buffer array concat

### DIFF
--- a/src/CsharpClient/QuixStreams.Telemetry/Models/Telemetry/Parameters/Codecs/TimeseriesDataCustomJsonCodec.cs
+++ b/src/CsharpClient/QuixStreams.Telemetry/Models/Telemetry/Parameters/Codecs/TimeseriesDataCustomJsonCodec.cs
@@ -54,45 +54,24 @@ namespace QuixStreams.Telemetry.Models.Telemetry.Parameters.Codecs
         }
 
 
-        private void Serialize1(StreamWriter streamWriter, TimeseriesDataRaw obj)
-        {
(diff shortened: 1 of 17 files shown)


peter-quix:
Compound effect is 10-20% performance gain when using JSON without introducing breaking changes to the protocol.

API has breaking changes in Transport, but it is not expected to be used anywhere else

src/CsharpClient/QuixStreams.Telemetry/Models/Telemetry/Parameters/Codecs/TimeseriesDataCustomJsonCodec.cs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        private void Serialize2(StreamWriter streamWriter, TimeseriesDataRaw obj)
+        {
+            using (var jsonWriter = new JsonTextWriter(streamWriter))
+            {
+                jsonWriter.CloseOutput = false;
+                new JsonSerializer().Serialize(jsonWriter, obj);
+            }
+        }
```
harisbotic: Serialize2 doesn't seem to be used, and Serialize1 could have a better name

peter-quix: fixed!

harisbotic (approved):
LGTM

PR #142, opened 2023-06-17
Add Leading edge delay config for Timeseries buffer

harisbotic:
### Configuration
``` csharp
streamProducer.Timeseries.Buffer.LeadingEdgeDelay = 5000;
streamProducer.Timeseries.Buffer.TimeSpanInMilliseconds = 500;

streamProducer.Timeseries.Buffer.OnBackfill += (s, e) => 
```

### Principle
Data arriving into buffer not ordered. The principle of delay is wait for data to arrive and send them ordered.

### Backfill event
Items arriving after LeadingEdgeDelay are discarded from the output topic but released in this event for further processing or forwarding.

#### Example
Image data with columns Timestamp and Speed and buffer configured to send data in 500ms batches in 5 seconds delay.

10:00:00.750: 250 arrived

10:00:00.500: 240 arrived

10:00:01.250: 260 arrived
10:00:00.500: 240 sent

10:00:01.000: 255 arrived

10:00:01.500: 270 arrived
10:00:00.750: 250 sent


peter-quix:
Closing PR in favour of https://github.com/quixio/quix-streams/pull/146

Issue #143, opened 2023-06-20
Add ability to optionally overwrite values for existing timestamps when adding to the buffer


w-as:
**Is your feature request related to a problem? Please describe.**
When adding data to the buffer, before data in the buffer is published it would be helpful to have the ability to optionally overwrite data that is already in the buffer for a specific parameter and timestamp.

Given the following use-case:

Source data arrives in the producer via a series of data packets, each containing data for a subset of parameters and timestamps, and these packets of data may include back-filled data.
For example:
```
"packets": [
  { // First packet - Note samples 0.4 and 0.8 are missing
     "timestamps": [0.0, 0.1, 0.2, 0.3, 0.5, 0.6, 0.7, 0.9], 
     "parameter1": [  0,   1,   2,   3,   5,   6,   7,   9],
     "parameter2": [  0,   2,   4,   6,  10,  12,  14,  18]
  },
  { // Second packet - Data sequential and complete
     "timestamps": [1.0, 1.1, 1.2, 1.3, 1.4, 1.5],
     "parameter1": [ 10,  11,  12,  13,  14,  15],
     "parameter2": [ 20,  22,  24,  26,  28,  30]
  },
  { // Third packet - Includes back-fill data for first packet and data which follows the second packet
     "timestamps": [0.4, 0.8, 1.6, 1.7, 1.8, 1.9],
     "parameter1": [  4,   8,  32,  34,  36,  38],
     "parameter2": [  8,  16,  64,  68,  72,  76]
  }
]
```

In order to present a consistent timeseries to the consumers of the Quix stream at a consistent sample rate, when processing each packet we project all timestamps between the first and last timestamp in the packet and carry forward the most recent value when producing our Quix timestamp data if an expected timestamp is not available.

For example, as the first packet above is missing samples for 0.4 and 0.8, when processing the packet, we output the following:
```
{ // Projected samples:                 __+__                    ___+__
     "timestamps": [0.0, 0.1, 0.2, 0.3, |0.4|,  0.5,  0.6,  0.7, | 0.8|,  0.9], 
     "parameter1": [0.0, 1.0, 2.0, 3.0, |3.0|,  5.0,  6.0,  7.0, | 7.0|,  9.0],
     "parameter2": [0.0, 2.0, 4.0, 6.0, |8.0|, 10.0, 12.0, 14.0, |16.0|, 18.0]
}
```

However, for the third packet, we would project all timestamps between 0.4 and 1.9 as below:
```
[...]

PR #144, opened 2023-06-20
Adds GetOrAddTimestamp methods and overwrite overloads for AddValue.

w-as:
Proposal for solution to issue #143 - _Add ability to optionally overwrite values for existing timestamps when adding to the buffer_, enabling the following C# API:

```CSharp
for (var i = 0; i < samples.Timestamps.Length; i++)
{
    var sampleTimestamp = samples.Timestamps[i];
    var value = samples.Values[i];
    var status = samples.Status[i];

    var timestamp = producer.Timeseries.Buffer.GetOrAddTimestampNanoseconds(sampleTimestamp);
    timestamp.AddValue(parameterId, value, overwrite: status); // Only overwrite *existing* samples if status is true
}
```

### DIFF
--- a/src/CsharpClient/QuixStreams.Streaming/Models/StreamProducer/TimeseriesDataBuilder.cs
+++ b/src/CsharpClient/QuixStreams.Streaming/Models/StreamProducer/TimeseriesDataBuilder.cs
@@ -3,7 +3,7 @@
 namespace QuixStreams.Streaming.Models.StreamProducer
 {
     /// <summary>
-    /// Builder for managing <see cref="TimeseriesDataTimestamp"/> instances on <see cref="TimeseriesBufferProducer"/> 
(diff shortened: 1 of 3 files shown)


peter-quix:
@w-as As discussed in alternate channels, we can have something similar to this, just need to pay attention to the tags.

Please have a look at #146, there we have the following syntax:

```
        public LeadingEdgeRow GetOrCreateTimestamp(long timestampInNanoseconds, Dictionary<string, string> tags = null)
```

By passing the tags here, we can do a lookup and give you the correct entry. If your tags are always the same, can always just pass the same dictionary to avoid reallocating a new one every time.

Thoughts?

peter-quix:
Hi!

I'm going to close this PR for now, as we have arrived to an interim solution that will sort your needs. I'll keep the ticket (https://github.com/quixio/quix-streams/issues/143) open, as this is still a valid request. We'll see how we can introduce something like this probably in 0.6.1 or so.

We have some changes coming in 0.6.0, which could clash with this implementation.

Thanks for the submission!

PR #147, opened 2023-06-28
Fix integrations tests failing with segfaults

ghost:
# Expected Behavior
All tests either pass or fail with an informative exception. Segmentation faults are hard to debug and usually don't contain a traceback.

# Actual behavior
The integration tests always fail with a "segmentation fault" error when you run all of them at once in a suite.

# Reason
Consumer and Producer instances are not closed at the end of each test.

# Solution
To close each TopicConsumer, TopicProducer, RawTopicConsumer and RawTopicProducer at the end of each test.

## Updates in this PR

**Consumer and Producer classes**:
    - Added `dispose()` method and context manager interface for each Consumer and Producer class. This method calls `Dispose()` on a respective interop class.
    - `quixstreams.TopicProducer` now uses interop class `TopicProducer` instead of `ITopicProducer`. `ITopicProducer` doesn't provide a `Dispose()` method.
 
**Tests:**
    - Moved integration tests from unittest to pytest to use fixtures. 
    - Introduced `pytest.fixture` functions to create and close Consumer and Producer instances for each test.
    - Split integration tests in multiple classes.
    - Moved custom assertion functions to `tests.utils`.
    - Other minor changes like formatting and refactoring.

### DIFF
--- a/src/PythonClient/src/quixstreams/kafkastreamingclient.py
+++ b/src/PythonClient/src/quixstreams/kafkastreamingclient.py
@@ -120,3 +120,9 @@ class KafkaStreamingClient(object):
         """
         raw_topic_hptr = self._interop.GetRawTopicProducer(topic)
         return RawTopicProducer(raw_topic_hptr)
+
+    def __enter__(self):
+        return self._interop.__enter__()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return self._interop.__exit__(exc_type, exc_val, exc_tb)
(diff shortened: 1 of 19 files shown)


src/PythonClient/src/quixstreams/kafkastreamingclient.py
```diff
@@ -120,3 +120,9 @@ def get_raw_topic_producer(self, topic: str) -> RawTopicProducer:
         """
         raw_topic_hptr = self._interop.GetRawTopicProducer(topic)
         return RawTopicProducer(raw_topic_hptr)
+
+    def __enter__(self):
```
peter-quix: This smells, the interop is very unlikely to have this?

peter-quix:
As per our discussion elsewhere, there are a couple of test that seem to fail, if you can verify those and update the PR accordingly (even if just skipping them for now if segfault), we could get this merged.

peter-quix (approved):

PR #152, opened 2023-06-28
Fix segmentation faults in tests

peter-quix:
- Fix segmentation faults in python (and therefore tests) related to the new json serializer
- Reenabled all tests marked as segfault issue, local testing resulted in no more segfault
- Fix net_list incorrect function arguments, now the tests pass
- Fix several C# tests


tim-quix:
On a somewhat fun tangent, you swapping the inputs to the `assertEqual` made me try to remember the "correct ordering" of expected vs actual, only to find it's actually `Pycharm` doing some print formatting such that the ordering only matters only for that specifically (obviously you want it to be consistent across the tests in general).

https://stackoverflow.com/a/60567202

tim-quix (approved):
Looks pretty straight-forward to me =)

PR #156, opened 2023-07-11
Use client.id to connect to Confluent Cloud Kafka (#155)

### LINKED ISSUE #155: Use client.id to connect to Confluent Cloud Kafka
In the workspaces endpoint there is a new field "brokerSettings.confluentCloudSettings.clientID" that should be sent as "client.id" when we are connecting to a Confluent Cloud kafka cluster.

The field will be present only if brokerType == "ConfluentCloud". This is an example of json returned by the portal:

```
{
  "workspaceId": "quix-workspace-confluent",
  "name": "ConfluentCloud example",
  "status": "Ready",
  "brokerType": "ConfluentCloud",
  "broker": {
    "address": "xxxxx:9092",
    "securityMode": "SaslSsl",
    "sslPassword": "",
    "saslMechanism": "Plain",
    "username": "xxxxx",
    "password": "xxxxx",
    "hasCertificate": false
  },
  "workspaceClassId": "Standard",
  "storageClassId": "Standard",
  "createdAt": "2023-07-04T15:20:04.58Z",
  "brokerSettings": {
    "brokerType": "ConfluentCloud",
    "syncTopics": false,
    "confluentCloudSettings": {
      "apiKey": "xxxxx",
      "apiSecret": "xxxxx",
      "clusterID": "xxxxx",
      "bootstrapServer": "xxxxx:9092",
      "restEndpoint": "xxxxx",
      "clientID": "testclientid"
    }
  },
  "repositoryId": "xxxxx",
  "branch": "main",
  "environmentName": "prod",
  "version": 2,
  "branchProtected": true
}
```

jordivicedo:


src/CsharpClient/QuixStreams.Streaming/QuixApi/Portal/Workspace.cs
```diff
@@ -176,4 +181,194 @@ internal enum WorkspaceStatus
         /// </summary>
         Disabled
     }
+    
+    internal enum WorkspaceBrokerType
```
peter-quix, line 186: Need "Unknown" as first, or something similar, so when can't find a matching enum, it doesn't give the wrong enum.

Also need to test it

src/CsharpClient/QuixStreams.Streaming/QuixApi/Portal/Workspace.cs
```diff
@@ -176,4 +181,194 @@ internal enum WorkspaceStatus
...
+    {
+        /// <summary>
+        /// Broker type
+        /// </summary>
+        public WorkspaceBrokerType BrokerType { get; set; }
+
+        /// <summary>
+        /// Syncronize existing topics in the broker
+        /// </summary>
+        public bool SyncTopics { get; set; }
```
peter-quix: Lets limit the config to only what we need for now. Which is BrokerType and ClientId from the confluent type.

src/CsharpClient/QuixStreams.Streaming.UnitTests/StreamingClientShould.cs
```diff
@@ -0,0 +1,178 @@
...
+        ""branch"": ""chris"",
+        ""environmentName"": ""Testing"",
+        ""version"": 2,
+        ""branchProtected"": false
+    },
+    {
+        ""workspaceId"": ""different"",
+        ""name"": ""Unknown Kafka Workspace"",
+        ""status"": ""Ready"",
+        ""brokerType"": ""SharedKafka"",
```
peter-quix: Why do we have this & brokerSettings.brokerType ?

Is this an error here or in the API?

jordivicedo: This is in the API. I think it is there because frontend is using it to show which kind of kafka we are using (and kept for compatibility of old workspaces)

peter-quix (approved):

harisbotic (approved):

PR #157, opened 2023-07-11
Add AsyncConsumer and AsyncProducer

ghost:
Added AsyncConsumer and AsyncProducer classes to consume and produce messages asynchronously.

## Details
- Both classes are intended to be thin wrappers around `confluent_kafka.Consumer` and `confluent_kafka.Producer` with a similar interface.
- They wrap IO-related methods in threads in order to consume and produce messages in non-blocking way.
- These classes are meant to be a transport layer for the future high-level `StreamConsumer` and `StreamProducer`, and they don't implement any functionality around rebalancing, or grouping data in streams or handling the SerDes


src/PythonClient/tests/quixstreams/test_dataframes/test_kafka/test_consumer.py
```diff
@@ -0,0 +1,956 @@
...
+        self,
+        consumer,
+        topic_factory,
+    ):
+        topic_name, num_partitions = await topic_factory()
+
+        async with consumer:
+            await consumer.subscribe(topics=[topic_name])
+            msg = await consumer.poll(timeout=1)
+            assert msg is None
```
tim-quix, line 60: One thing to consider checking as well: if you do a consumer.poll with a topic that has messages on it (and of course, a consumer that is not on the latest offset), make sure that polling doesn't accidentally consume a message.

I don't fully remember the circumstances, but I think I was polling to get metadata updates and it was possible to get a message if it was ready, which if you do not handle correctly, you will miss that message.

src/PythonClient/src/quixstreams/dataframes/kafka/producer.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def _default_error_cb(error: KafkaError):
+    logger.error(
+        "Kafka producer error",
+        extra={"error_code": error.code(), "error_desc": error.str()},
+    )
+
+
+class AsyncProducer:
```
tomas-quix, line 37: I don't see anything specific to SDF in this class so I would move it outside dataframes folder.

src/PythonClient/tests/quixstreams/test_dataframes/test_kafka/test_producer.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+
+    async def test_producer_produce_blocking_success(self, producer, topic_factory):
+        topic_name, _ = await topic_factory()
+        async with producer:
+            await producer.produce(
+                topic=topic_name, key="test", value=b"test", blocking=True
+            )
+
+    async def test_producer_produce_blocking_failure_raises_error(
```
tim-quix, line 27: This test has failed for me twice in a row:

```       async with producer_factory(extra_config=extra_config) as producer:
>           with pytest.raises(confluent_kafka.KafkaException) as raised:
E           Failed: DID NOT RAISE <class 'cimpl.KafkaException'>

quixstreams/test_dataframes/test_kafka/test_producer.py:39: Failed
```

src/PythonClient/tests/requirements.txt
```diff
@@ -1,4 +1,5 @@
 testcontainers~=3.7.0
```
tim-quix, line 1: I also needed to install `python-dateutil`

tim-quix (commented):
Really small things, looks good though!

src/PythonClient/tests/requirements.txt
tim-quix: CORRECTION:

After doing everything correctly, I actually only needed to install `confluent-kafka`...which is actually a part of the main package, and not tests. Resolving.

src/PythonClient/src/requirements.txt
```diff
@@ -1 +1,2 @@
-pandas>=1.0.0,<2
\ No newline at end of file
+pandas>=1.0.0,<2
+confluent-kafka>=2.1.1,<2.2
```
tim-quix, line 2: I see this added here but it is missing from the setup.py, not sure if intentional or not...don't really know the full development flow  yet =)

src/PythonClient/src/quixstreams/dataframes/kafka/producer.py
ghost: Hey @tomas-quix,
The `dataframes` folder is supposed to be the home for the new API. 
`AsyncProducer` and `AsyncConsumer` are not supposed to be used anywhere else, that's why I'd like to keep it in the same dir.

src/PythonClient/tests/quixstreams/test_dataframes/test_kafka/test_consumer.py
ghost: In this test (and others too), the topic is created as a first step with `uuid` as a name, which guarantees that it's always empty.

src/PythonClient/tests/quixstreams/test_dataframes/test_kafka/test_producer.py
ghost: Thanks for spotting that!
I fixed it in 599622de1a72c24a319f2e227814fb2cffb9f3b4

src/PythonClient/src/requirements.txt
ghost: Yep, thanks, I missed it.
I updated requirements in de34b843a74741ad6486ffbe48b4e47639fd9c9d

src/PythonClient/src/quixstreams/dataframes/kafka/producer.py
tomas-quix: Well, I disagree. There is still a use case for just reading events directly. I did not plan to hide this API from users.

PR #159, opened 2023-07-18
Handle nulls better in State store

peter-quix:
State store will handle nulls and null state values better from now on.

Instead of saving as null and possibly running into null refs at places, it will be removed from the underlying storage. Added extra null checks for safety.

(unrelated) Also fixed accessibility of a class used for internal deserialization as it broke interop generation.


harisbotic (approved):
Good.
I would just suggest that we use string.Empty instead of "" for the next time.

PR #161, opened 2023-07-20
Refactor state management and implement RocksDb state storage support

harisbotic:
Refactor the states management system to support RocksDB and be safer for topic with multiple partitions

Important changes:
* Implement RocksDb storage support for state storage
* Remove FileStorage as a state storage option
* Remove the state support on App and Topic level. (might be added later but with different implementation)
* Remove APIs such as GetAppStates, GetTopicStates and GetStreamStates
* StreamStateManager is revoked on topic rebalancing
- NEW public APIs:
 (App) Get / SetStateStorageType(enum),
 (App) Get / SetStateStorageRootDir(string),
 StreamStateManager.DeleteStates()
 StreamStateManager.DeleteState(stateName)
 TopicConsumer.GetStreamStateManager(streamId)
 IStateStorage.Flush()

If using RocksDb as state storage, the default location on the disk would be: ./states/{consumer_group}/{topic}/{partitionNumber}/

Saving a Scalar state in RocksDB looks like this:
DbDirectory: ./states/{consumer_group}/{topic}/{partitionNumber}/
Key: {stream}/{stateName}/SCALAR
Value: {value}

Saving a Dictionary state in RocksDB looks like this:
DbDirectory: ./states/{consumer_group}/{topic}/{partitionNumber}/
Key: {stream}/{stateName}/{dictStateKey}
Value: {value}


src/CsharpClient/QuixStreams.State.ParallelWriteTest/Program.cs
```diff
@@ -13,9 +13,9 @@ static void Main(string[] args)
         {
             Console.WriteLine("Parallel write test");
 
-            var storage = new LocalFileStorage();
```
peter-quix, line 16: prob can generalize this a bit

src/CsharpClient/QuixStreams.State/Storage/FileStorage/BaseFileStorage.cs
```diff
@@ -214,7 +214,7 @@ public async Task<int> Count()
         public bool IsCaseSensitive => false;
         
         /// <inheritdoc/>
-        public IStateStorage GetOrCreateSubStorage(string subStorageName)
+        public IStateStorage GetOrCreateSubStorage(string subStorageName, string dbName = null)
```
peter-quix: Hm, original idea was to use the substoragename in places wherever a name is needed. I'll check the rest of the code, but this looks suspicious.

Edit: Lets have a chat to understand the need for this. I saw you using it in relation to the upcoming partition support.

peter-quix (changes requested):
As discussed, we need to remove the RocksDB python part, as the underlying library relies on emitting new types which is not supported by NAOT.

src/CsharpClient/QuixStreams.State/Serializers/Crc32.cs
peter-quix, line 1: We can prob remove this, this used to be used for filestorage to make sure file was properly written to disk, but with RocksDB this is on the db

src/CsharpClient/QuixStreams.Streaming.UnitTests/AppShould.cs
```diff
@@ -12,24 +13,24 @@ public class AppShould
         public void GetStateManager_WithoutSetStateStorage_ShouldNotThrowException()
         {
             // Act
-            var manager = App.GetStateManager();
+            var manager = App.GetStateStorageRootDir();
```
peter-quix: Not a manager any more, same applies to state name

src/CsharpClient/QuixStreams.State/Serializers/Crc32.cs
harisbotic: Its used in classes, and I could remove that too,
but since we are going to bring LocalFileStorage back, lets just keep it as we will need it

peter-quix (approved):
Approved, as per our discussion, further testing will be done on 0.6.0 branch and reimplementation of statestorage + python changes

PR #168, opened 2023-08-08
Implement serialization/deserialization for StreamingDataframes

ghost:
- Add Serializer and Deserializer classes for various data types
- Add class Topic that accepts Serializer and Deserializer instances and is responsible for SerDes of the topic messages


tim-quix (approved):
Gave a good look over, looks good to me!

PR #171, opened 2023-08-11
small update to Row, pipeline clones Rows and itself where needed. Pi…

tim-quix:
…peline handles basic ops and filtering


src/PythonClient/src/quixstreams/dataframes/sdf/pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+            data = [func(d) for d in data]
+            return data if any(data) else None
+
+    def process(self, event: Row | List[Row]) -> Optional[Row | List[Row]]:
+        result = event
+        for func in self._functions:
+            print(f'processing func {func._func.__name__}')
+            result = self._process_function(func, result)
+            print(result)
+            if not result:
```
ghost: I think we should do `is None` check instead of bool. Given that Row has `__bool__` implemented, we might filter out the Row that is has `0` or `""` as a value by mistake.

src/PythonClient/src/quixstreams/dataframes/models/rows.py
```diff
@@ -21,3 +20,15 @@ class Row:
     headers: Optional[Mapping | MessageHeaders] = None
     latency: Optional[float] = None
     leader_epoch: Optional[int] = None
+
+    def __getitem__(self, item):
+        return self.value[item]
+
+    def __bool__(self):
+        return bool(self.value)
```
ghost: What is the purpose of having `__bool__` implemented on Row object? 
Plus it checks `bool(self.value)` which can be `False` if value is any of `{}, [], 0, ""`, which are valid values.

src/PythonClient/src/quixstreams/dataframes/sdf/pipeline.py
```diff
@@ -0,0 +1,186 @@
+import uuid
+from typing import Self, Optional, Any, Callable, TypeAlias, List
+import operator
+from copy import deepcopy
+from quixstreams.dataframes.models import Row
+
+UNITS = {"s": "seconds", "m": "minutes", "h": "hours", "d": "days", "w": "weeks"}
+
+OpValue: TypeAlias = int | float | bool
```
ghost: The `|` syntax is not supported in Python 3.8 yet, and Python 3.8 is our minimal supported version for now.
In this case we should use `Union[int, float, bool]` for now.
The pipe syntax is added in Python 3.9

tim-quix:
Leaving the branch for now since it has some history around things like row and df copying, but this branch is now outdated.

PR #173, opened 2023-08-21
StreamingDataFrame skeleton

ghost:
- Implement StreamingDataFrame to perform lazy computations on messages
- Replace AsyncConsumer and AsyncProducer with Consumer & Producer classes to consume messages from Kafka
- Add RowConsumer and RowProducer to serialize/deserialize messages to/from Rows
- Add Runner class to execute StreamingDataFrames


src/PythonClient/src/quixstreams/dataframes/kafka/consumer.py
```diff
@@ -527,70 +477,21 @@ async def close(self):
...
-        Create an instance of `confluent_kafka.Consumer` in a thread and return it.
-        We need threads here because `Consumer` does IO on `__init__`.
-
-        :return: instance of `confluent_kafka.Consumer`
+        Initialize consumer on demand to avoid network calls on object __init__
+        :return: confluent_kafka.Consumer
         """
-        consumer = await self._run_common_executor(Consumer, self._consumer_config)
-        return consumer
+        if self._inner_consumer is None:
```
tim-quix, line 488: I definitely see why you'd want to have a way of avoiding network calls during init, but it might be worth considering another approach so we don't perform this check so many times (every `.poll`, `.store_offset`, etc). Doing a `is None` check like this seems to be on average ~10% slower than not doing a `is None`, so that'll definitely add up in this case!

What about maybe just having some sort of `.run` or `.start` or something on the consumer to kick off the true consumer init? Or maybe the subscribe step does it? Anywho, I'm sure you have many ideas too!

src/PythonClient/src/quixstreams/dataframes/kafka/consumer.py
```diff
@@ -204,27 +183,22 @@ def _on_lost_wrapper(consumer: Consumer, partitions: List[TopicPartition]):
...
         """
         Remove current subscription.
-
           :raises: KafkaException
           :raises: RuntimeError if called on a closed consumer
         """
-        return await self._run_common_executor(self._consumer.unsubscribe)
+        return self._consumer.unsubscribe()
 
     def store_offsets(
```
tim-quix, line 203: Technically out of scope of this commit (I wasn't sure what would be changed WRT consumer/producer), but wondering if we want to do so many conditionals every time we try to store an offset...especially since I imagine the second conditional would prob throw an exception via `confluent-kafka` anyway? (`ValueError('One of "message" or "offsets" must be passed')`)

There are a few other similar instances of these checks throughout as well, and I'm just wondering if the performance hits are worth it? 

TBH I probably wouldn't care othewise, but we have been emphasizing performance so I think it might be worth reconsidering things like that.

src/PythonClient/src/quixstreams/dataframes/rowconsumer.py
```diff
@@ -0,0 +1,196 @@
...
+        topics_names = list(topics_map.keys())
+        super().subscribe(
+            topics=topics_names,
+            on_assign=on_assign,
+            on_revoke=on_revoke,
+            on_lost=on_lost,
+        )
+        self._topics = {t.name: t for t in topics}
+
+    def poll_row(self, timeout: float = None) -> Union[Row, List[Row], None]:
```
tim-quix, line 149: Out of curiosity, why not just extend the original `poll` command? I'm assuming you were worried that you might further extend this class later with future classes?

Might be nice to keep the original/consistent API across the consumers. Not sure if you intend `RowConsumer` to only be used by the `Runner` or not.

src/PythonClient/src/quixstreams/dataframes/rowproducer.py
```diff
@@ -0,0 +1,113 @@
...
+        super().__init__(
+            broker_address=broker_address,
+            partitioner=partitioner,
+            extra_config=extra_config,
+        )
+        self._on_error: Optional[ProducerErrorCallback] = (
+            on_error or default_on_producer_error
+        )
+
+    def produce_row(
```
tim-quix, line 64: Similar question as the `poll_row`, though I think this one potentially makes more sense to keep separate depending on what needs to be produced. I might argue the user probably shouldn't have to worry about the distinction, though.

src/PythonClient/src/quixstreams/dataframes/runner.py
```diff
@@ -0,0 +1,188 @@
...
+from .rowconsumer import RowConsumer
+from .rowproducer import RowProducer
+
+logger = logging.getLogger(__name__)
+
+MessageProcessedCallback = Callable[[str, int, int], None]
+
+
+class Runner:
+    def __init__(
```
tim-quix, line 37: I mostly bring this up now because it would probably dramatically alter the interface here, but we should probably consider config files or something sooner rather than later...there's a bunch of replicated arguments and across everything that we could really condense.

Of course, this is a lot faster to get out for an MVP, so I definitely understand why =). Just saying it might be worth considering soon.

src/PythonClient/src/quixstreams/dataframes/runner.py
```diff
@@ -0,0 +1,188 @@
...
+    ):
+        """
+        Start processing data from Kafka using provided StreamingDataFrame
+
+        :param dataframe: StreamingDataFrame instance
+        """
+        logger.info("Start processing of the streaming dataframe")
+        self._running = True
+        # Initialize consumer and producer instances
+        consumer = RowConsumer(
```
tim-quix: IIRC, the consumer and producers do not make network calls when init, so it feels like the initializations of `RowConsumer` and `RowProducer` might better belong in the `__init__` here. Might allow you to inspect/see that something is wrong with configuration (or other things?) before even attempting to connect/run.

tim-quix (approved):
I would say there's nothing major/necessary to fix, but some things to consider going forward

src/PythonClient/src/quixstreams/dataframes/kafka/consumer.py
ghost: That's a good point, I'll see if we can reduce the amount of `is None` checks

src/PythonClient/src/quixstreams/dataframes/kafka/consumer.py
ghost: It's done this way because `confluent_kafka` handles optional arguments differently. It doesn't like `None` values being passed as a default but instead prefers the argument not to be passed at all.

I'll check later what the performance hit is, but it doesn't block this PR right now.

src/PythonClient/src/quixstreams/dataframes/rowconsumer.py
ghost: I keep the original interface intact because in some situations we might need to call "raw"  version of `.poll()`

src/PythonClient/src/quixstreams/dataframes/runner.py
ghost: Good point, we can discuss it later down the road.

src/PythonClient/src/quixstreams/dataframes/runner.py
ghost: Agree. Moved `consumer` and `producer` to `Runner.__init__` in [6b1aabb](https://github.com/quixio/quix-streams/pull/173/commits/6b1aabbfc2bda6e8578134c564834fec4337950c)

src/PythonClient/src/quixstreams/dataframes/rowproducer.py
ghost: Similar to https://github.com/quixio/quix-streams/pull/173#discussion_r1301798981

PR #175, opened 2023-08-24
added 'to_topic' method to SDF, futher minor cleanup and test fixture…

tim-quix:
… additions

Will probably clean up/consolidate a few fixtures as well since there's some overlap from our separate efforts.


src/PythonClient/tests/quixstreams/test_dataframes/fixtures.py
```diff
@@ -106,6 +125,23 @@ def factory(
     return factory
 
 
+@pytest.fixture()
+def producable_row_factory():
```
ghost: There's already a row_factory in one of the fixture modules. Let's re-use the existing one.

src/PythonClient/tests/quixstreams/test_dataframes/test_rowproducer.py
```diff
@@ -4,45 +4,18 @@
...
-    MessageTimestamp,
-    TimestampType,
     Topic,
     JSONSerializer,
-    JSONDeserializer,
     SerializationError,
 )
 
 
-def row_factory(
```
ghost, line 17: Oh, this is the one that I meant in https://github.com/quixio/quix-streams/pull/175/files#r1304068807 :) 
Why did you decide to rename it?

src/PythonClient/src/quixstreams/dataframes/dataframe/dataframe.py
```diff
@@ -58,6 +57,18 @@ def process(self, row: Row) -> Optional[Union[Row, list[Row]]]:
         """
         return self._pipeline.process(row)
 
+    # TODO: maybe we should just allow list(Topics) as well (in many spots actually)
```
ghost, line 60: Could you pls clarify this one? 
To me it's +- the same:
```
sdf.to_topic(topic1)
sdf.to_topic(topic2)

# vs 

sdf.to_topics([topic1, topic2])
```

src/PythonClient/tests/quixstreams/test_dataframes/test_rowproducer.py
tim-quix: originally I renamed it just in case there were other `row_factory`s lurking around and not make it confusing which was being used; but mostly,  was trying to find a way of adding some qualifier to the name to distinguish why certain fields were being provided in the factory and not others (since there are many attributes on a Row). 

Ultimately, `producable_row_factory` seemed most applicable since it was mostly fields you would provide a typical producer during a `produce` step and thus the ones you'd likely validate on, but I don't like the name I came up with (and the IDE doesn't like the word either, LOL). 

I think I might revert the name to just `row_factory` since it seems like we can use the single factory universally across the whole suite for the moment. We can maybe think of a better name if other `Row` factories are needed.

src/PythonClient/src/quixstreams/dataframes/dataframe/dataframe.py
tim-quix: Resolving this as we discussed that we should just leave it as-is since it's not often you'd produce the exact same message to multiple topics.

src/PythonClient/tests/quixstreams/test_dataframes/fixtures.py
tim-quix: This is now cleaned up in latest commit so that everything uses one Row fixture =)

ghost:
Looks good to me 👍

PR #176, opened 2023-08-25
Add SerDes for Quix messages

ghost:
Added SerDes for Quix data types:
- Timeseries
- Events

Extra:
- Updated `confluent_kafka` to 2.2.0
- Updated how the headers are produced. 
  - Serializers now can define `extra_headers` property/class attribute to specify what headers must be attached with serialized messages
  - `RowProducer` doesn't forward incoming headers by default anymore.

Limitations:
- The current SDK can work with Events and Timeseries at the same time within the same Topic. SDF cannot consume/produce messages of 2 different at the same time.


src/PythonClient/src/quixstreams/dataframes/models/serializers/quix.py
```diff
@@ -0,0 +1,411 @@
...
+        if model_key not in (QModelKey.TIMESERIESDATA, QModelKey.PARAMETERDATA):
+            raise SerializationError(
+                f'"{QModelKey.HEADER_NAME}" must be "{QModelKey.TIMESERIESDATA}" '
+                f'or {QModelKey.PARAMETERDATA}, got"{model_key}"'
+            )
+        if not isinstance(value, Mapping):
+            raise SerializationError(f'Expected mapping, got type "{type(value)}"')
+
+        timestamps = value["Timestamps"]
+        all_params = list(
```
tim-quix: This was not really a critique as much as it was me thinking you did something cool here, and I ended up playing with it a little to see what it was doing =)

If you're interested I also came up with this while I was playing with it, feel free to use it if you'd like, it seems to be 20-25% faster:

```
all_params = [
    (param, iter(values), False if t != "Binary" else True) 
    for t in ['Numeric', 'String', 'Binary'] 
    for param, values in value.get(f'{t}Values', {}).items()
]
```

src/PythonClient/src/quixstreams/dataframes/models/serializers/base.py
```diff
@@ -73,6 +73,19 @@ class Serializer(abc.ABC):
     A base class for all Serializers
     """
 
+    @property
+    def extra_headers(self) -> MessageHeadersMapping:
```
tim-quix, line 77: I don't see a huge issue with this approach overall, but maybe the language should better match the intention?

I like "default_headers", or maybe we want to be stronger with it and call it "enforced_headers" or "required_headers" even (and then maybe enforce they exist before producing, similar to the checks you're doing in the serializer directly...but you might be able to further generalize that to headers as a whole).

tim-quix (approved):
Only a couple minor things that are pretty much just suggestions, seems good to me =)

src/PythonClient/src/quixstreams/dataframes/models/serializers/quix.py
ghost: Agree. I updated the code and added some comments to it.

src/PythonClient/src/quixstreams/dataframes/models/serializers/base.py
ghost: TBH, nothing better than "extra" comes to my mind. 
These are the headers that will be set on top of the already specified by the user, so the word "extra" seems to fit.

- "Required" would mean that a user must specify the headers, which is not exactly the case
- "Enforced" also seems  to miss the target

PR #177, opened 2023-08-25
Feature/sdf column invert

tim-quix:
Small feature addition, at least put some documentation around SDF class (definitely not final).

### DIFF
--- a/src/PythonClient/src/quixstreams/dataframes/dataframe/column.py
+++ b/src/PythonClient/src/quixstreams/dataframes/dataframe/column.py
@@ -10,6 +10,13 @@ ColumnApplier: TypeAlias = Callable[[ColumnValue], OpValue]
 __all__ = ("Column", "OpValue", "ColumnValue", "ColumnApplier")
 
 
+def invert(value):
+    if isinstance(value, bool):
+        return operator.not_(value)
+    else:
+        return operator.invert(value)
+
+
 class Column:
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


src/PythonClient/src/quixstreams/dataframes/dataframe/column.py
```diff
@@ -10,6 +10,13 @@
 __all__ = ("Column", "OpValue", "ColumnValue", "ColumnApplier")
 
 
+def invert(value):
+    if isinstance(value, bool):
+        return operator.not_(value)
+    else:
+        return operator.invert(value)
+
+
```
ghost, line 19: Why do we even check if the value is `bool` to apply `.not_()`? 
I thought that in the context of SDF `~` should always work as boolean negation.

src/PythonClient/src/quixstreams/dataframes/dataframe/dataframe.py
```diff
@@ -23,6 +23,61 @@ def setitem(k: str, v: Union[Column, OpValue], row: Row) -> Row:
...
+    df.to_topic('my_output_topic')
+
+    # Incomplete Rows to showcase what data you are actually interacting with
+    record_0 = Row(value={'column_a': 'a_string', 'column_b': 3, 'column_c': 5})
+    record_1 = Row(value={'column_a': 'a_string', 'column_b': 1, 'column_c': 10})
+
+    # process records
+    df.process(record_0) -> produces {'column_b': 4, 'column_c': 6} to "my_output_topic"
+    df.process(record_1) -> filters row, does NOT produce to "my_output_topic"
+    """
```
ghost: Can it be shortened and simplified?

tim-quix: I am now no longer getting my PhD =)

PR #178, opened 2023-08-29
Remove broker abstraction, introduce kafka message details

peter-quix:
Refactor of the library to remove broker abstraction and expose some crucial features while simplifying implementation. As a result the code now layers in the following manner:
Kafka -> Transport -> Telemetry -> Streaming instead of
Transport -> Kafka -> Telemetry -> Streaming. 

Important changes:
- Introduce KafkaMessage as replacement for RawMessage. Exposes key, value, headers, message time. This is breaking API change for those using raw messages.
- Introduced Header based Quix package information transmission rather than value based. For now value based ser/des is the default to allow time to switch over to 0.6 by clients.

Important Notes:
- This should not cause breaking change for anyone using Streaming layer other than raw messages.
- Disabled message split by default. It can still be re-enabled. 

Additional changes:
- Switch back to .net 7 for C# unless building for python
- Refactor Transport Package to make direct use of offsets and all components to use it
- Refactored Message splitter to read message size information from broker/topic rather than use a configuration from user
- Introduced additional integration tests for the first kafka wrapping lib without additional streaming logic
- Update mac build to use local .dotnet when the version is not the same from global
- Fixed significant number of disabled tests
- Removed the need to open KafkaProducer.  KafkaConsumer is still required to be opened to avoid missing messages before event handlers are subscribed.
- Fixed a possible issue when using datetime with Timezone other than UTC in python


peter-quix:
Notes to self:
- PackageSerializerSettings can't be set from python
- Need to add documentation around PackageSerializerSettings, especially how to re-enable message split

src/CsharpClient/QuixStreams.Kafka.Transport/AutoCommitter.cs
```diff
@@ -0,0 +1,311 @@
...
+        /// </summary>
+        private readonly object commitCheckLock = new object();
+
+        private List<TopicPartitionOffset> committableOffsets = new List<TopicPartitionOffset>();
+        private bool committing = false;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="AutoCommitter"/>
+        /// </summary>
+        /// <param name="commitOptions">Auto commit options</param>
```
harisbotic, line 36: missing commitCallback param here

src/CsharpClient/QuixStreams.Kafka.Transport/AutoCommitter.cs
```diff
@@ -0,0 +1,311 @@
...
+            // copy the settings, not interested in changes.
+            var commitEvery = commitOptions.CommitEvery;
+            var autoCommit = commitOptions.AutoCommitEnabled;
+            var commitInterval = commitOptions.CommitInterval;
+            closeAction = () => { };
+            
+            if (!autoCommit)
+            {
+                // In case there is no auto committing then all we have to do is pass the message up in the chain
+                onPublish = (package, cancellationToken) => this.OnPackageAvailable?.Invoke(package) ?? Task.CompletedTask;
```
harisbotic, line 64: should we async / await this call as well?

src/CsharpClient/QuixStreams.Kafka.Transport/SerDes/KafkaMessageMerger.cs
```diff
@@ -0,0 +1,212 @@
...
+        public KafkaMessageMerger(KafkaMessageBuffer messageBuffer)
+        {
+            this.helper = new KafkaMessageMergerHelper(messageBuffer, this.logger);
+            this.helper.OnMessageSegmentsPurged += (bufferId) =>
+            {
+                if (this.RemoveFromBuffer(bufferId))
+                {
+                    RaiseNextPackageIfReady().GetAwaiter().GetResult();
+                }
+            };
```
harisbotic, line 38: I think this should also be possible
```suggestion
            this.helper.OnMessageSegmentsPurged += async (bufferId) =>
            {
                if (this.RemoveFromBuffer(bufferId))
                {
                    await RaiseNextPackageIfReady();
                }
            };
```

src/CsharpClient/QuixStreams.Kafka/KafkaMessage.cs
```diff
@@ -0,0 +1,188 @@
...
+        
+        /// <summary>
+        /// The headers of the message. Can be null.
+        /// </summary>
+        public KafkaHeader[] Headers { get; protected set; }
+
+        /// <summary>
+        /// Kafka headers
+        /// </summary>
+        protected internal Headers KafkaHeaders { get; protected set; }
```
harisbotic: Its a little bit hard to understand what is what here. I would consider maybe renaming KafkaHeaders to RawKafkaHeaders
OR actually,
write Confluent.Headers instead of just Headers

harisbotic (commented):
Amazing piece of work, the naming and logic is much easier to follow now.
I left few comments

src/CsharpClient/QuixStreams.Kafka.Transport/AutoCommitter.cs
peter-quix: Not necessary, this is returning a task, so whatever is awaiting this task will await that

src/CsharpClient/QuixStreams.Kafka.Transport/SerDes/KafkaMessageMerger.cs
peter-quix: On first glance it looks like it could be done, but it has quite the cascading effect up the chain, making a lot of the code async or do the same thing I did here. After some investigation I decided to not go for this for now.

src/CsharpClient/QuixStreams.Kafka.Transport/AutoCommitter.cs
peter-quix: done

src/CsharpClient/QuixStreams.Kafka/KafkaMessage.cs
peter-quix: clarified by renaming KafkaHeaders to ConfluentHeaders

harisbotic (approved):

PR #181, opened 2023-09-06
first push of API stuff

tim-quix:
Still a little more to clean up around docstrings, exceptions, maybe a little more unit testing? Wanted to at least get it pushed to see if this is along the lines of what we want.

To have a runner work with Quix, you basically would do something like this:

```
cfg_factory = QuixKafkaConfigsFactory()  # automatically sets args via env if in platform, else pass token/ws_id manually
cfgs, topics, group_name = cfg_factory.get_consumer_config(['topic_a', 'topic_b'], 'my_consumer_group_name')

df = StreamingDataFrame(topics)
runner = Runner(cfgs, group_name, ...other args...)
```

We will prob have to alter the runner/clients to either just accept a config dict or add a bunch of args due to various SSL auth stuff and certs....so the former prob makes more sense (or separate out auth stuff?)


tim-quix:
## A couple other stories might come out of this.

Something that is probably fine for now, as workspaces have the workspace ID directly available, and only comes up if you are using a Personal Access Token instead of the SDK-token (which is not the typical way the API is accessed) AFAIK:

- Handle multiple workspace matches when searching. This is especially true with platform v2 via the new "environments" where the same workspace `name` will show up in addition to other potential one-to-many topic-to-workspace matches (though each environment is considered a separate workspace, so SDK-tokens work as usual here).
Will probably add an "environment' property/argument to the `Builder` class, along with some other additional checks with the topic search part. 

Make a Runner config class?

More Optional/QoL:
- Potentially handle other combinations of variable checks (ex: if a topic is provided and only 1 workspace shows up, search confirm the topic is valid, or create it maybe?), searching, error handlings.
- Auto-create topics?
- Make a workspace class?
- Make a QuixTopic or add to current class where there's a name and id attribute to match with how the topics are structured in the Quix platform.

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+            value = self.quix_broker_config[q_cfg]
+            cfg_out[c_cfg] = self.QuixApiKafkaAuthConfigMap.values.get(value, value)
+        cfg_out["ssl.endpoint.identification.algorithm"] = "none"
+        cfg_out["ssl.ca.location"] = self.workspace_cert_path
+        self._confluent_broker_config = cfg_out
+        return self._confluent_broker_config
+
+    def get_confluent_client_configs(
+        self, topics: list, consumer_group_id: Optional[str] = None
+    ):
```
daniil-quix, line 268: The type annotation is missing

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+        filename = "ca.cert"
+        full_path = extract_to_folder / filename
+        if not full_path.is_file():
+            extract_to_folder.mkdir(parents=True, exist_ok=True)
+            self.api.get_workspace_certificate(workspace_id=self._workspace_id).extract(
+                "ca.cert", extract_to_folder
+            )
+        return full_path.as_posix()
+
+    def _set_workspace_cert(self):
```
daniil-quix, line 223: Same here, no typing

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+            workspace_name_or_id = self._workspace_id
+        try:
+            return self.api.get_workspace(workspace_id=workspace_name_or_id)
+        except HTTPError:
+            # check to see if they provided the workspace name instead
+            ws_list = self.api.get_workspaces()
+            for ws in ws_list:
+                if ws["name"] == workspace_name_or_id:
+                    return ws
+            return None
```
ghost: `return None` is not required here, can be omitted

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
ghost: Same here, no typing

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+        cfg_out = {}
+        for q_cfg, c_cfg in self.QuixApiKafkaAuthConfigMap.names.items():
+            value = self.quix_broker_config[q_cfg]
+            cfg_out[c_cfg] = self.QuixApiKafkaAuthConfigMap.values.get(value, value)
+        cfg_out["ssl.endpoint.identification.algorithm"] = "none"
+        cfg_out["ssl.ca.location"] = self.workspace_cert_path
+        self._confluent_broker_config = cfg_out
+        return self._confluent_broker_config
+
+    def get_confluent_client_configs(
```
ghost, line 271: The type annotation is missing

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+
+        :param workspace_id: the workspace to search in
+        :param topic: the topic to search for
+        :return: the workspace_id if success, else None
+        """
+        topics = self.api.get_topics(workspace_id=workspace_id)
+        for t in topics:
+            if t["name"] == topic or t["id"] == topic:
+                return workspace_id
+        return None
```
ghost: `return None` is not required here, can be omitted

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+        :return: workspace data dict if topic search success, else None
+        """
+        ws_list = self.api.get_workspaces()
+        if len(ws_list) > 1:
+            if topic is None:
+                raise self.MultipleWorkspaces(
+                    "More than 1 workspace was found, so you must provide a topic name "
+                    "to find the correct workspace"
+                )
+            print(ws_list)
```
ghost: Do we need this print() here?

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        Find what workspace a topic belongs to.
+        If there is only one workspace altogether, it is assumed to be the workspace.
+        More than one means each workspace will be searched until the first hit.
+
+        :param topic: the topic to search for
+        :return: workspace data dict if topic search success, else None
+        """
+        ws_list = self.api.get_workspaces()
+        if len(ws_list) > 1:
```
ghost: This block can be a bit simplified to reduce the nesting:
```
if len(ws_list) == 1:
    return ws_list[0]

# The rest of the block...

File: src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py [hunk 8]
```diff
@@ -0,0 +1,285 @@
...
+
+    def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> str:
+        """
+        Gets and extracts zipped certificate from the API to provided folder.
+        If no path was provided, will dump to /tmp. Expects cert named 'ca.cert'.
+
+        :param extract_to_folder: path to folder to dump zipped cert file to
+        :return: full cert filepath as string
+        """
+        if not extract_to_folder:
```
ghost: 2 things here:

1. This part can be one-lined to:
```
extract_to_folder = extract_to_folder or Path('/tmp')
```
2. This code will fail on Windows, because there's no `/tmp`
Please have a look at `tempfile.gettempdir()` function

ghost: Plus, I'm not sure how often librdkafka will be re-reading the file. 
If the file is in /tmp, some OS might clean it up someday. 

It's probably fine, but the current SDK keep the certificate in the normal folder AFAIK

src/PythonClient/src/quixstreams/dataframes/platforms/quix/api.py
```diff
@@ -0,0 +1,129 @@
...
+        s.hooks = {"response": lambda r, *args, **kwargs: r.raise_for_status()}
+        s.headers.update(
+            {
+                "X-Version": self.api_version,
+                "Authorization": f"Bearer {self._auth_token}",
+            }
+        )
+        return s
+
+    def get_workspace_certificate(self, workspace_id: Optional[str] = None) -> ZipFile:
```
ghost: Returning the ZipFile here complicates the work for the downstream code a bit.

Instead, we could extract `ca.cert` from the archive right in this function using https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open and return plain `bytes`.

It is a cleaner option, and it's easier for the downstream code to decide what to do with `bytes` than to manipulate the ZIP archive.

ghost (changes requested):

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
tim-quix: oops nope =)

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
tim-quix: Oh man, good catch with `/tmp`. You can see how much I mess with Windows =)

In terms of using the dir, that function is mostly if a user wanted an easy way to grab the cert themselves for whatever reason; the class actually uses `_set_workspace_cert` under the hood (which calls `get_workspace_ssl_cert`) when you call things like `get_confluent_client_configs`.

It uses/sets the same dir as the C client does (if you didn't hand a path at object init): `Path(getcwd()) / "certificates" / self.workspace_id`. Let me know if you think that's too janky.

src/PythonClient/src/quixstreams/dataframes/platforms/quix/api.py
tim-quix: Yeah, makes sense. Changed it =)

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
```diff
@@ -265,7 +264,7 @@ def get_confluent_broker_config(self, known_topic: Optional[str] = None) -> dict
 
     def get_confluent_client_configs(
         self, topics: list, consumer_group_id: Optional[str] = None
-    ):
+    ) -> tuple:
```
ghost: Could you pls specify what types this tuple contains? 
You can do it with `typing.Tuple` class

ghost (changes requested):

src/PythonClient/src/quixstreams/dataframes/platforms/quix/config.py
ghost: Ok, that part makes sense to me.
There's one thing that's not clear:

The `_set_workspace_cert()` function contains `if self._workspace_cert_path` clause.
But the function itself is executed only if `not self._workspace_cert_path`.

In this case, will this code ever be executed?:
```
 if self._workspace_cert_path:
            folder = Path(self._workspace_cert_path)
            if folder.name.endswith("cert"):
                folder = folder.parent
  ```

tim-quix: Ah yeah, you were right; I meant to have `cfg_out["ssl.ca.location"] = self._set_workspace_cert()` so that if the user provided a cert path at init, it would still generate it correctly (if needed) when getting configs. I think that got lost in translation as I was refactoring, good catch!

ghost (approved):
Looks good to go.

PR #182, opened 2023-09-07
Improve throughput of kafka.Producer

ghost:
- Use `poll()` instead of `flush` in case of BufferError
- Add `debug` parameter to enable on_delivery callbacks


tim-quix (approved):
Looks good

ghost:
Rebased from `feature/sdf` and updated the docstring a bit to reflect the changes

PR #183, opened 2023-09-07
Fix off-by-one error in Runner

ghost:
The offset must be incremented when it's passed as `offsets` to `store_offsets()`


tim-quix (approved):
Looks good

PR #184, opened 2023-09-12
Reduce debug log verbosity when broker connections are reaped due to inactivity or other

peter-quix:
Only log if the message is different or there was a reconnect since.


peter-quix:
Don't merge yet, I have included some additional code that should not be in this. Will review on Monday

tomas-quix (approved):

PR #185, opened 2023-09-18
0.5.6 update and improving file based storage resiliency

peter-quix:
- References are only saved on set, not on flush. Required to avoid excessive serialisation or IO
- Added logging capability
- FileStorage now writes to a temp file and moves to original location (framework dependent how). Should provide better resiliency in case app crashes
- Set version to 0.5.6

### DIFF
--- a/src/PythonClient/tests/quixstreams/unittests/state/test_localfilestorage.py
+++ b/src/PythonClient/tests/quixstreams/unittests/state/test_localfilestorage.py
@@ -4,6 +4,9 @@ import unittest
 from src.quixstreams.state.statevalue import StateValue
 from src.quixstreams.state.localfilestorage import LocalFileStorage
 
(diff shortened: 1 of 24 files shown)


harisbotic (dismissed):

src/PythonClient/tests/quixstreams/unittests/state/test_localfilestorage.py
```diff
@@ -4,6 +4,9 @@
 from src.quixstreams.state.statevalue import StateValue
 from src.quixstreams.state.localfilestorage import LocalFileStorage
 
+#from src.quixstreams.logging import Logging, LogLevel
+#Logging.update_factory(LogLevel.Trace)
+
```
harisbotic, line 9: comments left

harisbotic (approved):

PR #188, opened 2023-09-21
Move SDF code to the separate folder in src

ghost:
- Move everything related to Streaming DataFrames to `src/StreamingDataFrames` to simplify testing and packaging
- Update the packaging: simplify `setup.py`, move config to `setup.cfg`


src/StreamingDataFrames/requirements.txt
```diff
@@ -0,0 +1,2 @@
+confluent-kafka>=2.2,<2.3
+requests>=2.31,<2.4
```
tim-quix, line 2: I think I'm the one who did this...but I think this should be <2.32?

src/StreamingDataFrames/.gitignore
```diff
@@ -0,0 +1,25 @@
...
+.ipynb_checkpoints/
+
+# pytest cache
+.pytest_cache
+
+# coverage reports
+.coverage
+
+#ignore environment
+env/
```
tim-quix, line 14: might be a good idea to add `venv/` too since that's what pycharm uses by default

tim-quix (approved):

src/StreamingDataFrames/requirements.txt
ghost: I think it's fine to keep it this way for now.

I was thinking about relaxing the `requests` version even more in case `quixstreams` is used as a library and `requests` is already in the project's requirements, so we don't enforce the version too much

src/StreamingDataFrames/.gitignore
ghost: Yep, I can include it to one of the next PRs

PR #189, opened 2023-09-21
Add CI/CD for StreamingDataFrames

ghost:
# Main changes
- Add pre-commit hooks to run `black` before commit
- Reformat code with black
- Add CI pipeline that lints and tests code

# Notes
The Runner and Consumer tests needed to be updated. 
They were getting stuck in the CI pipeline because `Consumer` sometimes couldn't close itself, even though it works fine locally. 
Apparently, `confluent_kafka.Consumer` doesn't like being passed around between threads, and we use threads in tests.
When you open Consumer in one thread, do `poll()` in another thread, and then close it in the first thread, it might get stuck indefinitely.

The fix is to always keep the work with the Consumer instance in the same thread where it's created.


tim-quix:
Additions and changes make sense :+1:

tim-quix (approved):

PR #194, opened 2023-10-05
Implement basic Rocksdb storage for state based on rocksdict lib

ghost:
- Add RocksDBStorage to access the db partition
- Add TransactionStore to atomically flush the updates

### DIFF
--- a/src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
+++ b/src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
@@ -1,11 +1,24 @@
 import dataclasses
-from typing import Optional
+from typing import Optional, Mapping, Literal
 
 import rocksdict
+from rocksdict import DBCompressionType
 
 from .types import RocksDBOptionsProto
 
 __all__ = ("RocksDBOptions",)
+
+CompressionType = Literal["none", "snappy", "zlib", "bz2", "lz4", "lz4hc", "zstd"]
+
+COMPRESSION_TYPES: Mapping[CompressionType, DBCompressionType] = {
+    "none": DBCompressionType.none(),
+    "snappy": DBCompressionType.snappy(),
+    "zlib": DBCompressionType.zlib(),
+    "bz2": DBCompressionType.bz2(),
+    "lz4": DBCompressionType.lz4(),
+    "lz4hc": DBCompressionType.lz4hc(),
+    "zstd": DBCompressionType.zstd(),
+}
 
 
 @dataclasses.dataclass(frozen=True)
@@ -24,6 +37,7 @@ class RocksDBOptions(RocksDBOptionsProto):
     max_write_buffer_number: int = 3
     block_cache_size: int = 128 * 1024 * 1024  # 128MB
     enable_pipelined_write: bool = False
+    compression_type: CompressionType = "lz4"
     wal_dir: Optional[str] = None
     db_log_dir: Optional[str] = None
 
@@ -46,4 +60,7 @@ class RocksDBOptions(RocksDBOptionsProto):
         table_factory_options = rocksdict.BlockBasedOptions()
         table_factory_options.set_block_cache(rocksdict.Cache(self.block_cache_size))
         opts.set_block_based_table_factory(table_factory_options)
+
+        compression_type = COMPRESSION_TYPES[self.compression_type]
+        opts.set_compression_type(compression_type)
         return opts
--- a/src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
+++ b/src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
@@ -1,0 +1,49 @@
+import dataclasses
+from typing import Optional
+
+import rocksdict
+
+from .types import RocksDBOptionsProto
+
+__all__ = ("RocksDBOptions",)
+
+
+@dataclasses.dataclass(frozen=True)
+class RocksDBOptions(RocksDBOptionsProto):
+    """
+    Common RocksDB database options.
+
+    Please see `rocksdict.Options` for a complete description of each option.
+
+    To provide extra options that are not presented in this class, feel free
+    to override it and specify the additional values.
+    """
+
+    write_buffer_size: int = 64 * 1024 * 1024  # 64MB
+    target_file_size_base: int = 64 * 1024 * 1024  # 64MB
+    max_write_buffer_number: int = 3
+    block_cache_size: int = 128 * 1024 * 1024  # 128MB
+    enable_pipelined_write: bool = False
+    wal_dir: Optional[str] = None
+    db_log_dir: Optional[str] = None
+
+    def to_options(self) -> rocksdict.Options:
+        """
+        Convert parameters to `rocksdict.Options`
+        :return: instance of `rocksdict.Options`
+        """
+        opts = rocksdict.Options(raw_mode=True)
+        opts.create_if_missing(True)
+        opts.set_write_buffer_size(self.write_buffer_size)
+        opts.set_target_file_size_base(self.target_file_size_base)
+        opts.set_max_write_buffer_number(self.max_write_buffer_number)
+        opts.set_enable_pipelined_write(self.enable_pipelined_write)
+        if self.wal_dir is not None:
+            opts.set_wal_dir(self.wal_dir)
+        if self.db_log_dir is not None:
+            opts.set_db_log_dir(self.db_log_dir)
+
+        table_factory_options = rocksdict.BlockBasedOptions()
+        table_factory_options.set_block_cache(rocksdict.Cache(self.block_cache_size))
+        opts.set_block_based_table_factory(table_factory_options)
+        return opts
--- a/src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
+++ b/src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
@@ -252,7 +252,7 @@ class TransactionStore:
         self._storage = storage
         self._update_cache = {}
         self._batch = rocksdict.WriteBatch(raw_mode=True)
-        self._prefix = b""
+        self._prefix = _DEFAULT_PREFIX
         self._failed = False
         self._completed = False
         self._dumps = dumps
@@ -286,7 +286,7 @@ class TransactionStore:
         try:
             yield self
         finally:
-            self._prefix = b""
+            self._prefix = _DEFAULT_PREFIX
 
     @_validate_transaction_state
     def get(self, key: Any, default: Any = None) -> Optional[Any]:
--- a/src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
+++ b/src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
@@ -1,0 +1,432 @@
+import contextlib
+import logging
+import time
+from typing import Any, Union, Optional
+
+import rocksdict
+from typing_extensions import Self
+
+from .exceptions import (
+    StateTransactionError,
+    NestedPrefixError,
+)
+from .options import RocksDBOptions
+from .serialization import serialize, deserialize, serialize_key
+from .types import RocksDBOptionsProto, DumpsFunc, LoadsFunc
+
+__all__ = (
+    "RocksDBStorage",
+    "TransactionStore",
+)
+
+logger = logging.getLogger(__name__)
+
+_sentinel = object()
+
+_PREFIX_SEPARATOR = b"|"
+
+_DEFAULT_PREFIX = b""
+
+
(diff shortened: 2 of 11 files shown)


src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Common RocksDB database options.
+
+    Please see `rocksdict.Options` for a complete description of each option.
+
+    To provide extra options that are not presented in this class, feel free
+    to override it and specify the additional values.
+    """
+
+    write_buffer_size: int = 64 * 1024 * 1024  # 64MB
```
tim-quix, line 35: might be nice to append a human readable unit shorthand to the names, like `write_buffer_size_mb`...especially if we think we can use MB for most things.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+    Serialization
+    *************
+    `TransactionStore` automatically serializes keys and values to JSON.
+
+    Prefixing
+    *********
+    `TransactionStore` allows to set prefixes for the keys in the given code block
+    using :meth:`with_prefix()` context manager.
+    Normally, Streaming DataFrames will set prefixes order to scope
+    the store operations
```
tim-quix: Could you expand on this a little more? It's not clear to me what the intentions of the prefixes are meant to be based on this description of them (not necc important for the code itself, but just for my understanding).

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+        self._prefix = (
+            prefix if isinstance(prefix, bytes) else self._serialize_value(prefix)
+        )
+        if self._prefix:
+            self._prefix += _PREFIX_SEPARATOR
+
+        try:
+            yield self
+        finally:
+            self._prefix = b""
```
tim-quix: very minor thing, but maybe should just do `self._prefix=_DEFAULT_PREFIX` for the sake of consistency?

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+        Check if the transaction has failed.
+
+        The failed transaction should not be re-used because the update cache
+        and
+
+        :return: `True` if transaction is failed, `False` otherwise.
+        """
+        return self._failed
+
+    def _validate_transaction_state(self):
```
tim-quix: Obviously just syntactical sugar, but seems like a good candidate for a decorator

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+        :param key: a JSON-deserializable key.
+        :param value: a JSON-serializable value
+        """
+        self._validate_transaction_state()
+
+        key_serialized = self._serialize_key(key)
+        value_serialized = self._serialize_value(value)
+
+        try:
+            self._batch.put(key_serialized, value_serialized)
```
tim-quix, line 335: It might be more memory efficient to only update the cache here. Then, upon flushing, pop from the cache and add it to the batch object then (and then of course, finally write it all).

Downside of course is then put vs delete behaves slightly diff, which might not be ideal (though deletes are likely to be rare in comparison).

That being said, I have no idea if RocksDB is doing stuff under the hood when you add to the batch that may totally nullify any supposed performance gains due to maybe some other optimizations it might do behind the scenes.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+    )
+
+    def __init__(
+        self,
+        storage: RocksDBStorage,
+        dumps: Optional[DumpsFunc] = None,
+        loads: Optional[LoadsFunc] = None,
+    ):
+        self._storage = storage
+        self._update_cache = {}
```
tim-quix, line 253: This is more of a general comment related to caching and not specific to what you did here. 

I never really tested the speed diff of managing a cache beyond a transaction vs solely depending on RocksDB/Sqlite. Our storage was so slow it wasn't even a contest for our use case at Red Hat. Might be worth investigating at some point (which could lead to some more general cache class or something?) May not be worth messing with it, at least not now. Def adds complexity.

Seeing you only keep a cache around per transaction made me think about it again (as obviously you'll need to look up a key again with a new transaction, but RocksDB should supposedly be quite fast with this since it should be in memory there), just mentioning my musings.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+    ):
+        self._path = path
+        self._options = options or RocksDBOptions()
+        self._open_max_retries = open_max_retries
+        self._open_retry_backoff = open_retry_backoff
+        self._dumps = dumps
+        self._loads = loads
+        self._db = self._open_db()
+
+    def begin(self) -> "TransactionStore":
```
tim-quix, line 73: `begin` feels a bit ambiguous. I feel like just calling it `transaction` would be more clear?

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def flush(self, wait: bool = False):
+        """
+        Manually flush the current mem-table.
+
+        :param wait: whether to wait for the flush to finish
+        """
+        self._db.flush(wait)
+
+    def close(self):
```
tim-quix, line 108: Is flushing upon closing something you need/should do? Or is that handled automatically, frowned upon, or just not something we want to do?

tim-quix (commented):
Seems good and pretty straight-forward. No blocking issues by any means, but take a look at my comments =)

src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
ghost: We use bytes here because the underlying lib also uses bytes. 
And the idea was to keep the names the same with `rocksdict` to avoid confusion. 
We don't provide a full description of the params here and refer to `rocksdict` instead

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: The purpose of prefixes is to namespace the keys used within stateful operations (i.e keys specified by users) across different message keys. 
E.g. when you do `TransactionStore.set('total_count', 123)` for the messages with keys `"key1"`  and `"key2"`, the actual keys in Rocksdb are `"key1"|"total_count"` and `"key2"|"total_count"`

In 99.99% of the cases, prefix equals == current message key. 
The prefixes are going to be set by `StreamingDataFrame` class.

I'll clarify it in the docs.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: Nice find, I missed it.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: The main reason I did it this way was the "delete" operation, yes.

As you noticed, we'd need to come up with a smarter way of keeping track of deleted keys, which seems to me an overcomplication in this case. 
Plus, we'd still spend +- the same memory because in the end you still do `batch.delete()`.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: There might be some misunderstanding about the purpose of the update cache here.

The only purpose of the update cache is to enable a reading of the same values that have been previously updated in the same transaction (aka "reading-your-own-writes" problem).
`rocksdict.WriteBatch` doesn't allow you to check the pending updates before they're flushed to the db.

Optimization of reads is not a goal here.

Example:
```
store = TransactionStore()
store.set('key', 1)  # Set a "key" to "1"
store.get('key')  # Must return "1". 

# Without update cache, it will return whatever value is set in the DB right now,
# not the updated one
```

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: `begin` is a widely used term for "start a transaction" in the DB world, so I want to use it too.
E.g. `BEGIN` in SQL, `Connection.begin()` in sqlalchemy, etc.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: Nice find. 
The `flush` function actually contains a mistake here (it doesn't expect any arguments).
And we don't really call it anywhere right now, `Rdict.close()` already flushes the current memtable inside.
I'll remove it.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: Agree, I moved `_validate_transaction_state` to a decorator

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
tim-quix: Sorry, I totally understand the confusion here...was worried about my comment implying this. 

Though I was in fact talking about optimization of reads as a general idea in my comment, I do understand that's not the purpose of this cache here (which is to just read your own writes until they are actually committed to the DB). 

I should have put it as a general comment, sorry!

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
tim-quix: I see, fair enough. I'm not well enough versed in DB's =)

src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
tim-quix: If you plan on largely being a pass-through to RocksDict, it makes sense. 

You probably saw in fluvii that we tried to make some kafka configs a little easier to manage, because some would randomly be in kb, some in bytes, etc...so we standardized everything to MB for the user, including in the name of the configs, to make it clear what units things were in.

As such, I personally err on the side of making the names more user friendly/clear even if that means not having it be perfectly equivalent to what RocksDict uses - reason being I don't think the average user cares about what RocksDB library we're using, so ultimately we dictate what configuration looks like. For the power users who really need to get into the nitty-gritty,  you could additionally enable a 1:1 config dict passthrough (same idea for Kafka level configs, too), where they'll know we use rocksdict and thus what configs are allowed.

In the end, I don't feel strongly enough about it to push back that much on it as it's largely subjective, but I thought I'd give a little more context to my reasoning.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
tim-quix: Yeah it doesn't really make sense or help with any additional memory with deleting...I also don't know a better way of managing deletes, so I think everything you did makes sense, including matching `put` behavior with `delete`.

It still might be worth considering handling the `put`s slightly differently though since it's is probably the majority of usage. Just depends on if you think the slight difference in behavior is worth it. I also am not 100% sure it would make a difference, but it seems like it should.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/options.py
ghost: In the case of these configs, they're supposed to be changed primarily by those who know how to tune RocksDB (which requires some knowledge). 
I'd rather keep the parameter names transparent to avoid ambiguity.

src/StreamingDataFrames/streamingdataframes/state/rocksdb/stores.py
ghost: From the top of my head, the only case when it makes a difference is when you set the same keys many times during a single transaction.
Then you might have multiple "put" entries in WriteBatch for a single key in the update cache.
But it doesn't seem like a frequent situation, and I don't think it's worth optimizing for right now.

PR #196, opened 2023-10-09
Fix some 0.6.0 issues and update librdkafka to 2.2.0

peter-quix:
- Consumer should complete Open on partition assign on top of wake-up event, which is not always a guarantee without message to consume or when topic and partition is auto-created.
- Fix unnecessary logs for producer when using admin client
- Kafka producer now includes changes from 5b2472, 53a3b6
- Update librdkafka to 2.2.0


peter-quix:
Moved here https://github.com/quixio/quix-streams-dotnet/pull/2

PR #197, opened 2023-10-09
QuixRunner + various changes for 3.8

tim-quix:
Many of the changes here are actually to fix various typing issues and thus allow compatibility with Python v3.8

- Added `QuixRunner` which automates configuration for running in the Quix Platform
    - edits `Topic` objects when `QuixRunner.run()` is called by setting the `real_name` attribute to the actual Quix topic name, which appends the workspace_id to it.
    - adds all the necessary config values for authenticating to Quix broker.
 
-  `Topic` class now has `real_name` attribute, which defaults to `name` attribute if not defined. 
    - functionality otherwise remains the same for non-quix applications.
    - You are still expected to  use the `name` attribute, QuixRunner handles finding the "real" name
  
- `RowProducer` and `RowConsumer` now internally reference `Topic.real_name`, but functionality is equivalent to previous version.


src/StreamingDataFrames/requirements.txt
```diff
@@ -1,2 +1,3 @@
 confluent-kafka>=2.2,<2.3
-requests>=2.28
+requests>=2.28,<2.29
```
ghost: Hey @tim-quix , could you pls provide some details on why we need `requests` to be `2.28` here? 

I relaxed the `requests` version specifically here because it's a widely used library, and other projects might already have it in their requirements. 
Otherwise, there might be either dependency resolution conflicts or we just enforce old version of `requests` for no reason.

src/StreamingDataFrames/streamingdataframes/models/serializers/base.py
```diff
@@ -65,7 +65,7 @@ def _to_dict(self, value: Any) -> Union[Any, dict]:
 
     @abc.abstractmethod
     def __call__(self, *args, **kwargs) -> Any:
-        ...
+        pass
```
ghost: Is there a reason why we can't use `...` instead of `pass`?

src/StreamingDataFrames/streamingdataframes/dataframe/dataframe.py
```diff
@@ -71,19 +72,22 @@ class StreamingDataFrame:
     """
 
     def __init__(
-        self, topics: List[Topic], _pipeline: Pipeline = None, _id: str = None
+        self,
+        topics_in: List[Topic],
+        topics_out: Optional[List[Topic]] = None,
```
ghost: The meaning of this parameter is a bit unclear to me.
Do we now expect users to provide the list of output topics on `StreamingDataFrame.__init__`?

src/StreamingDataFrames/streamingdataframes/models/rows.py
```diff
@@ -6,7 +6,7 @@
 
 
 # TODO: add other dict functions like .get() , __contains__  and .copy()
-@dataclasses.dataclass(eq=True, kw_only=True, slots=True)
+@dataclasses.dataclass(eq=True)
```
ghost: Since the `dataclasses` lib doesn't allow `slots=True` argument in 3.8, we need to manually specify the `__slots__`

src/StreamingDataFrames/streamingdataframes/models/timestamps.py
```diff
@@ -12,7 +12,7 @@ class TimestampType(enum.IntEnum):
     TIMESTAMP_LOG_APPEND_TIME = 2  # broker receive time
 
 
-@dataclasses.dataclass(slots=True, init=True, frozen=True)
+@dataclasses.dataclass(init=True, frozen=True)
```
ghost: `__slots__` now need to be added manually

ghost (changes requested):

src/StreamingDataFrames/streamingdataframes/models/serializers/base.py
tim-quix: In most cases you still can and I was wrong here (IDE was not very clear about where the error was); I added them back where applicable. You can't do it with respect to types, so those remain replaced =)

src/StreamingDataFrames/streamingdataframes/dataframe/dataframe.py
tim-quix: Removed as an argument; now handled automatically by the runner

src/StreamingDataFrames/streamingdataframes/models/rows.py
tim-quix: Removed all data classes and replaced them with a very basic class that enables simple equality checks on the slotted items (useful for a bunch of the unit tests we had)

ghost (approved):
Looks good to me 👍

PR #199, opened 2023-10-12
Introduce Application class as a replacement for Runner and QuixRunner

ghost:
- Application is now responsible for running StreamingDataFrames and for initialization of StreamingDataFrames and Topics
- "Application.Quix()" method is to be used for configuration of the app to work with Quix platform instead of QuixRunner
- Removed Runner and QuixRunner classes
- Updated docs


tim-quix (approved):
Looks good, MERGE IT BB

PR #200, opened 2023-10-12
Sdf apply row values

tim-quix:
Changed the behavior of `StreamingDataFrame.apply(func)`:

- As a point of clarification, it still returns a `Row` or `List[Row]`
    - however, can no longer return `None`
- The `func` argument now expects to receive a `dict` as an argument, and a `dict`, `list`, or `None` as return type
    - This return value will replace the original `Row.value`
- The `func` now treats `None`s (i.e. no return) as an in-place modification and will then return the `Row` with the original (and likely modified) `Row.value`, as previously mentioned
    -  This means "filtering" an entire `Row` with `.apply()` is now no longer possible since `None` used to denote the row was filtered (i.e. used to work how `sdf[sdf['blah'] > 0]` currently works).
- To have a `Row` be parsed into multiple Rows, you must now pass a new `expand` argument to apply, defaults to `expand=False`

Added a `Row.clone()` method, which allows you to pass in any `kwargs` to pass as new values, else clone the originals. It is done in a more "manual" way because it is much more performant than a simple full `deepcopy`.


src/StreamingDataFrames/streamingdataframes/dataframe/dataframe.py
```diff
@@ -23,6 +24,34 @@ def setitem(k: str, v: Union[Column, OpValue], row: Row) -> Row:
...
+    func: Callable[[dict], Optional[Union[dict, List[dict]]]],
+    expand: bool = False,
+) -> Union[Row, List[Row]]:
+    """
+    For StreamingDataFrame.apply()
+
+    Gives `func` the flexibility of either returning a new dict (or list if expanding)
+    or modifying the row value dict in-place (i.e. returning None).
+
+    This means "filtering" is not possible; a Row will always be returned.
```
ghost: I'd remove this sentence from here. It basically explains what this function does not do, and it's confusing :)

src/StreamingDataFrames/streamingdataframes/dataframe/dataframe.py
```diff
@@ -23,6 +24,34 @@ def setitem(k: str, v: Union[Column, OpValue], row: Row) -> Row:
...
+    result = func(row.value)
+    if result is None and isinstance(row.value, dict):  # assume edited in-place
+        return row
+    if isinstance(result, dict):
+        row.value = result
+        return row
+    if isinstance(result, list):
+        if expand:
+            return [row.clone(value=r) for r in result]
+        raise InvalidApplyResultType("'list' not allowed unless 'expand=True'")
```
ghost: How does this sound?
> "Returning "list" type is not allowed unless "expand=True" is passed"

src/StreamingDataFrames/streamingdataframes/dataframe/dataframe.py
```diff
@@ -86,18 +115,12 @@ def __init__(
...
-        :param func: callable that accepts and (usually) returns a QuixStreams Row
-        :return: self (StreamingDataFrame)
-        """
-        self._pipeline.apply(func)
-        return self
+    def apply(
+        self,
+        func: Callable[[dict], Optional[Union[dict, List[dict]]]],
+        expand: bool = False,
+    ) -> Self:
```
ghost, line 117: It's better to keep the docstring here rather than in `apply()` func because this a part of public interface

src/StreamingDataFrames/streamingdataframes/models/rows.py
```diff
@@ -23,14 +26,20 @@ class Row:
...
-        topic: str,
-        partition: int,
-        offset: int,
-        size: int,
-        timestamp: MessageTimestamp,
+        topic: Optional[str] = None,
+        partition: Optional[int] = None,
+        offset: Optional[int] = None,
+        size: Optional[int] = None,
+        timestamp: Optional[MessageTimestamp] = None,
```
ghost: Why these params became optional?

src/StreamingDataFrames/streamingdataframes/models/rows.py
```diff
@@ -23,14 +26,20 @@ class Row:
         "leader_epoch",
     )
 
+    # TODO: Maybe include headers here for if/when it's a dict?
+    _copy_map = {
+        "value": lambda self, k: deepcopy(getattr(self, k)),
+        "timestamp": lambda self, k: copy(getattr(self, k)),
```
ghost: Do we really need to use `getattr` here? 
Looks like `k` is always supposed to be the same as a dictionary key, right?

E.g.
```
"value": lambda self, k: deepcopy(self.value),
```

src/StreamingDataFrames/streamingdataframes/models/rows.py
tim-quix: This one is a bit more of a "suggestion" that I'm fine reverting: I thought this might help where you have to provide a bunch of these args that generally don't make sense, or arent really required, in the context of producing a new Row (my example app I made for show and tell comes to mind).

Might be a better way to do it, but this seemed like an easy thing to do in the interim. If you're not a fan we can revert it =)

src/StreamingDataFrames/streamingdataframes/models/rows.py
tim-quix: True, good catch...faster =)

src/StreamingDataFrames/streamingdataframes/models/rows.py
tim-quix: Reverting

ghost (approved):
It's good to go

PR #202, opened 2023-10-17
auto-create topics

tim-quix:
Topics are now allowed be auto-created when using the Quix Platform (via `Application.Quix()`) when calling `.run()`.

Since auto-creating topics is not a valid option for non-Quix apps, the variable is stored by the `QuixKafkaConfigsBuilder`, and is set by the `Application.Quix(auto_create_topics=BOOL)` (default=`True`)

If True, create any missing topics, ignoring existing. There is a brief period we must wait for them to be finalized.
If False, validate that the expected topics exist, else raise an exception.

The topics referenced are whatever topics are defined on the SDF, both `input_topics` and `output_topics`. 

For topic configuration, a new class `TopicCreationConfigs` was added, which `Topic.creation_configs` now optionally references (and can be handed at init via `creation_configs`). These are basically all the broker-level settings you could normally provide Kafka during topic creation. Any non-required configuration settings are handled with `TopicCreationConfigs.optionals`...which is almost all of them except partitions and rep factor.

There is some other general cleanup, mostly around the API wrapper.

### DIFF
--- a/src/StreamingDataFrames/streamingdataframes/app.py
+++ b/src/StreamingDataFrames/streamingdataframes/app.py
@@ -1,8 +1,9 @@
 import contextlib
 import logging
-from typing import Optional, List, Callable
+from itertools import chain
 
 from confluent_kafka import TopicPartition
+from typing import Optional, List, Callable
 from typing_extensions import Self
 
 from .dataframe import StreamingDataFrame
@@ -13,7 +14,14 @@ from .error_callbacks import (
     default_on_processing_error,
 )
 from .kafka import AutoOffsetReset, AssignmentStrategy, Partitioner
-from .models import Topic, Deserializer, BytesDeserializer, Serializer, BytesSerializer
+from .models import (
+    Topic,
+    TopicCreationConfigs,
+    Deserializer,
+    BytesDeserializer,
+    Serializer,
+    BytesSerializer,
+)
 from .platforms.quix import QuixKafkaConfigsBuilder
 from .rowconsumer import RowConsumer
 from .rowproducer import RowProducer
@@ -154,12 +162,13 @@ class Application:
         consumer_poll_timeout: float = 1.0,
         producer_poll_timeout: float = 0.0,
         quix_config_builder: Optional[QuixKafkaConfigsBuilder] = None,
+        auto_create_topics: bool = False,
     ) -> Self:
         """
         Initialize an Application to work with Quix platform,
         assuming environment is properly configured (by default in the platform).
 
-        It takes the credenetials from the environment and configures consumer and
+        It takes the credentials from the environment and configures consumer and
         producer to properly connect to the Quix platform.
 
         .. note:: Quix platform requires `consumer_group` and topic names to be prefixed
@@ -206,10 +215,13 @@ class Application:
 
         :param quix_config_builder: instance of `QuixKafkaConfigsBuilder` to be used
             instead of the default one.
+        :param auto_create_topics: Whether to auto-create any topics handed to a
+            StreamingDataFrame instance (topics_in + topics_out).
 
         :return: `Application` object
         """
         quix_config_builder = quix_config_builder or QuixKafkaConfigsBuilder()
+        quix_config_builder.app_auto_create_topics = auto_create_topics
         quix_configs = quix_config_builder.get_confluent_broker_config()
         broker_address = quix_configs.pop("bootstrap.servers")
         # Quix platform prefixes consumer group with workspace id
@@ -245,6 +257,7 @@ class Application:
         key_deserializer: Optional[Deserializer] = BytesDeserializer(),
         value_serializer: Optional[Serializer] = None,
         key_serializer: Optional[Serializer] = BytesSerializer(),
+        creation_configs: Optional[TopicCreationConfigs] = None,
     ) -> Topic:
         """
         Create a topic definition.
@@ -257,19 +270,24 @@ class Application:
         :param key_deserializer: a deserializer for keys
         :param value_serializer: a serializer for values
         :param key_serializer: a serializer for keys
-
+        :param creation_configs: settings for topic creation, if needed
         :return: `Topic` object
         """
         if self._quix_config_builder is not None:
             # This Application object was created via `.Quix()` method.
             # Quix platform's workspace id is used as topic prefix.
             name = self._quix_config_builder.append_workspace_id(name)
+            if not creation_configs:
+                creation_configs = TopicCreationConfigs(
+                    num_partitions=2, replication_factor=2
+                )
         return Topic(
             name=name,
             value_serializer=value_serializer,
             value_deserializer=value_deserializer,
             key_serializer=key_serializer,
             key_deserializer=key_deserializer,
+            creation_configs=creation_configs,
         )
 
     def dataframe(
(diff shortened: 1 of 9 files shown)


src/StreamingDataFrames/streamingdataframes/platforms/quix/config.py
```diff
@@ -210,6 +239,64 @@ def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> st
...
+                )
+            else:
+                self._create_topic(topic)
+                finalize.add(topic.name)
+        logger.info(
+            "Topic creations acknowledged; waiting for 'Ready' statuses..."
+            if finalize
+            else "No topic creations required!"
+        )
+        while finalize:
```
ghost: Should we maybe add some timeout here? 
In case the topic creation takes more than expected.

ghost: BTW, does the API class set any request timeouts at all?

src/StreamingDataFrames/streamingdataframes/app.py
```diff
@@ -270,6 +288,20 @@ def stop(self):
         """
         self._running = False
 
+    def _quix_runtime_init(self, dataframe: StreamingDataFrame):
+        """
+        Do some runtime setup only applicable to an Application.Quix instance
+
+        :param dataframe: instance of `StreamingDataFrame`
+        """
+        if self._quix_config_builder:
```
ghost: I think it's a good point to introduce some  `is_quix_app()` property vs checking the `self._quix_config_builder` everywhere

src/StreamingDataFrames/streamingdataframes/platforms/quix/config.py
tim-quix: Added

src/StreamingDataFrames/streamingdataframes/app.py
tim-quix: done!

tim-quix:
Sorry, took a little longer than expected to make the changes, mostly because I thought of a few other things that should be included in addition to your suggestions.

I realized a few different edge-cases (well, honestly if you run more than 1 app, it's probably likely to run into some of them) around topic creation, like if multiple apps try to make the topics at the same time, or someone uses the config builder to create topics directly, which wouldn't use the correct name. 

TL;DR topic creation is more robust now.

Anywho, feel free to look over the new logic!

src/StreamingDataFrames/streamingdataframes/platforms/quix/config.py
```diff
@@ -210,6 +242,89 @@ def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> st
...
+            f"Creation of topic {topic_name} acknowledged by broker. Must wait "
+            f"for 'Ready' status before topic is actually available"
+        )
+
+    def _finalize_create(self, topics: Set[str], timeout: Optional[int] = None):
+        stop_time = time.time() + (timeout or len(topics) * 30)
+        while topics and time.time() < stop_time:
+            # Each topic seems to take 10-15 seconds each to finalize (at least in dev)
+            time.sleep(1)
+            for topic in [t for t in self.get_topics() if t["id"] in topics.copy()]:
```
ghost, line 299: Do we need to copy `topics` on each iteration?
Looks like we do could it once before the loop.

src/StreamingDataFrames/streamingdataframes/platforms/quix/config.py
```diff
@@ -210,6 +242,89 @@ def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> st
...
+                if topic["status"] == "Ready":
+                    logger.debug(f"Topic {topic['name']} creation finalized")
+                    topics.remove(topic["id"])
+        if topics:
+            raise self.CreateTopicTimeout(
+                f"Creation succeeded, but waiting for 'Ready' status timed out "
+                f"for topics: {[self.strip_workspace_id(t) for t in topics]}"
+            )
+
+    def create_topics(self, topics: Iterable[Topic], finalize_timeout_seconds=None):
```
ghost: missing type annotation for `finalize_timeout_seconds`

src/StreamingDataFrames/streamingdataframes/models/topics.py
```diff
@@ -23,6 +24,20 @@
...
 
+__all__ = (
+    "TopicCreationConfigs",
+    "Topic",
+)
+
+
+class TopicCreationConfigs:
+    def __init__(
+        self, num_partitions: int, replication_factor: int, optionals: dict = None
```
ghost: Would it be cleaner if we had a dataclass here instead, and defined all the default values on the class level?

`retention.bytes` and `retention.ms` could also be converted to `retention_bytes` and `retention_minutes` and passed to `post_topic()` as they are.

And we could also remove the default values from [here](https://github.com/quixio/quix-streams/pull/202/files#diff-49b0ec4d7de4eb00fd3a27a08e2edc351bfea1468f4ce4455b88ce573c1a0e5eR136-R139) and from `post_topic()` so they're not spread across the code

ghost (changes requested):

src/StreamingDataFrames/streamingdataframes/platforms/quix/config.py
tim-quix: Yeah, not sure which is better....I was cloning it each time since it was popping things out of it each loop.

ghost: Oh, I see, it throws `RuntimeError: Set changed size during iteration`.
Ok then

src/StreamingDataFrames/streamingdataframes/models/topics.py
tim-quix: Yeah, I think we can use a dataclass! 

The main reason I didn't include `retention.bytes` and `retention.minutes` as class values (written as you suggested with underscores) is that they are technically optional wrt topic creation, so I only made the required arguments as attributes. 

If we think we'll only use this with Quix ever, then I'm fine with changing that...or overall we surface the most common ones and leave some of the more obscure ones under "optional" (which I'd def say those two are common enough). I can see the argument either way.

Similarly, It's definitely awkward to have the default spread across the code, but wasn't sure the best way to do it since you could create topics with the API or with the Quix library, and not sure if we should have defaults in general for partition count and rep factor, and whether it should be set to the Quix defaults vs the Kafka defaults (aka use it for non-quix stuff)?

TL;DR...do we expect to use this beyond Quix stuff and/or should we be opinionated about the "common" topic options and hide the rest?

ghost: Well, this is the code to work with Quix platform API, so I'd consider this as Quix-only. 
This object is only used to specify parameters for the API call.

ghost: I suggest we also move it to `platforms/quix` to avoid confusion

tim-quix: Seems reasonable, will do

tim-quix:
Moved some things around per Daniils second review. Mostly, things were simplified by moving the topic creation management more to the `QuixConfigBuilder` and off the Topic object since we don't expect to mess with it with non-Quix apps. Also no longer need to inspect the SDF object to see what topics we should create, it just goes off of what topic objects were created with the App.topic call.

PR #203, opened 2023-10-19
add some examples (uses auto-create-topics)

tim-quix:
Created some examples along with some boilerplate/helpers to easily get started. 

As part of it, I have included some `.env` files for users to fill out as needed. Unfortunately there's no good way to force files to be otherwise untracked while also including them in the repo (at least not without the user doing something themselves). Another option was make a script to generate them from templates, like `.env.template`, but that also felt a bit meh. Not thrilled with the solution but it's probably fine for now.

See `StreamingDataFrame/examples/README.md` for details on the examples themselves.

Still not overall sure about how to organize it but figured this is a good start and things will prob move around as we eventually add more things.


src/StreamingDataFrames/examples/platforms/quix/timeseries_row_producer.py
```diff
@@ -0,0 +1,66 @@
...
+    "Mal-Wart",
+    "Bikey Bikes",
+    "Board Game Grove",
+    "Food Emporium",
+]
+
+
+# strings for key and headers will be serialized to bytes by default
+i = 0
+try:
```
ghost: You can use `with RowProducer ...` instead of `try-finally` here. 
It flushes itself on closing

tim-quix: Oh good catch!

src/StreamingDataFrames/examples/bank_example/quix_platform_version/producer.py
```diff
@@ -0,0 +1,65 @@
...
+            headers=headers,
+            key=account_id,
+            value=serialize(
+                value=value, ctx=SerializationContext(topic=topic, headers=headers)
+            ),
+        )
+        i += 1
+        sleep(random())
+finally:
+    producer.flush(10)
```
ghost: Producer can be used as a context manager

src/StreamingDataFrames/examples/bank_example/quix_platform_version/alternate_producer.py
ghost, line 1: I suggest we remove this example for now.
RowProducer is not really supposed to be used outside of SDF right now.

The need to pass all of these dummy params to `Row` makes it not really usable, so there's no value in having it as an example.

We can add it as an example when we make it more user-friendly.

src/StreamingDataFrames/examples/bank_example/json_version/producer.py
```diff
@@ -0,0 +1,49 @@
...
+    "Tasty Pete's Burgers",
+    "Mal-Wart",
+    "Bikey Bikes",
+    "Board Game Grove",
+    "Food Emporium",
+]
+
+# strings for key, value, and headers will be serialized to bytes by default
+i = 0
+try:
```
ghost: Same, Producer can be used as a context manager

ghost (changes requested):

ghost (approved):
Looks good to go!

PR #204, opened 2023-10-19
Add StateStoreManager and connect state to Application

ghost:
# Overview
Introduce `StateStoreManager` class that is responsible for reacting to rebalancing callbacks and managing the individual store partitions according to Consumer assignment.

# State workflow:

1. Create `Application` instance
2. Initialize `StreamingDataFrame`
3. Define the StreamingDataFrame pipeline.
`StreamingDataFrame` informs `StateStoreManager` what stores it needs and for which topics.
`StateStoreManager` saves this info and will use it when the application starts.
4. Run the Application and subscribe to topics
5. If the application is stateful (`state_dir` parameter passed) and `StateStoreManager` has at least one store registered:
- - `StateStoreManager` starts reacting to on_assign, on_revoke and on_lost callbacks, opening the necessary DB partitions.
- - For each incoming message, `StateStoreManager` starts a special transaction (`_MultiStoreTransaction`), that manages the actual transactions (instances of `PartitionTransaction`) for each store partition.

`PartitionTransaction` is the primary interface for interacting with key-value data, which will later be provided to `StreamingDataFrame`.

6. If the message is processed successfully, the store partitions will be flushed, and the partition offset will be committed to Kafka.

### DIFF
--- a/src/StreamingDataFrames/streamingdataframes/kafka/consumer.py
+++ b/src/StreamingDataFrames/streamingdataframes/kafka/consumer.py
@@ -11,6 +11,8 @@ from confluent_kafka import (
 )
 from confluent_kafka.admin import ClusterMetadata
 
+from streamingdataframes.exceptions import PartitionAssignmentError, KafkaPartitionError
+
 __all__ = (
     "Consumer",
     "AutoOffsetReset",
@@ -45,6 +47,22 @@ def _default_on_commit_cb(
         )
     if on_commit is not None:
         on_commit(error, partitions)
+
+
+def _wrap_assignment_errors(func):
+    """
+    Wrap exceptions raised from "on_assign", "on_revoke" and "on_lost" callbacks
+    into `PartitionAssignmentError`
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as exc:
+            raise PartitionAssignmentError("Error during partition assignment") from exc
+
+    return wrapper
 
 
 class Consumer:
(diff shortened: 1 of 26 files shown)


src/StreamingDataFrames/streamingdataframes/kafka/consumer.py
```diff
@@ -158,30 +176,54 @@ def subscribe(
...
                     "Assigned partition to a consumer",
                     extra={"topic": partition.topic, "partition": partition.partition},
                 )
             if on_assign is not None:
                 on_assign(consumer, partitions)
 
+        @_wrap_assignment_errors
         def _on_revoke_wrapper(consumer: Consumer, partitions: List[TopicPartition]):
             for partition in partitions:
+                if partition.error:
```
tim-quix, line 199: Just out of curiosity, have you seen what any of these errors look like? Curious what causes them

tim-quix (approved):
Don't really have a lot to add: overall it's well organized, the operations are easy to understand and generally functionality is straight-forward, and your supplemental documentation helps to follow how it all intertwines. 

It might be worthwhile taking what you had in Notion and adding it here somewhere (like the diagram and general breakdown of the hierarchy, plus some clarifications that came from our chat)? Maybe that's really a task for the "documentation" portion of things.

In the end there might be some things that we could combine/refactor so we dont have so many objects so manage, but it's probably not gonna make enough of a difference for the time it would take, and really it would likely be subjective changes anyway. Plus we are still adding new functionality and so things could change from that anyway!

Anywho, well done, good foundation =). LGTM

src/StreamingDataFrames/streamingdataframes/kafka/consumer.py
ghost: Not really, to be honest. I tried to find an easy way to simulate them but to no avail.
My best guess is they might happen if something went wrong on the broker, e.g. the partition is failed to replicate or so.

The `TopicPartition` object is also not mockable with `unittest.mock` - it's immutable, and you cannot set any attribute to it.

PR #205, opened 2023-10-25
Implement stateful functions for StreamingDataFrames

ghost:
- Set default state dir to `state`
- Add `stateful` parameter to `StreamingDataFrame.apply()`
- Add TransactionState class that implements a very limited key-value storage interface to be passed to stateful functions


src/StreamingDataFrames/tests/test_dataframes/test_dataframe/test_dataframe.py
```diff
@@ -273,3 +275,73 @@ def test_to_topic_no_producer_assigned(self, dataframe_factory, row_factory):
...
                 )
             )
+
+
+class TestDataframeStateful:
+    def test_apply_stateful(self, dataframe_factory, state_manager, row_factory):
+        topic = Topic("test")
+
+        def stateful_func(value, state: State):
+            current_max = state.get("max")
```
tim-quix, line 285: It wasn't until this example that I think I understand what you meant by prefixing the key, lol.

Am I correct in understanding this is how it works structurally:

- folder: consumer_group
  - folder: topic_T
    - folder: partition_N
      - RDB file: "default"
        - kafka msg key: `key_X`
          - stored key: `max` (via `key_X-max`) = `10`
        - kafka msg key `key_Y`
          - stored key `max` (via `key_Y-max`) = `5`
      - RDB file: "rolling10s"
        - kafka msg key: `key_X`
          - mean?

This could be my own biases here based on how we implemented state with Fluvii, but I feel like I'd prefer to just store state without the need to define keys with it (i.e. `max`) since you can store whatever object you want. The only upside I can see is if you only need a certain part of the state, you can potentially load only what you need...but you could also have way more keys to counteract that.

We can discuss it more if my understanding is correct.

I'll wait to confirm my understanding is correct though.

src/StreamingDataFrames/streamingdataframes/app.py
```diff
@@ -388,8 +380,8 @@ def run(
                     first_row.offset,
                 )
 
-                if self.is_stateful:
-                    # Store manager has stores registered, starting a transaction
+                if self._state_manager.stores:
```
tim-quix: Can we do this check once before entering the run loop to determine which approach we should use? 

Also, I'm assuming this is only the case when SDF registers a stateful function, right? Just making sure I'm following

src/StreamingDataFrames/tests/test_dataframes/test_dataframe/test_dataframe.py
ghost: Yes, you're correct.
- The actual DB dir is `<consumer_group>/<topic>/<partition>/<store_name>`
- The actual keys inside the store are prefixed with the message key. 

> The only upside I can see is if you only need a certain part of the state, you can potentially load only what you need...but you could also have way more keys to counteract that.

There's another upside - you can update only what you need. 
If you store the whole object, on each update you need to read and update the whole object too.

src/StreamingDataFrames/streamingdataframes/app.py
ghost: Yes, here we check if SDF requires any stores before processing. 
If there are no stores, it's a stateless SDF  and we don't need to manage any state partitions, etc.

Technically, you can do `if self._state_manager.stores:` check before the loop, but you still would need to have some `if` before starting each transaction because it requires topic-partition-offset

src/StreamingDataFrames/tests/test_dataframes/test_dataframe/test_dataframe.py
tim-quix: Also true, yeah!

Not sure which is better overall...it's probably just a preference, or depends on how many of these keys you store. I suppose nothing is stopping you from storing a big json or something if you really needed to (would still have to specify a key and loads/dumps but whatev).

I think as long as the documentation is clear in how it works, it probably doesn't matter.

ghost: I think the approach with individual keys gives more flexibility, especially when you need to update the state multiple times during a single function.
Also, it simplifies managing the state when you have multiple stateful functions in the pipeline.

src/StreamingDataFrames/streamingdataframes/app.py
tim-quix: Right, I was thinking about something like:

```
### before loop
if stores:
    t_maker = self._state_manager.start_store_transaction
else:
    t_maker = lambda *args: contextlib.nullcontext()

### in loop:
state_transaction = t_maker(t, p, o)
```

TBH not sure if a function call or if statement is faster though...probably an if statement, which would nullify my suggestion.

tim-quix:
Looks good to me and ready to merge

src/StreamingDataFrames/streamingdataframes/app.py
ghost: Oh, I see. I like this idea.
It's indeed a bit faster because we can re-use the same `nullcontext` object many times.

I pushed the change

PR #208, opened 2023-10-26
Quix legacy SERDES support for SDF

tim-quix:
# Legacy Quix SerDes additions
Added/changed several things around serdes to accomodate parsing of Quix legacy (aka pre-header) formats (`quixstreams` <0.6.0) in a fully backwards-compatible manner.

## Quix Deserialization 
- Deserialization now has one universal class, `QuixDeserialzer`
  - other Quix deserializer classes removed/consolidated
- You can now handle multiple quix message types upon consume instead of just 1
  - Includes `EventData`, `EventData[]`,  and`TimeSeries`/`ParameterData` 
  - Ignore list remains the same
  - If your app needs access to different messages in different formats, you will need to do some sort of conversion with SDF manually to properly convert them to the same data structure post-deserialization
- Legacy formats of `ParameterData`, `EventData`, and `EventData[]` are now supported (<0.6.0) and fully backwards-compatible
  - "Split" messages remain NOT supported
  - Messages with multiple "rows" (lists of entries) are still parsed/handled as individual rows (just like the current SDF format).
  
  
## Quix Serialization
- Added two new "Legacy" serializers, `QuixLegacyTimeseriesSerializer` and `QuixLegacyEventSerializer`

### DIFF
--- a/src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
+++ b/src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
@@ -83,8 +83,8 @@ class QuixDeserializer(JSONDeserializer):
             column_name=column_name, loads=loads, loads_kwargs=loads_kwargs
         )
         self._deserializers = {
-            QModelKey.TIMESERIESDATA: self.deserialize_timestamp,
-            QModelKey.PARAMETERDATA: self.deserialize_timestamp,
+            QModelKey.TIMESERIESDATA: self.deserialize_timeseries,
+            QModelKey.PARAMETERDATA: self.deserialize_timeseries,
             QModelKey.EVENTDATA: self.deserialize_event_data,
             QModelKey.EVENTDATA_LIST: self.deserialize_event_data_list,
         }
@@ -98,7 +98,7 @@ class QuixDeserializer(JSONDeserializer):
         """
         return True
 
-    def deserialize_timestamp(
+    def deserialize_timeseries(
         self, value: Union[List[Mapping], Mapping]
     ) -> Iterable[Mapping]:
         if not isinstance(value, Mapping):
@@ -180,10 +180,10 @@ class QuixDeserializer(JSONDeserializer):
         # Warning: headers can have multiple values for the same key, but in context
         # of this function it doesn't matter because we're looking for specific keys.
         if ctx.headers is None:
-            legacy = True
+            as_legacy = True
             headers_dict, value = self._parse_legacy_format(value)
         else:
-            legacy = False
+            as_legacy = False
             headers_dict = {key: value.decode() for key, value in ctx.headers}
         # Fail if the message is a part of a split
         if Q_SPLITMESSAGEID_NAME in headers_dict:
@@ -224,7 +224,7 @@ class QuixDeserializer(JSONDeserializer):
         # Parse JSON from the message value
         try:
             deserialized = (
-                self._loads(value, **self._loads_kwargs) if not legacy else value
+                self._loads(value, **self._loads_kwargs) if not as_legacy else value
             )
         except (ValueError, TypeError) as exc:
             raise SerializationError(str(exc)) from exc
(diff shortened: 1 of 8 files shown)


src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -54,12 +51,44 @@ class QCodecId:
...
+            dictionary with `column_name` as a key.
+        :param loads: function to parse json from bytes. Default - `json.loads`.
+        :param loads_kwargs: dict with named arguments for `loads` function.
+        """
+        super().__init__(
+            column_name=column_name, loads=loads, loads_kwargs=loads_kwargs
+        )
+        self._deserializers = {
+            QModelKey.TIMESERIESDATA: self.deserialize_timestamp,
+            QModelKey.PARAMETERDATA: self.deserialize_timestamp,
```
ghost: Maybe call it `deserialize_timeseries`?

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -94,14 +177,14 @@ def __call__(self, value: bytes, ctx: SerializationContext) -> Iterable[dict]:
...
 
-        # Validate message headers
         # Warning: headers can have multiple values for the same key, but in context
         # of this function it doesn't matter because we're looking for specific keys.
-        headers_dict = (
-            {key: value.decode() for key, value in ctx.headers}
-            if ctx.headers is not None
-            else {}
-        )
+        if ctx.headers is None:
```
ghost: `ctx.headers` could be `not None`, but miss the required headers, and still be in "legacy" format.

If we want to parse legacy and header-based formats with the same class, let's make the check more robust and validate that at least one of the required headers is present before setting the `legacy = True` flag.

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -407,3 +418,8 @@ def __call__(
         }
 
         return self._to_json(result)
+
+
+class QuixLegacyEventsSerializer(QuixEventsSerializer, QuixLegacySerializer):
```
ghost: If we still deserialize legacy and new format with the same class, could we maybe do the same with serialization?

E.g. we could have one `QuixEventsSerializer` class with the flag `legacy: bool` in the constructor and go from there. 
Same for `QuixTimeseriesSerializer`.

ghost (changes requested):

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: I changed the approach to inspect the raw message and see if "C" and "K" exist at the beginning of the message

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: lol, IDK how I accidently called it that. Must have had timestamps on the brain =)

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: Changed

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -132,91 +213,96 @@ def __call__(self, value: bytes, ctx: SerializationContext) -> Iterable[dict]:
...
 
-        # Iterate over timestamps and get a corresponding parameter value
-        # for each timestamp
-        for timestamp_ns in timestamps:
-            row_value = {
-                param: _b64_decode_or_none(next(values)) if is_binary else next(values)
-                for param, values, is_binary in all_params
+    def __init__(
+        self,
+        as_legacy: Optional[bool] = None,
```
ghost: Why to use `Optional[bool]` here when it can be just `True` or `False`?

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -239,12 +325,31 @@ class QuixTimeseriesSerializer(JSONSerializer):
...
 
-    # Quix Timeseries data must have the following headers set
     extra_headers = {
         QModelKey.HEADER_NAME: QModelKey.TIMESERIESDATA,
         QCodecId.HEADER_NAME: QCodecId.JSON_TYPED,
     }
 
+    def __init__(
+        self,
+        as_legacy: Optional[bool] = None,
```
ghost: Similar, here looks like it can be `as_legacy: bool = True` vs `None` and extra check in __init__

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -94,20 +177,18 @@ def __call__(self, value: bytes, ctx: SerializationContext) -> Iterable[dict]:
...
-            if ctx.headers is not None
-            else {}
-        )
-        # Fail if the message is a part of a split
-        if Q_SPLITMESSAGEID_NAME in headers_dict:
-            raise SerializationError(
-                f'Detected "{Q_SPLITMESSAGEID_NAME}" header but message splitting '
-                f"is not supported"
-            )
+        if as_legacy := self._is_legacy_format(value):
```
ghost, line 182: Is the `:=` operator required here? 
The `as_legacy` variable doesn't seem to be used anywhere else. 

It can be replaced with `if self._is_legacy_format(value):`

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
```diff
@@ -132,91 +213,96 @@ def __call__(self, value: bytes, ctx: SerializationContext) -> Iterable[dict]:
...
+        if self.as_legacy:
+            self._legacy = {
+                QModelKey.HEADER_NAME: QModelKey.PARAMETERDATA,
+                QCodecId.HEADER_NAME: QCodecId.JSON_TYPED,
+            }
+        else:
+            self.extra_headers = {
+                QModelKey.HEADER_NAME: QModelKey.TIMESERIESDATA,
+                QCodecId.HEADER_NAME: QCodecId.JSON_TYPED,
             }
```
ghost: Could you please clarify how does this part play with the descendant classes?

I see it changes `self._as_legacy` here, but the descendants also [change it](https://github.com/quixio/quix-streams/pull/208/files#diff-432db6a0e20f9f6a1b464971827eef744d306f48c8adcc1b0b3a7b207c0b8de3R345-R352)


Looks like this base class can be replaced by a simple function `_as_quix_legacy(value: bytes, model_key: str,  codec_id: str)` that will wrap the serialized JSON into some envelope

ghost (changes requested):

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: It is used on line 225

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: Yeah true, I can def do that cleaner.

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
ghost: Oh ok, it didn't notice it so far away :)

src/StreamingDataFrames/streamingdataframes/models/serializers/quix.py
tim-quix: I have cleaned up the approach (and removed stuff that I thought I cleaned up, sorry, I should have double-checked)...wanted you to feel Halloween in code form apparently, haha!!

While I probably could remove the inheritance if needed, please take a look at my latest commit and see if you are happy with my improved approach...i think it should make a lot more sense now.

Basically, I still have a `QuixSerializer` base class for the sake of not repeating stuff across both the `Timeseries` and `EventData` serializers.

Also, just in case you were wondering what I was doing with the headers, I am basically making sure I do not include the `extra_headers` if it's in the legacy format (in case that wasn't clear). I want to replicate the original format as closely as possible so the C# client doesn't know any different.

PR #211, opened 2023-10-30
Validate state settings for Quix apps

ghost:
- Check if the "State Management" flag is enabled if the app is deployed on Quix
- Check if the app's "state_dir" points to the correct location if deployed on Quix
- Add QuixEnvironment singleton to access Quix-specific env vars


tim-quix (approved):
Looks good

PR #212, opened 2023-10-30
Pass MessageContext as a second argument to ".apply()" functions for SDF and Column

ghost:
# Changes 

**1. Introduce a new class - `MessageContext` - for keeping Kafka message props outside of `Row`**

The message metadata is now packed into `MessageContext`.
`MessageContext` is supposed to be a constant object for all Rows in the given Kafka message.
It is pseudo-immutable (public properties don't have setters).

It gives us a couple of bonuses:
- We can simplify `Row.clone` method a lot because only value needs to be copied now, and MessageContext is supposed to stay unchanged.
- We don't need to copy the same values for each Row with `QuixDeserializer` - we can create MessageContext once and pass a reference to it for each row.

And there are some gotchas. Both of them are present now too, so we might need to address them in the future
- Message headers are parsed as either a list of tuples or `dict`, and both data structures can be mutated.\
- Message keys can also be deserialized to mutable data structures. And while `Row.key` or `MessageContext.key` props cannot be overwritten, the key itself can be.

**2. Pass `MessageContext` as a second argument to `.apply()` functions in StreamingDataFrame and Column**

Each function passed to `.apply()` now must accept at least 2 arguments.

### DIFF
--- a/src/StreamingDataFrames/examples/bank_example/json_version/consumer.py
+++ b/src/StreamingDataFrames/examples/bank_example/json_version/consumer.py
@@ -1,12 +1,22 @@
+"""
+An application to process imaginary purchase transactions in real-time using Kafka
+
+In this application, we will simulate notifications for "Gold" accounts about
+purchase events larger than $1000
+"""
+
+
 from os import environ
 
 from dotenv import load_dotenv
 
-from streamingdataframes import Application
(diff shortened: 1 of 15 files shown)


tim-quix (approved):
Looks good!

PR #221, opened 2023-11-03
allow string param arguments for defining serializers

tim-quix:


src/StreamingDataFrames/streamingdataframes/models/topics.py
```diff
@@ -58,6 +64,18 @@ def name(self) -> str:
...
+    def get_serializer(serializer: SerializerType):
+        if isinstance(serializer, str):
+            return SERIALIZERS[serializer]()
+        return serializer
+
+    @staticmethod
+    def get_deserializer(deserializer: SerializerType):
+        if isinstance(deserializer, str):
+            return DESERIALIZERS[deserializer]()
+        return deserializer
```
ghost: Hey!
I like the approach overall, but there are several things that can be improved:

1. We don't expect these funcs to be called outside of this class, so we don't need to expose this API. Let's move these methods from the class to simple private functions in the module like `_get_serializer`. 
2. The `None` type is not handled in typings. Perhaps we should not accept `None` in the `.topic()` func anymore since there's now a default value.
3. `get_serializer()` & `get_deserializer()` should specify the return type
4. Let's raise a nice ValueError in case the passed string is not in the dictionary. Right now it will fail with KeyError. Something like `Unknown serialization format "blabla". Should be one of "str", "json"  etc`

ghost (changes requested):

ghost (approved):
Looks good 👍

PR #223, opened 2023-11-06
Temporarily remove "expand=True" feature

ghost:
It breaks the SDF if called >1 times in a row


tim-quix (approved):
I think you can also remove the `more_rows_func` from `test_dataframe/fixtures`

PR #224, opened 2023-11-06
Add a failing test for `expand=True`

ghost:
It breaks the SDF if called >1 times in a row


ghost:
Fixed in https://github.com/quixio/quix-streams/pull/246

PR #232, opened 2023-11-07
create interface to access row context from sdf

tim-quix:
Adds a new attribute, `StreamingDataFrame.context`, which generates a object called `ColumnContext`.

`ColumnContext` is a very simple object that basically inherits from `MessageContext` and gives you access to its methods (auto-complete), returning a Column in its place when calling the `MessageContext` methods like `.key`, `.timestamp`, etc. 

It also clones the values where appropriate (timestamp, headers if dict)

So, end result could be something like  

```python
sdf["new_column"] = sdf.context.key + sdf["my_append_value"]
```

One thing to consider would be to not have it be a property, but a function i.e. context(), to better designate it's not some attribute of the dataframe itself, or name it "message_context" or something. I used the name `context` for consistency in naming from `.apply` and such, in case that wasn't obvious =)

<br><br>
An alternative approach would be to do something like this on the SDF for each item:

```python
@staticmethod
def context_key():
    return Column(_eval_func=lambda row: row.context.key)
```

but I thought the former was a lot cleaner overall in terms of using the same objects we already have.

### DIFF
--- a/src/StreamingDataFrames/quixstreams/dataframe/column.py
+++ b/src/StreamingDataFrames/quixstreams/dataframe/column.py
@@ -1,12 +1,11 @@
 import operator
-from typing import Optional, Any, Callable, Container
+from copy import deepcopy
 
+from typing import Optional, Any, Callable, Container
 from typing_extensions import Self, TypeAlias, Union
 
 from ..models import Row, MessageContext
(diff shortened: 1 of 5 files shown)


tim-quix:
Closing this since Daniil's latest MR around the new Stream renders this largely irrelevant

PR #235, opened 2023-11-09
Add sdf.contains() method to check if column exists in a message

harisbotic:
## Why?
Currently there's no way to check if the key exists in a message in SDF

## What?
Add sdf.contains("key") method that can be passed to SDF as a filter.

Usage: 

```python
# Add a new column 'has_column' that will contain either True or False whether a 'column_x' exists or not in a message
sdf['has_column'] = sdf.contains('column_x')
```
or

```python
# Filter out messages that don't contain 'column_x'
sdf[sdf.contains('column_x')]
```

### DIFF
--- a/src/StreamingDataFrames/quixstreams/dataframe/dataframe.py
+++ b/src/StreamingDataFrames/quixstreams/dataframe/dataframe.py
@@ -238,10 +238,10 @@ class StreamingDataFrame:
     @staticmethod
     def contains(key: str) -> Column:
         """
-        Returns a Column configured to assess whether a specified key is present in rows.
+        Check if the key is present in the Row value.
 
-        :param key: A string representing the column name to check for in each row.
-        :return: A Column object configured to evaluate the presence of a column.
+        :param key: a column name to check.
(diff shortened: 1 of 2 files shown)


src/StreamingDataFrames/quixstreams/dataframe/dataframe.py
```diff
@@ -235,6 +235,22 @@ def producer(self) -> RowProducerProto:
     def producer(self, producer: RowProducerProto):
         self._real_producer = producer
 
+    @staticmethod
+    def contains(key: str) -> Column:
+        """
+        Returns a Column configured to assess whether a specified key is present in rows.
+
+        :param key: A string representing the column name to check for in each row.
+        :return: A Column object configured to evaluate the presence of a column.
```
ghost: Could we maybe re-phrase it a bit to make it shorter?

Like:
"
Check if the key is present in the `Row` value.

:param key: a column name to check
:returns: a `Column` object that evaluates to `True` if the key is present or `False` otherwise.
"

The examples are great 👍

ghost (approved):
Looks good 👍

PR #238, opened 2023-11-13
Allow StreamingDataFrame.apply() to be assigned to keys and used as a filter

ghost:
This PR contains many changes to the whole `quixstreams`, so read on to learn more

# Changes

## A new `Stream` class to schedule functions in SDF and Series
`Stream` is a graph-like structure to schedule functions in StreamingDataFrames and StreamingSeries.
Each scheduled function creates a new `Stream` object with a reference to its parent, so we can track the relationship between functions this way.

`Stream` allows to execute 3 types of functions:
- "apply" - generate new values based on input. The result is passed to downstream functions
- "update" - mutate values in place. The results are ignored and the input value is passed downstream
- "filter" - filter values based on a condition.  The result is interpreted as a boolean, and the input is passed downstream only if the result is `True`.

## Changes to `StreamingDataFrame.apply()` and new methods `StreamingDataFrame.update()` & `StreamingDataFrame.filter()`

- All these functions are now expected to **accept only one argument** - value (except stateful functions, they accept "value" and "state").

- There are no limitations to the value types anymore in `.apply()`. 
It can now accept and return any type of values (previously only `dict`s were supported).
It makes processing and producing simple types (ints, strings, etc) much easier.

- These functions now always return a new instance of `StreamingDataFrame` similar to `pandas`. Previously, `.apply()`  could be used to mutate existing `StreamingDataFrame` instances.

- `StreamingDataFrame.apply()` - should always return a new value without mutation. 
Though mutation is still possible, it's discouraged, and `.update()` should be used for it now.

- `StreamingDataFrame.update()` should be used to mutate values in place instead of `.apply()`.

- `StreamingDataFrame.filter()` should be used to filter values based on condition.

##  `MessageContext` is now available globally
`MessageContext` is not passed to the functions anymore, and it is now globally available for the whole app via `contextvars`. 
- To get a current message key use `quixstreams.context.message_key()` 
- To get the whole context with topic, partition, etc. use `quixstreams.context.message_context()`
- The `Application` class is responsible for setting `MessageContext` before passing values to `StreamingDataFrame`/ 

## `sdf[sdf.apply(lambda v: v < 0)]`  and `sdf['x'] = sdf.apply(lambda v: v < 0) ` now work!

`StreamingDataFrame.apply()` can now be used to filter values or to set new columns.

There are a couple of rules that they must follow in this case, though:
1. SDF passed to another SDF cannot contain "filter" functions
2. SDF passed to another SDF cannot contain "update" functions

Example:
```
sdf = StreamingDataFrame()

# This will work 
sdf = sdf[sdf.apply(lambda v: v < 0)]

# This will fail with ValueError
sdf = sdf[sdf.update(lambda v: v.append(1)]

# This will fail with ValueError too
sdf = sdf[sdf.filter(lambda v: v.append(1)]
```

## The logical operators `|`, `~`  and `&` now always do logical comparisons.

 Previously it worked this way only with booleans, when for numbers it would perform a bitwise comparison. Even though it's different from how `pandas` does it, I believe it simplifies the interface and reduces the chance for mistakes. 
 To do a true bitwise operation, please use `StreamingDataFrame.apply()` or `StreamingSeries.apply()`  with a custom function.

## The whole `quixstreams` processes messages on average 15-20% faster
The `Stream` class converts a list of scheduled functions to a single big function closure.
It runs much faster than executing functions one-by-one in a list like`Pipeline` does.


tim-quix (approved):
Looks good; will definitely need to update the docs and such and obviously remove/replace the old stuff, but that can be done in a separate MR or feel free to do it on this one. 

Really nice work, was easy to follow everything!

PR #239, opened 2023-11-13
Add a method to clear the application local state

harisbotic:
Add Application.clear_state() method to clear the local state directory before the process starts.

Usage

```python
app.clear_state() # Clears the state of the application
app.run()
```

### DIFF
--- a/src/StreamingDataFrames/docs/stateful-processing.md
+++ b/src/StreamingDataFrames/docs/stateful-processing.md
@@ -72,13 +72,13 @@ Currently, only functions passed to `StreamingDataFrame.apply()` may use State.
 
 ## Clearing the State
 
-To clear all the state data, use the `app.clear_state()` command. This will delete all data stored in the state, allowing you to start from a clean slate:
+To clear all the state data, use the `app.clear_state()` command. This will delete all data stored in the state stores, allowing you to start from a clean slate:
 
 ```python
 app.clear_state()
 ```
 
-Note that clearing the app state using `app.clear_state()` is only possible when the `app.run()` is not running. Meaning that the state should be cleared either before calling `app.run()` or after.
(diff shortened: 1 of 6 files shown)


ghost:
Hey @harisbotic , do I understand correctly that this `app.clear_state()` will work only after the StreamingDataFrame is defined?
Wouldn't it be easier to clean up the whole state directory for the given consumer group?

harisbotic:
> Hey @harisbotic , do I understand correctly that this `app.clear_state()` will work only after the StreamingDataFrame is defined? Wouldn't it be easier to clean up the whole state directory for the given consumer group?

Actually the PR is not ready yet, when I added a test that tried to run `app.clear_state()` it failed to do so because it deletes the databases but doesn't re-create them. Which by current design should happen.  So I am fixing that first.

Regarding you comment, yes, we could do that as well, I didn't think of that

harisbotic:
If we ever want to clear a single partition, here is how to do it:

```python
    def clear(self):
        """
        Clear all data from the store
        """
        logger.debug(
            f'Clearing store on: "{self._path}"',
        )

        self.close()
        self.destroy(self._path)
        self._db = self._init_db()

        logger.debug(
            f'Successfully cleared store on: "{self._path}"',
        )
        
```

with test:

```python
    def test_clear(self, rocksdb_partition_factory):
        with rocksdb_partition_factory() as storage:
            with storage.begin() as tx:
                tx.set("key", "value")

            assert storage.exists(serialize_key("key", dumps=dumps))
            storage.clear()
            assert not storage.exists(serialize_key("key", dumps=dumps))
```

ghost:
Looks good! I like the approach, it's safer to check if the store has partitions assigned.
Could you please update the docs here too?
https://github.com/quixio/quix-streams/blob/main/src/StreamingDataFrames/docs/stateful-processing.md

ghost (approved):
LGTM 👍

PR #257, opened 2023-12-04
Implement Tumbling and Hopping windows

ghost:
# What changed
Implemented stateful windowed aggregations based on Tumbling and Hopping windows.

- Tumbling windows divide time into even non-overlapping periods. 
Example: tumbling window of size `10` will generate the following periods: [0, 10), [10, 20), [20, 30), etc.
- Hopping windows divide time into even overlapping periods. The overlap is defined by step.
Example: hopping window of size `10` and step `5` will generate the following periods: [0, 10), [5, 15), [10, 20), etc.

The windowed aggregations use a separate state store - `WindowedRocksDBStore` - that stores data for fixed-time windows and can expire windows from the state based on event time.


quixstreams/dataframe/windows/definitions.py
```diff
@@ -0,0 +1,350 @@
...
+        self._name = name
+        self._step_ms = step_ms
+
+        if self._duration_ms <= 0:
+            raise ValueError(
+                f"Window duration must be bigger than zero, got {self._duration_ms}"
+            )
+
+        if self._grace_ms < 0:
+            raise ValueError(f"Window grace must be positive, got {self._grace_ms}")
```
harisbotic: these are duplicated checks. They can be removed

quixstreams/dataframe/windows/definitions.py
```diff
@@ -0,0 +1,350 @@
...
+
+        if self._step_ms is not None and (
+            self._step_ms <= 0 or self._step_ms >= duration_ms
+        ):
+            raise ValueError(
+                f"Window step size must be smaller than duration and bigger than zero, "
```
harisbotic: I would say `bigger than 0ms`

quixstreams/dataframe/windows/definitions.py
```diff
@@ -0,0 +1,350 @@
...
+                f"got {self._step_ms}"
```
harisbotic: because we are not just accepting straight ms values but also timedelta maybe we should change this to `got {x_ms} ms` (add ms at end).
If we do this, we should do it for all other checks in this method

quixstreams/dataframe/windows/definitions.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+
+    def reduce(
+        self, reducer: Callable[[Any, Any], Any], initializer: Callable[[Any], Any]
+    ) -> "FixedTimeWindow":
+        """
+        Configure the window to perform a custom aggregation using `reducer`
+        and `initializer` functions.
+
+        Example:
```
harisbotic: In some places we say `Example:` in some `Example usage:`. I think we should stick to one across the codebase. I know in the docs build.py script, we have code that is putting some special formating on `Example usage:`

quixstreams/dataframe/windows/time_based.py
```diff
@@ -0,0 +1,196 @@
...
+        latest_timestamp = state.get_latest_timestamp()
+        ranges = get_window_ranges(
+            timestamp_ms=timestamp_ms,
+            duration_ms=duration_ms,
+            step_ms=self._step_ms,
+        )
+
+        updated_windows = []
+        for start, end in ranges:
+            # TODO: Log when the value is late and no window is updated
```
harisbotic, line 70: When are we going to do this TODO

harisbotic (commented):
50% reviewed so far

harisbotic (approved):

quixstreams/dataframe/windows/definitions.py
ghost: Agree, updated

quixstreams/dataframe/windows/definitions.py
ghost: updated

quixstreams/dataframe/windows/definitions.py
ghost: We should use `Example Snippet:\n` actually, yes

quixstreams/dataframe/windows/time_based.py
ghost: In one of the next releases.

PR #258, opened 2023-12-07
changelog topics and topic management refactor

tim-quix:
## Scope/Goal:

The outcome of this PR is:

1. Consolidating topic management (especially quix vs non-quix topics)
2. Adding the auto-creation of changelog topics (made easier by the above)
3. Adding auto topic creation for non-quix (and work the same way as quix)
4. adding/consolidating topic validation approach.

#### Organization-related

There were numerous things shuffled around and cleaned up as part of this, most
of which was consolidated to below:

- New `TopicConfig` object
  - stores all topic attributes (partition, rep_factor) and optional 
    configs necessary to create a topic.
  - Typically stored on `Topic.configs` but is used in other spots

- New `Admin` class
  - handles "admin"-level requests like getting topic metadata from Kafka
  - handles creation requests for non-quix topics
  - generally used by the `TopicManager` class but can be used stand-alone

- New `TopicManager` class
  - The core of most changes
  - by default, handles non-Quix topics.
  - has all the proper defaults for a topic and changelog
  - stores all `Topic` objects (external and internal "changelog")
    - divides "topics" from "changelogs"
  - has simple interface for creating these `Topics` from the Application (or other sources)
    - Can be used directly to make generating topics without an Application easier.
  - Has a "Quix" extension called `QuixTopicManager`
    - called the same way as `Application.Quix()` (`TopicManager.Quix()`)
    - has defaults and functions adjusted accordingly (uses Quix API to generate topics)

#### Changelog-related

- `StateStoreManager`
  - Now uses `TopicManager` to generate changelog `Topic` objects
    - `TopicManager` passed from `Application`
    - If changelogs are disabled, will not be passed a `TopicManager`

## Intention behind organizational changes

In current state, managing Quix vs non-Quix topics is a bit of a pain, 
especially since they have different default configurations, and are not
generated in the same fashion (there's currently no sanctioned way to 
auto-create topics for a non-Quix `Application` with desired settings).

This matter is only further compounded with managing "changelog" topics, which
are intended to largely be "behind-the-scenes" and have strict requirements
around how they are configured (including dependencies on other topics!)

To not have users have to manage this no matter their platform, I went ahead 
and implemented a more generalized approach to topic creation that keeps all 
functionality centralized to one primary class regardless of quix vs non-quix,
the `TopicManager`.

## Intention behind "changelog" topic additions

With the above implemented, changelog-related work was minimal. 

Added the ability to turn off changelog topic usage from the `Application` 
itself, which is just handed to `StateStoreManager`.


quixstreams/models/topics.py
```diff
@@ -28,7 +33,79 @@
...
+
+    def _as_quix_create(self) -> Self:
+        new = deepcopy(self)
+        new._filter_quix_optionals()
+        new._set_quix_optional_defaults()
+        return new
+
+    def _as_create(self) -> NewTopic:
+        if self.is_quix_topic:
+            raise Exception(
```
ghost: The exception class must be specific. Raising plain `Exception` is not a good practice

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,195 @@
...
+        source_topic_name: str,
+        suffix: str,
+        configs_to_import: tuple = ("retention.bytes", "retention.ms"),
+    ) -> Topic:
+        topic_configs = (
+            self.inspect_topics([source_topic_name])[source_topic_name]
+            or self._topics[source_topic_name].kafka_configs
+        )
+        name = f"__changelog--{self._consumer_group}--{source_topic_name}--{suffix}"
+        base_optionals = {} if self.has_quix_builder else {"cleanup.policy": "compact"}
```
ghost: I'd consider expanding "optionals" into the individual named parameters if it's possible.
(In other places too)

If not, let's rename it to `extra` like it's done in Producer and Consumer classes

ghost (changes requested):
Hey Tim, the overall approach looks good so far, but I'd change some details:

**Naming:**
- `TopicAdmin` -> `TopicManager`. "Manager" seems like a better description of what's going on in the class
- `_admin` -> `_admin_client` - the property returns `AdminClient`
- `TopicKafkaConfigs` -> `TopicConfig`

**Composition:**
- It seems to me we need a common interface for the TopicManager, but two different implementations: for Quix and non-Quix apps. 
There are too many `if`'s here and there handling different logic, and probably more will come in the future (e.g. different error handling). 
Having one big class will also complicate testing and debugging.

Also, see other comments

tim-quix:
The code has significantly changed since it was last reviewed, so closing this. As such, opened a new PR, plus changed the  branch name (put all recovery-related changes under `feature/recovery/`).

PR #259, opened 2023-12-11
Add tumbling window

harisbotic:


quixstreams/dataframe/dataframe.py
```diff
@@ -334,6 +342,52 @@ def test(self, value: object, ctx: Optional[MessageContext] = None) -> Any:
         composed = self.compose()
         return context.run(composed, value)
 
+    def tumbling_window(
+        self,
+        duration: Union[float, timedelta] = 30.0,
```
ghost: Let's not set the default duration for the window. The duration must be picked manually

quixstreams/state/windows.py
```diff
@@ -0,0 +1,355 @@
...
+    def __init__(
+        self,
+        duration: float | timedelta,
+        grace: float | timedelta,
+        dataframe: StreamingDataFrame,
+        name: Optional[str] = None,
+    ):
+        self._duration = (
+            duration.total_seconds() if isinstance(duration, timedelta) else duration
+        )
```
ghost: I suggest moving the conversion part to the `StreamingDataFrame.tumbling_window()` func and accept strictly floats here.
Plus, we need to add some additional validation (e.g. that duration and grace are > 0).

quixstreams/state/windows.py
```diff
@@ -0,0 +1,355 @@
...
+
+        return TumblingWindow(
+            duration=self._duration,
+            grace=self._grace,
+            name=name,
+            func=func,
+            dataframe=self._dataframe,
+        )
+
+    def reduce(self, reduce_func: Callable[[Any, Any], Any]) -> "TumblingWindow":
```
ghost: For reduce function I think we need to have a dedicated `default` or `initializer` param similar to https://docs.python.org/3/library/functools.html#functools.reduce

Using the first window item as initializer is not exactly correct, because the expected return type of the window can be different from the incoming value type.
E.g. your incoming values are numbers, but you need to produce a dictionary with "min", "max", "sum" keys. If you simply return a number, it will not work

tests/test_quixstreams/test_state/test_windows.py
```diff
@@ -0,0 +1,152 @@
...
+    TumblingWindow,
+    WindowResult,
+    TumblingWindowDefinition,
+)
+
+
+class TestTumblingWindow:
+    @pytest.fixture
+    def state_mock(self):
+        return MagicMock(spec=WindowedTransactionState)
```
ghost: Why would we mock the state here when we can use the actual DB instance?

ghost (changes requested):

quixstreams/models/timestamps.py
```diff
@@ -10,6 +10,7 @@ class TimestampType(enum.IntEnum):
     TIMESTAMP_NOT_AVAILABLE = 0  # timestamps not supported by broker
     TIMESTAMP_CREATE_TIME = 1  # message creation time (or source / producer time)
     TIMESTAMP_LOG_APPEND_TIME = 2  # broker receive time
+    TIMESTAMP_FROM_MESSAGE = 3  # extracted from message by custom logic
```
ghost: I don't think we need the extra timestamp type here.
The custom timestamp fits the "CREATE_TIME" type.

quixstreams/models/timestamps.py
```diff
@@ -34,6 +35,10 @@ def __init__(
     def milliseconds(self) -> Optional[int]:
         return self._milliseconds
 
+    @property
+    def seconds(self) -> Optional[int]:
```
ghost: The seconds will have type `float` because we divide int by 1000

quixstreams/models/timestamps.py
```diff
@@ -34,6 +35,10 @@ def __init__(
...
+        return self._milliseconds / 1000 if self._milliseconds is not None else None
```
ghost: Can we enforce milliseconds to always be `int` so we don't need to check the type on every access?

quixstreams/models/topics.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         key_deserializer: Optional[DeserializerType] = BytesDeserializer(),
         value_serializer: Optional[SerializerType] = None,
         key_serializer: Optional[SerializerType] = BytesSerializer(),
+        timestamp_extractor: Callable[[dict], int] = None,
```
ghost: The first parameter can be of any type, not just a dictionary

ghost: Let's also pass the deserialized headers, timestamp, and `TimestampType` from the message to the extractor to make it more flexible.

quixstreams/dataframe/dataframe.py
```diff
@@ -89,9 +93,13 @@ def stream(self) -> Stream:
...
         return self._topic
 
+    @property
+    def state_manager(self) -> StateStoreManager:
+        return self._state_manager
+
     def apply(
         self,
-        func: Union[DataFrameFunc, DataFrameStatefulFunc],
+        func: Union[DataFrameFunc, DataFrameStatefulFunc, DataFrameWindowFunc],
```
ghost: I don't  think we need the `DataFrameWindowFunc` type here because there's no way the `.apply()` function will be called like this from the outside

quixstreams/windows/base.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def sum(self) -> "FixedWindow":
+        """
+        Configure the window to aggregate data by summing up values within each window period.
+
+        :return: A window configured to perform sum aggregation.
+        """
+
+        def func(start, end, timestamp, value: Any, state: WindowedTransactionState):
+            current_value = state.get_window(start=start, end=end) or 0
```
ghost: We can use `.get_window(default=0)` vs `or 0`

quixstreams/windows/base.py
```diff
@@ -0,0 +1,337 @@
...
+        ...
+
+    def sum(self) -> "FixedWindow":
+        """
+        Configure the window to aggregate data by summing up values within each window period.
+
+        :return: A window configured to perform sum aggregation.
+        """
+
+        def func(start, end, timestamp, value: Any, state: WindowedTransactionState):
```
ghost: Missing type hints

ghost (changes requested):

ghost (approved):

PR #261, opened 2024-01-05
Add API to create a pre-configured Producer and Consumer on the Application level

harisbotic:
# Problem
Configuring plain Producer and Consumer for Quix apps is a non-trivial task and currently requires a lot of cryptic boilerplate to achieve.

It's especially noticeable when users want to try out Quixstreams for the first time with some local data - you need to produce this data first to the Kafka topic, and for that you need a Quix Producer.

# Solution
Add 2 new public methods:
-  `Application.get_producer()` that initializes a disposable producer (i.e. we don't re-use the already initialized Producer, but instead create a one-time producer with the right config) that users can use when they simply need to produce the data.
- `Application.get_consumer()` which similarly creates a disposable Consumer.

It doesn't solve all of the boilerplate problems, but at least it reduces the need to configure it yourself, especially for Quix apps.

So users can produce their toy data like this:
```

app = Application.Quix(...)
topic = app.topic("input")


def main():
    with app.get_producer() as producer:
        for i in range(100):
            producer.produce(topic=topic.name, key=b"key", value=b"value")


if __name__ == "__main__":
    main()
```


quixstreams/app.py
```diff
@@ -431,6 +441,90 @@ def stop(self):
...
+            for _ in range(100):
+                msg = consumer.poll(timeout=1.0)
+                if msg is not None:
+                    # Process message
+                    # Optionally commit the offset
+                    # consumer.store_offsets(msg)
+
+        ```
+        """
+        if self.is_quix_app:
```
tim-quix, line 513: not really a thing that needs changed now, but as an FYI, topic management will be a bit easier with the new TopicManager I made in a pending MR. With it, We might even consider not creating the topics unless you run an actual application, since creating topics becomes a 1-liner with it?

quixstreams/app.py
```diff
@@ -431,6 +441,90 @@ def stop(self):
...
+        Example Snippet:
+
+        ```python
+        from quixstreams import Application
+
+        app = Application.Quix(...)
+        topic = app.topic("input")
+
+        with app.get_consumer() as consumer:
+            for _ in range(100):
```
```suggestion
            consumer.subscribe([topic.name])
            While True:
```

quixstreams/app.py
```diff
@@ -431,6 +441,90 @@ def stop(self):
...
+
+        ```python
+        from quixstreams import Application
+
+        app = Application.Quix(...)
+        topic = app.topic("input")
+
+        with app.get_producer() as producer:
+            for i in range(100):
+                producer.produce(topic=topic.name, key=b"key", value=b"value")
```
tim-quix, line 464: small QoL thing, but I feel like the user's should just get to use `Topic` objects instead of having to call `Topic.name`. Maybe we want to have a small subclass of producer with this behavior? Kind of overlaps with the `RowProducer` though, which I know is also due for an overhaul, and maybe we just use that instead?

quixstreams/app.py
tim-quix: Also, similar comment WRT using `Topic` vs `Topic.name` here.

tim-quix (approved):
Small suggestions/comments that shouldn't block the merge, but take a look and let me know your thoughts. =)

PR #264, opened 2024-01-09
Implement better handling for all broker down error

peter-quix:
- Avoid logging all broker down exception when it does not affect publishing
- Improve CONTRIBUTING guide


ghost (approved):
LGTM 👍

PR #265, opened 2024-01-10
 topic management refactor + add changelog topics

tim-quix:
# Topic Manager

## Scope/Goal:

The outcome of this PR is:

1. Consolidating topic management (especially quix vs non-quix topics)
2. Adding the auto-creation of changelog topics (made easier by the above)
3. Adding auto topic creation for non-quix (and work the same way as quix)
4. adding/consolidating topic validation approach.

#### Organization-related

There were numerous things shuffled around and cleaned up as part of this, most
of which was consolidated to below:

- New `TopicConfig` object
  - stores all topic attributes (partition, rep_factor) and optional 
    configs necessary to create a topic.
  - Typically stored on `Topic.configs` but is used in other spots

- New `Admin` class
  - handles "admin"-level requests like getting topic metadata from Kafka
  - handles creation requests for non-quix topics
  - generally used by the `TopicManager` class but can be used stand-alone

- New `TopicManager` class
  - The core of most changes
  - by default, handles non-Quix topics.
  - has all the proper defaults for a topic and changelog
  - stores all `Topic` objects (external and internal "changelog")
    - divides "topics" from "changelogs"
  - has simple interface for creating these `Topics` from the Application (or other sources)
    - Can be used directly to make generating topics without an Application easier.
  - Has a "Quix" extension called `QuixTopicManager`
    - called the same way as `Application.Quix()` (`TopicManager.Quix()`)
    - has defaults and functions adjusted accordingly (uses Quix API to generate topics)

#### Changelog-related

- `StateStoreManager`
  - Now uses `TopicManager` to generate changelog `Topic` objects
    - `TopicManager` passed from `Application`
    - If changelogs are disabled, will not be passed a `TopicManager`

## Intention behind organizational changes

In current state, managing Quix vs non-Quix topics is a bit of a pain, 
especially since they have different default configurations, and are not
generated in the same fashion (there's currently no sanctioned way to 
auto-create topics for a non-Quix `Application` with desired settings).

This matter is only further compounded with managing "changelog" topics, which
are intended to largely be "behind-the-scenes" and have strict requirements
around how they are configured (including dependencies on other topics!)

To not have users have to manage this no matter their platform, I went ahead 
and implemented a more generalized approach to topic creation that keeps all 
functionality centralized to one primary class regardless of quix vs non-quix,
the `TopicManager`.

# Recovery

## Goal

We want to add recovery behavior for stateful processing via changelog topics.

Also, it is desirable to avoid creating large/complex coupling's between State (which primarily concerns itself with local DBs) and Kafka changelogs, including making it easy to turn off changelogs if desired, and not have a bunch of if/then branching logic.

Obviously some coupling is unavoidable since it all relates, but I think this solution keeps it very minimal.

## Overview

The recovery process is facilitated by producing state changes via changelog topics, where each `Store` has a corresponding changelog topic (as a reminder, there can be multiple Stores due to windowing). The changelog topics are now the source of truth for state.

As state changes happen, they get produced to their corresponding changelog topic before being written to state.

Then, as rebalances (due to new assignments) happen during normal Application operation (including init), a recovery check is performed.

The recovery check confirms whether the recorded changelog offset in state corresponds to the actual changelog topic highwater; if not, we will pause normal `Application` operation to instead read and update the state store until it is up-to-date with said changelog. Once complete, we resume normal operation.

Also, for clarification, it's fair in general to say that "recovery" == "changelogs", as they are basically interdependent, but recovery refers to the more broad operation around using changelog topics to rebuild state.

## Changes to `Application`

In order to perform recovery, there were a few new methods added to `Application` so that it could swap between `_processing` mode and `_recovery` mode, where the current method is stored on `_run_mode`. The general looping approach has remained the same, and changes were overall fairly minimal in this respect.

You can also choose to not use changelog topics via `use_changelog_topics=False`.

Also, the default consumer `assignment_strategy` was changed to "cooperative_sticky".

## Changes to `StateStoreManager`

The `StateStoreManager` essentially added a new attribute to store a new `ChangelogManager` instance, automatically generated unless you specified not to use changelogs. `StateStoreManager` only interacts with the `ChangelogManager` at a few points (rebalances and creating stores), so the coupling remains fairly light.

## Changes to `StorePartition`

A changelog offset is now stored on each DB.

Some other very minor changes, mostly around allowing the `ChangelogManager` to hand it `ChangelogWriters` and the `RecoveryManager` to initialize a special recovery transaction (`RocksDBPartitionRecoveryTransaction`).

You can learn about all of these below.

## NEW - `ChangelogManager`

Meant to be the primary, simple interface for all changelog-related operations (including kicking off recovery). Used for:
- generating changelog topics, 
- handling assigning/revoking of those changelogs concurrently with `StateStoreManager` adding/revoking its respective `StorePartition`s, 
- handling the `RecoveryManager`.
- generating `ChangelogWriter`s, which are used to produce messages to the changelog during a `PartitionTransaction`.

## NEW - `ChangelogWriter`

- generated by `ChangelogManagers` via `Store`s, which will hand them off to a `PartitionTransaction`
    - basically, `Store.start_partition_transaction(ChangelogWriter) -> StorePartition.begin(ChangelogWriter)`
    - this allows `StorePartition`, which solely concerns itself with DB-level management, to remain agnostic/ignorant to anything related to changelogs other than this argument handoff, 
    - writer then has access to all the information it will need to produce state changes with through the normal `PartitionTransaction` process.

## NEW - `RecoveryManager`

- kicked off with `ChangelogManager` via the `Application` anytime a new partition is assigned.
- handles all the "kafka-related" operations, including everything related to consuming messages from changelog topics and all the partition assignments/revokes necessary to do so.
- uses `RecoveryPartition`s to make tracking the recovery status of various changelog partitions easier.

NOTE: only partitions that require recovery will get their changelog partitions actually assigned in order to reduce unnecessary rebalancing overhead.

## NEW - `RocksDBPartitionRecoveryTransaction`

This is used solely by the `RecoveryManager` as an easy way to write to the DB during recovery, which has its own simplified consume/produce flow (compared to a normal transaction).

# Detailed Breakdown/Example of Recovery Flow

Lets say we have 3 partition topic named "source_topic" with two stores, "default" and "window-10s", and p[2]'s state was corrupted somehow, oh no! So we delete the store and we start our app up to recover it.

Here is the process flow for recovery.

NOTE: changelog partition assignments DO NOT trigger a consumer group rebalance!

1. (app init) `StateStoreManager` registers all the stores and changelog topics (via ChangelogManager).
2. `Application`s consume loop begins in "processing" mode:
	- it polls, triggering an assignment for "source_topic"s partitions.
3. on_assign callback is triggered, 
	- "source_topic" partitions are officially assigned
	- `StateStoreManager` assigns its StorePartitions. 
	- The `ChangelogManager` adds all the respective changelog partitions to a "pending" queue in the `RecoveryManager`.
		- `RecoveryManager` mode changed to "assign pending partitions" from "recover", for later
		- "source_topic" partitions are all PAUSED.
	- `Application` processing loop type is set to "recovery" mode
4. Resuming from the initial poll that triggered the assignment, the poll now returns None since all the partitions are paused. So, this loop iteration finishes.
5. `Application` consume loop starts next iteration, now calling "recovery" instead of "processing", which calls the `RecoveryManager`
	- `RecoveryManager` is in "assign pending partitions" mode, so it:
		- checks pending partition queue to see what actually needs recovery by checking highwaters
		- it finds only p[2] needs recovery, so it ditches the rest
		- `changelog__default` and `changelog__window-10s` are officially assigned to consumer
		- `RecoveryManager` mode changed to "recover"
6. Next consume iteration begins. `Application` still in "recovery" mode, so it calls `RecoveryManager` like before
	- `RecoveryManager` is in "recover" mode, so now it will consume a changelog message and write its contents to state
7. Step (6.) above repeats until all changelog messages are read.
	- when a changelog partition matches the highwater, it is unassigned.
	- when no more changelog partitions remain, `RecoveryManager` raises a `RecoveryComplete` exception
	- the exception is caught by `Application`, which changes its mode back to "processing"
8. loop iterations continue on via "processing" mode, consuming normal messages until the next rebalance, which would go back to Step (3.)


examples/bank_example/json_version/producer.py
```diff
@@ -6,10 +6,21 @@
...
+from quixstreams import TopicManager
 
 load_dotenv("./env_vars.env")
 
+topic_manager = TopicManager(
+    admin_client=Admin(broker_address=environ["BROKER_ADDRESS"])
+)
+topic_name = topic_manager.topic(
+    name="json__purchase_events",
+    # "config" only needed if you wish to not use the defaults!
```
harisbotic, line 17: `config` -> `extra_config`?

examples/bank_example/quix_platform_version/producer.py
```diff
@@ -5,22 +5,27 @@
...
-topic = "qts__purchase_events"
 cfg_builder = QuixKafkaConfigsBuilder()
-cfgs, topics, _ = cfg_builder.get_confluent_client_configs([topic])
-topic = topics[0]
-cfg_builder.create_topics([TopicCreationConfigs(name=topic)])
+cfgs = cfg_builder.get_confluent_broker_config()
+topic_manager = TopicManager.Quix(quix_config_builder=cfg_builder)
+topic_name = topic_manager.topic(
+    name="qts__purchase_events",
+    # "config" only needed if you wish to not use the defaults!
```
harisbotic, line 25: `config` -> `extra_config` ?

quixstreams/app.py
```diff
@@ -362,25 +406,43 @@ def topic(
...
+            value_serializer=value_serializer,
             key_deserializer=key_deserializer,
+            value_deserializer=value_deserializer,
+            config=config,
+            auto_create_config=self._auto_create_topics,
+        )
+
+    def topic_config(
+        self,
+        name: str,
```
harisbotic: what is `name` used here for?
I think we can remove it

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,191 @@
...
+import pprint
+import time
+from asyncio import Future
+from typing import List, Dict, Mapping, Optional
+
+from confluent_kafka.admin import (
+    AdminClient,
+    ConfigResource,
+    KafkaException,  # type: ignore
+    NewTopic,  # type: ignore
```
harisbotic: I tried removing `# type: ignore` comments and it seems to me like it has no effect, but I might be wrong.
If that is try can we can remove it

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,191 @@
...
+)
+
+from quixstreams.models.topics import TopicConfig, TopicList
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("Admin",)
+
+
+def convert_topic_list(topics: TopicList) -> List[NewTopic]:
```
harisbotic: Does it make sense to import NewTopic with 
`import NewTopic as ConfluentNewTopic` or something similar, in my opinion it makes the code more readable but its really just an opinion, feel free to simply ignore this

quixstreams/platforms/quix/config.py
```diff
@@ -1,36 +1,46 @@
...
 
-@dataclasses.dataclass
-class TopicCreationConfigs:
-    name: Optional[str] = None  # Required when not created by a Quix App.
-    num_partitions: int = 2
-    replication_factor: int = 2
-    retention_bytes: int = 52428800
-    retention_minutes: int = 10080
-    optionals: Optional[Mapping] = None
+def strip_workspace_id_prefix(workspace_id, s: str) -> str:
```
harisbotic: Missing workspace_id typings

harisbotic (commented):
Good work! 
I left few comments which mainly involve some cleanup work and few suggestions which are just something to think about

quixstreams/topic_manager.py
```diff
@@ -0,0 +1,770 @@
...
+    if d:
+        if isinstance(d, dict):
+            return [i for v in d.values() for i in dict_values(v)]
+        elif isinstance(d, list):
+            return d
+        return [d]
+    return []
+
+
+def affirm_ready_for_create(topics: TopicList):
```
harisbotic, line 30: I think `validate_ready_for_create ` would be a better fit, or at least something more often used than `affirm`.

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,191 @@
...
+            raise self.CreateTopicTimeout(
+                f"Timed out waiting for creation status for topics:\n"
+                f"{pprint.pformat([topic_name for topic_name in futures])}"
+            )
+
+    def create_topics(
+        self, topics: TopicList, timeout: int = 10, finalize_timeout: int = 20
+    ):
+        """
+        Create the given list of topics and confirm they are ready.
```
harisbotic, line 167: not sure if `create_topics` is the best name here, because it can happen that some topics already exist and it will no create them.
Here are few of my suggestion, but feel free to use any name you think fits best:
`ensure_topics_created`, `setup_topics `, `initialize_topics`.

With this I would mofify the first sentence in the docs to for ex.:
   `Ensures that the specified Kafka topics are created and ready for use.`

quixstreams/topic_manager.py
```diff
@@ -0,0 +1,770 @@
...
+
+        :return: formatted topic name
+        """
+        ...
+
+    @abstractmethod
+    def _create_topics(self, topics: TopicList):
+        """
+        Method that actually creates the topics in Kafka via an `Admin` instance.
+
```
harisbotic, line 280: if we change the method name in `Admin`, we should change it here as well to `ensure_topic_created` to similar

quixstreams/topic_manager.py
```diff
@@ -0,0 +1,770 @@
...
+        value_deserializer: Optional[DeserializerType] = None,
+        key_deserializer: Optional[DeserializerType] = "bytes",
+        value_serializer: Optional[SerializerType] = None,
+        key_serializer: Optional[SerializerType] = "bytes",
+        config: Optional[TopicConfig] = None,
+        auto_create_config: bool = True,
+    ) -> Topic:
+        """
+        A convenience method for generating a `Topic`. Will use default config options
+        as dictated by the TopicManager.
```
harisbotic, line 164: It was confusing for me wether this method actually creates a topic in kafka or just "sets it up".
In one way I understand why the name is just `topic` (and `changelog_topic` for changelogs) and I am not 100% sure how to improve it but in another hand I feel like the name of a method should be a verb and should describe what it does better. I would maybe change it to `add_topic` to still keep it short. But I am not sure. `register_topic` would be nice but its kinda long. Leaving this comment to open a discussion.

tim-quix:
Remaining work as of Jan 24:

- Review changes
- Cleaning up variable/argument name consistency (i.e. `topic` vs `topic_name` vs `source_topic_name`, etc)
- adding more tests, especially at the Application level and rebalancing behavior WRT recovery
- improve logging a little
- doc strings
- documentation summarizing the recovery procedure
- cleaning up little odds and ends
- Rebasing/adding any changes necessary around windowing additions (if any)

quixstreams/app.py
tim-quix: Indeed, good catch! Residual from previous iterations.

quixstreams/kafka/admin.py
tim-quix: Annoyingly, Pycharm incorrectly believes these are not import-able, this just removes that warning in the IDE. Everything works regardless of the warning. 

If it's not a huge annoyance I'd like to leave it there, though admittedly IDK if this interacts weirdly with other IDE's.

quixstreams/kafka/admin.py
tim-quix: Nah I agree with you, I basically did a similar thing later on with Confluent's `TopicPartition`, renaming it `ConfluentPartition`.

I think I'll just rename it `ConfluentTopic` and similarly `TopicMetadata` as `ConfluentTopicMetadata`. Lemme know what you think

examples/bank_example/json_version/producer.py
tim-quix: As written is correct; the `config` (a `TopicConfig` object which itself has an `extra_config` argument) is auto-generated behind the scenes if you do not manually define it.

Perhaps I should change `extra_config` to `extra_settings` to better distinguish them, especially since they will likely be used together often?

examples/bank_example/quix_platform_version/producer.py
tim-quix: see above

quixstreams/topic_manager.py
tim-quix: I agree overall; I used this language on purpose because I basically didn't want to mix it up with `validating` the topics (aka checking they exist), which is a different step altogether. Maybe it's not really worth the trouble distinguishing it to that level?

quixstreams/kafka/admin.py
tim-quix: I agree...I thought similarly when making the function name. In the end, I didn't feel it was necessary to make a complicated name around the operation, as it usually had to be long to avoid running into similar ambiguity issues and made it kinda clunky to call.

I think as long as we are explicit in the docstring about the behavior (like you mentioned) + logging what it's doing (especially when topics already exist), its behavior should be pretty clear. I'll be sure to add those things. 

Do you think that's sufficient?

quixstreams/topic_manager.py
tim-quix: I kind of like the `register_topic` name, especially since the `TopicManager` tracks all `Topic`s generated by it, so I think it would better capture _that_ behavior. However, some may interpret that as creating the topic, as to me it gives the impression that it's generating the actual Kafka topic (which is mostly why I didn't add a verb, but I also didn't wanna mess with the existing API). 

It's all subjective, really...as evidenced by you (understandably) not being sure if it created the topic or not. IDK. I think it might just boil down to better documentation in the docstring?

tim-quix:
Note: added the PR here https://github.com/quixio/quix-streams/pull/275 that has this rebased with main (and thus windowing and etc).

Rebasing was getting a little hairy with all the PR's and rebasing I had already done, so I started a fresh one over there in case it would mess this PR's history up.

tim-quix:
Most of what's left is:

- doc strings
- adding changelog functionality to the windowing
- more tests
- converting the summary here into more official docs.
- adjusting the key deleting (prob remove the caching I was doing)

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,199 @@
...
+        self._config = {
+            "bootstrap.servers": broker_address,
+            **(extra_config or {}),
+        }
+
+    class CreateTopicTimeout(Exception):
+        ...
+
+    class CreateTopicFailure(Exception):
+        ...
```
ghost, line 71: Let's not bake exceptions in the class. Reasons:
1. It makes the class itself harder to read because there's more code
2. It makes the exceptions harder to catch outside of this class.

Let's move them out of the class to the module level.

ghost: Another thing: let's inherit the library-related exceptions from the `quixstreams.exceptions.base.QuixException` base class at least.

quixstreams/kafka/admin.py
```diff
@@ -0,0 +1,199 @@
...
+    ]
+
+
+def confluent_topic_config(topic: str) -> ConfigResource:
+    return ConfigResource(2, topic)
+
+
+class Admin:
+    """
+    For performing "admin"-level operations on a Kafka cluster, mostly around topics.
```
ghost, line 47: If the class is related to mostly topics, let's call it `TopicAdmin`

`Admin` is too generic

ghost (changes requested):

quixstreams/topic_manager.py
```diff
@@ -0,0 +1,756 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = ("TopicManager",)
+
+
+class BytesTopic(Topic):
+    def __init__(self, name: str, config: Optional[TopicConfig] = None):
+        super().__init__(
+            name=name,
+            key_serializer="bytes",
```
ghost, line 22: What is the purpose of this class? 
It doesn't seem to be different from the `Topic` class, but it still represents a different type.

ghost (changes requested):

ghost:
@tim-quix 
`topic_manager.py` should be separated into multiple files.
It has too many things inside, and I don't think they all need to be in the same file:
- A Protocol
- A base type
- 2 concrete types
- Multiple Exception types

tim-quix:
Changes suggested here have been implemented on another branch, and features have been split up as well. Closing.

PR #266, opened 2024-01-10
Produce state changes to changelog topics

tim-quix:
## Overview

Built on top of `feature/recovery/topic-changelog-manage`

Produces state changes to a changelog topic, which is successfully expanded on/used by another recovery feature branch. This is here to keep the logic specific to just producing to the changelogs separate (since it can be).

Uses column families.


tim-quix:
I will probably update the `delete_cache` logic after seeing and rebasing Haris's PR, which gave me in idea on how I can probably remove its necessity.

tim-quix:
merging everything into 1 branch....

PR #267, opened 2024-01-10
Rebuild state from changelog

tim-quix:
A PR built on top of the `feature/recovery/changelog-produce-cf` branch.

Has logic to recover from changelog topics by reading from them and updating each underlying state store.


tim-quix:
There are still some things to clean up here (one of the worst offenders is consistency of variables names and what they reference, like "topic" vs "source_topic_name", etc) and I want to add some more tests (especially using the Application class directly) and doc strings of course, but a majority of this PR is done at this point and at least worth a look over if someone has the time. I will continue to wrap it up over the next day or so.

tim-quix:
Merging everything into one branch...

PR #268, opened 2024-01-12
Add support for Column families

ghost:
- Add support for Column Families in RocksDBStore to prep the store for the rolling windows
- Keep the store metadata (e.g. last processed offset) in a separate column family


tim-quix (approved):
Looks good!

PR #272, opened 2024-01-22
Fix RocksDB does not retry opening if it's locked by another process

harisbotic:
### Problem
RocksDBStorePartition is supposed to retry opening the RocksDB instance if it's locked by another consumer.
E.g. when a consumer is kicked out of the group and still holds the state partition, another consumer will get this partition assigned and try to open the same DB.

Currently, it works only if you try to open the DB in the same process.
When another process is trying to open the same DB, it will fail with an exception without any retry.

### Solution
- Retry opening db on any error starting with "io error".
- Surface `open_max_retries` and `open_retry_backoff` to RocksDBOptions object so they are configurable by users on the Application level.

### DIFF
--- a/docs/api-reference/application.md
+++ b/docs/api-reference/application.md
@@ -10,7 +10,7 @@
 class Application()
 ```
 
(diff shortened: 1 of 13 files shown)


ghost (approved):
Looks good 👍

PR #273, opened 2024-01-24
Add option to set custom timestamp on producing messages

harisbotic:
### Problem
Currently, there's no way to set a Kafka timestamp in SDF.
By default, to_topic() operator will produce a message with the current epoch time.

But in some cases, like Windows, users might need to set a timestamp to the window "start" timestamp.

### Solution
Provide a callable timestamp similar to `to_topic(key)` to set a timestamp.
Like this `sdf.to_topic(output_topic, timestamp=lambda data: data["ts_field"])`

The callable returns an integer in milliseconds - it's important because Kafka broker only supports millis and ints.

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -255,7 +255,10 @@ class StreamingDataFrame(BaseStreaming):
         return StreamingSeries.from_func(lambda value: key in value)
 
     def to_topic(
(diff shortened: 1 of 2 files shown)


ghost:
Eventually, we decided to implement a different API for setting the timestamps in `StreamingDataFrame`.
I'm closing this PR.

PR #274, opened 2024-01-24
Add docs about windows

ghost:


docs/windowing.md
```diff
@@ -0,0 +1,359 @@
...
+
+- Only time-based windows are supported. 
+- Every window is grouped by the current Kafka message key. 
+- The minimal window unit is a **millisecond**.  
+More fine-grained values (e.g. microseconds) will be rounded towards the closest millisecond number.
+
+
+## Types of Time in Streaming
+
+There are two types of time in streaming systems:
```
tomas-quix, line 194: I think the attempt to not mention the time_extractor API makes the part of docs around timestamp very complicated. Maybe we can document it and mentioned that it will be changed in future release and even mention why?

docs/windowing.md
```diff
@@ -0,0 +1,359 @@
...
+
+   The following code snippet shows how to define a tumbling window over a `StreamingDataFrame`:
+   ```python
+    from datetime import timedelta
+
+
+    sdf = app.dataframe(...)
+
+    # Create a tumbling window of 1 hour
+    sdf.tumbling_window(duration_ms=timedelta(hours=1))
```
tomas-quix: I would put here sdf = sdf.tumbling:window just for safety. People will copy-paste this and will get stuck

docs/windowing.md
```diff
@@ -0,0 +1,359 @@
...
+- [`min()`](https://github.com/quixio/quix-streams/blob/main/docs/api-reference/quixstreams.md#fixedtimewindowdefinitionmin) - to get a minimum value within a window
+- [`max()`](https://github.com/quixio/quix-streams/blob/main/docs/api-reference/quixstreams.md#fixedtimewindowdefinitionmax) -  to get a maximum value within a window
+- [`mean()`](https://github.com/quixio/quix-streams/blob/main/docs/api-reference/quixstreams.md#fixedtimewindowdefinitionmean) - to get a mean value within a window 
+- [`sum()`](https://github.com/quixio/quix-streams/blob/main/docs/api-reference/quixstreams.md#fixedtimewindowdefinitionsum) - to sum values within a window 
+- [`count()`](https://github.com/quixio/quix-streams/blob/main/docs/api-reference/quixstreams.md#fixedtimewindowdefinitioncount) - to count values within a window 
+
+
+## Examples
+
+### Calculate the sum of a message field over a tumbling window
```
tomas-quix: I am thinking, that this simple example would be nice at the beginning of the page as well. I am the sort of person who would just want to start with an example and then dig dipper into the docs and always appreciate something like this example at the beginning.

docs/windowing.md
```diff
@@ -0,0 +1,359 @@
...
+
+    # Alternative syntax to specify window duration and step using milliseconds
+    sdf.hopping_window(duration_ms=60 * 60 * 1000, step_ms=10 * 60 * 1000)
+   ```
+
+    
+
+
+## Lateness and Out-of-Order Processing
+When working with event time, some events may be processed later than they're supposed to.
```
SteveRosam, line 302: i think the term `event` is overloaded. we should be careful because the reader might think we mean event data instead of timeseries data.

ghost: Hey @SteveRosam , thanks for your comment.
The "event vs timeseries" thing seems to come from the Quix format implemented in 0.5, where `EventData` and `TimeseriesData` were different message types.
In 2.0+ there are no plans to keep doing this, so we use the term "event" more broadly.

Plus, the concept of "event time" is also used across the web and other streaming frameworks.

docs/windowing.md
ghost: Agree, I added the docs about the timestamp extractor.

SteveRosam (approved):

PR #280, opened 2024-01-29
Topic Manager changes only with some refactoring

tim-quix:
Isolated out the TopicManager changes and Refactored some things based on Daniils suggestions


quixstreams/models/topics/admin.py
```diff
@@ -0,0 +1,194 @@
...
+                    convert_topic_list(topics),
+                    request_timeout=timeout,
+                ),
+                finalize_timeout,
+            )
+        except CreateTopicFailure as e:
+            failure = {
+                topic.name: {
+                    "failure_reason": e.args[0][topic.name],
+                    "topic_config": topic.config.__dict__,
```
ghost: Please don't use `__dict__` like this. 
There's a `dataclasses.asdict()` method for it.

quixstreams/models/topics/manager.py
```diff
@@ -0,0 +1,446 @@
...
+                    expected = topic.config
+                    if extras:
+                        actual.update_extra_config(
+                            allowed={k for k in expected.extra_config}
+                        )
+                    else:
+                        actual.extra_config = expected.extra_config
+                    if expected != actual:
+                        issues[topic.name] = {
+                            "expected": expected.__dict__,
```
ghost: Similar to other places, please use `dataclasses.asdict()` method to convert a dataclass to a dictionary.

quixstreams/models/topics/manager.py
```diff
@@ -0,0 +1,446 @@
...
+    @property
+    def pretty_formatted_topic_configs(self) -> str:
+        """
+        Returns a print-friendly version of all the topics and their configs
+
+        :return: a pprint-formatted string of all the topics
+        """
+        return pprint.pformat(
+            {
+                topic.name: topic.config.__dict__ if topic.config else "NO CONFIG"
```
ghost: Similar to other places, please use `dataclasses.asdict()` method to convert a dataclass to a dictionary.

ghost (changes requested):

tim-quix:
rebased and ready for merging after approval

quixstreams/models/topics/admin.py
```diff
@@ -0,0 +1,194 @@
...
+                topic.name: {
+                    "failure_reason": e.args[0][topic.name],
+                    "topic_config": topic.config.as_dict(),
+                }
+                for topic in topics
+                if topic.name in e.args[0]
+            }
+            raise CreateTopicFailure(
+                f"failed to create topics:\n{pprint.pformat(failure)}"
+            )
```
ghost: Why do we need to catch and re-raise and exception here vs doing the same thing on line [151](https://github.com/quixio/quix-streams/pull/280/files#diff-fb35e8f5a573400c6f8aa76155325e99b08205dc14b2b9e3d4aa5d9256bd9b2aR151) ?

It looks like the exception is raised only to be re-raised later

quixstreams/models/topics/admin.py
```diff
@@ -0,0 +1,194 @@
...
+        if exceptions:
+            raise CreateTopicFailure(exceptions)
+        if futures:
+            raise CreateTopicTimeout(
+                f"Timed out waiting for creation status for topics:\n"
+                f"{pprint.pformat([topic_name for topic_name in futures])}"
+            )
+
+    def create_topics(
+        self, topics: List[Topic], timeout: int = 10, finalize_timeout: int = 20
```
ghost: Let's increase the default value of `finalize_timeout` to 60.

"20" seems to be too close to "10" in case cluster doesn't respond fast enough and there're multiple topics

quixstreams/models/topics/manager.py
```diff
@@ -0,0 +1,365 @@
...
+            ),
+            timestamp_extractor=timestamp_extractor,
+        )
+        self._topics[name] = topic
+        return topic
+
+    def changelog_topic(
+        self,
+        topic_name: str,
+        suffix: str,
```
ghost: Why do we name it `suffix` instead of `store_name`?

The docstring says it's a store name:
> :param suffix: name of storage type (default, rolling10s, etc.)

ghost (changes requested):

PR #283, opened 2024-02-02
Updates to work in Redpanda brokers via Quix platform

ghost:
# What changed

Various updates to support connecting to Redpanda brokers via Quix Platform:

- Make SSL certificates optional for Quix apps
- Don't set default params for topics created in Quix when `auto_create_topics=True` is set. Use the API defaults instead.

### DIFF
--- a/quixstreams/platforms/quix/config.py
+++ b/quixstreams/platforms/quix/config.py
@@ -262,7 +262,9 @@ class QuixKafkaConfigsBuilder:
             if self.search_workspace_for_topic(ws["workspaceId"], topic):
                 return ws
 
-    def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> str:
+    def get_workspace_ssl_cert(
+        self, extract_to_folder: Optional[Path] = None
+    ) -> Optional[str]:
         """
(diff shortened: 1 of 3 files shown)


quixstreams/platforms/quix/config.py
```diff
@@ -262,22 +262,27 @@ def search_for_topic_workspace(self, topic: str) -> Optional[dict]:
...
-    def get_workspace_ssl_cert(self, extract_to_folder: Optional[Path] = None) -> str:
+    def get_workspace_ssl_cert(
+        self, extract_to_folder: Optional[Path] = None
+    ) -> Optional[str]:
         """
         Gets and extracts zipped certificate from the API to provided folder.
         If no path was provided, will dump to /tmp. Expects cert named 'ca.cert'.
 
         :param extract_to_folder: path to folder to dump zipped cert file to
         :return: full cert filepath as string
```
harisbotic: fix returns: `or None if certificate is empty`

harisbotic (approved):

quixstreams/platforms/quix/config.py
ghost: Good catch, updated the docstring 👍

PR #292, opened 2024-02-21
Chore/type fixing

tim-quix:
Fixed a bunch of typing issues (most of which were identified with Mypy). 

Didn't fix everything according to Mypy, but it does cover most, and practically every error in Pycharm.

Also, Pylance should now do type hints correctly wrt `StreamingSeries` and `StreamingDataFrames`

### DIFF
--- a/quixstreams/dataframe/series.py
+++ b/quixstreams/dataframe/series.py
@@ -1,6 +1,6 @@
 import contextvars
 import operator
-from typing import Optional, Union, Callable, Container, Any
+from typing import Optional, Union, Callable, Container, Any, Mapping, overload
 
 from typing_extensions import Self
 
@@ -185,8 +185,33 @@ class StreamingSeries(BaseStreaming):
         composed = self.compose()
         return context.run(composed, value)
 
+    @overload
     def _operation(
-        self, other: Union[Self, object], operator_: Callable[[object, object], object]
+        self,
+        other: Union[str, int],
+        operator_: Callable[
+            [Union[Mapping, Self], Union[str, int]], Union[Self, object]
+        ],
+    ) -> Self:
+        # getitem
+        ...
+
+    @overload
+    def _operation(
+        self,
+        other: Union[Self, object],
+        operator_: Callable[[Union[Container, Self], object], bool],
+    ) -> Self:
+        # contains
+        ...
+
+    def _operation(
+        self,
+        other: Union[Self, str, int, object],
+        operator_: Callable[
+            [Union[Self, Container, Mapping, object], Union[Self, str, int, object]],
+            Union[bool, object],
+        ],
     ) -> Self:
         self_composed = self.compose()
         if isinstance(other, self.__class__):
@@ -226,7 +251,7 @@ class StreamingSeries(BaseStreaming):
             other, lambda a, b, contains=operator.contains: contains(b, a)
         )
 
-    def contains(self, other: object) -> Self:
+    def contains(self, other: Union[Self, object]) -> Self:
         """
         Check if series value contains "other"
         Same as "other in StreamingSeries".
@@ -251,7 +276,7 @@ class StreamingSeries(BaseStreaming):
         """
         return self._operation(other, operator.contains)
 
-    def is_(self, other: object) -> Self:
+    def is_(self, other: Union[Self, object]) -> Self:
         """
         Check if series value refers to the same object as `other`
 
@@ -274,7 +299,7 @@ class StreamingSeries(BaseStreaming):
         """
         return self._operation(other, operator.is_)
 
-    def isnot(self, other: object) -> Self:
+    def isnot(self, other: Union[Self, object]) -> Self:
         """
         Check if series value does not refer to the same object as `other`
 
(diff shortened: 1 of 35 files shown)


quixstreams/models/serializers/quix.py
```diff
@@ -350,7 +348,7 @@ class QuixTimeseriesSerializer(QuixSerializer):
     def __call__(self, value: Mapping, ctx: SerializationContext) -> Union[str, bytes]:
         if not isinstance(value, Mapping):
             raise SerializationError(f"Expected Mapping, got {type(value)}")
-        result = {
+        result: dict = {
```
ghost: Why would you need the `dict` type hint here?
I mean, the variable is clearly a dictionary, and the hint doesn't specify extra info about the keys and values

quixstreams/kafka/producer.py
```diff
@@ -98,7 +94,7 @@ def __init__(
     def produce(
         self,
         topic: str,
-        value: Union[str, bytes],
+        value: Union[str, bytes] = None,
```
ghost: If we make the `value` param optional, the type hint must be `Optional[Union[str, bytes]]`

quixstreams/kafka/consumer.py
```diff
@@ -363,7 +365,7 @@ def get_watermark_offsets(
         return self._consumer.get_watermark_offsets(**kwargs)
 
     def list_topics(
-        self, topic: Optional[str] = None, timeout: float = -1
+        self, topic: Optional[str] = None, timeout: Optional[float] = -1
```
ghost: This is incorrect.

The `timeout` param is never optional, it's always a `float` - https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.list_topics

quixstreams/dataframe/windows/definitions.py
```diff
@@ -77,7 +80,7 @@ def func(
             start_ms: int,
             end_ms: int,
             timestamp_ms: int,
-            value: Any,
+            value: Numeric,
```
ghost: I disagree with this and other `value: Numeric` changes.

We don't really know the type here, that's why it is `Any`. 
I'd keep them as they are rather than providing the type that is not guaranteed to be correct

quixstreams/dataframe/series.py
```diff
@@ -185,8 +185,33 @@ def test(self, value: Any, ctx: Optional[MessageContext] = None) -> Any:
...
+        self,
+        other: Union[Self, object],
+        operator_: Callable[[Union[Container, Self], object], bool],
+    ) -> Self:
+        # contains
+        ...
+
+    def _operation(
+        self,
+        other: Union[Self, str, int, object],
```
ghost, line 190: Pylance still redlines the uses of `_operations` in some of the dunder methods.

ghost (changes requested):

quixstreams/models/serializers/quix.py
tim-quix: removed cuz I agree it's silly

quixstreams/kafka/consumer.py
tim-quix: adjusted all of these to be consistent on our end (with timeout option of None), along with making sure all arguments match in terms of args/kwargs

Producer poll is the only exception, where a "-1" can hang indefinitely if you poll when no messages are produced, so I set the default of that to 0 instead to be consistent with the "None == -1 == infinite" idea.

quixstreams/dataframe/windows/definitions.py
tim-quix: removed these

quixstreams/dataframe/series.py
tim-quix: As we discussed, these are a bit ugly based on the huge range of functions we're using for it.

I think we can maybe make a separate PR if we find they are causing issues on the user end, otherwise I'm not sure it's worth the time for us to mess with.

I went ahead and dropped the overloads since they weren't really that helpful.

quixstreams/models/topics/topic.py
```diff
@@ -249,8 +254,8 @@ def serialize(
         self,
         key: Optional[object] = None,
         value: Optional[object] = None,
-        headers: Optional[Union[Mapping, MessageHeadersTuples]] = None,
-        timestamp_ms: int = None,
+        headers: Optional[MessageHeadersTuples] = None,
```
ghost: In this method, `headers` can be a dictionary because Producer allows them to be so.
Please see here https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.produce

> headers (dict|list) – Message headers to set on the message. The header key must be a string while the value must be binary, unicode or None. Accepts a list of (key,value) or a dict.

quixstreams/models/topics/topic.py
```diff
@@ -160,6 +163,8 @@ def name(self) -> str:
 
     @property
     def config(self) -> TopicConfig:
+        if self._config is None:
+            raise TopicConfigUndefined("No TopicConfig was provided.")
```
ghost: I think raising an exception here is not the right strategy.

We should do one of the two:
1. Either make the `config` param required in `__init__`. It's easier and makes more sense than raising an error whenever the `.config` property is accessed.
2. Or provide some default value for `config`.

The plain `Topic` object is used only in tests, and it's not supposed to be created directly outside of the lib.
So I think we need to do №1, i.e. make it a required non-optional parameter and update the tests.

ghost (changes requested):

quixstreams/models/topics/topic.py
tim-quix: Yeah, did option 1 :+1:

tim-quix:
Done, ready to merge, just need approval

ghost (approved):

PR #294, opened 2024-02-22
Make env variable `Quix__Portal__Api` Optional

harisbotic:
Set a default value for portal API URL to `https://portal-api.platform.quix.io/` and remove the need to specify it in env variables when using Quix apps

### DIFF
--- a/quixstreams/platforms/quix/api.py
+++ b/quixstreams/platforms/quix/api.py
@@ -9,6 +9,7 @@ from quixstreams.exceptions import QuixException
 from .env import QUIX_ENVIRONMENT
 
 __all__ = ("QuixPortalApiService",)
+DEFAULT_PORTAL_API_URL = "https://portal-api.platform.quix.io/"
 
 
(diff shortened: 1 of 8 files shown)


quixstreams/platforms/quix/api.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         api_version: Optional[str] = None,
         default_workspace_id: Optional[str] = None,
     ):
-        self._portal_api = portal_api or QUIX_ENVIRONMENT.portal_api
+        self._portal_api = portal_api or QUIX_ENVIRONMENT.portal_api or "https://portal-api.platform.quix.io/"
```
ghost: One suggestion - let's move the URL to the module-level variable, e.g. `DEFAULT_PORTAL_API_URL`, to better explain what this URL means.

ghost (changes requested):

ghost (approved):

PR #299, opened 2024-02-28
Set the consumer group to default to 'quixstreams-default'

harisbotic:
Set the consumer group to default to `quixstreams-default` in `Application` class.
This is useful when user wants only to produce data and in that case consumer group is irrelevant.

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -704,7 +704,12 @@ class Application:
         """
         self._setup_signal_handlers()
 
-        logger.info("Initializing processing of StreamingDataFrame")
+        logger.info(
+            f"Starting the Application with the config: "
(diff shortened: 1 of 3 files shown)


quixstreams/app.py
```diff
@@ -231,7 +232,7 @@ def _set_quix_config_builder(self, config_builder: QuixKafkaConfigsBuilder):
     @classmethod
     def Quix(
         cls,
-        consumer_group: str,
+        consumer_group: Optional[str] = "quixstreams-default",
```
ghost: Using `Optional` is not correct here.
It means that the `consumer_group` may be `None` but the `Application` always expects a `str` type.

ghost (changes requested):

ghost (approved):

PR #305, opened 2024-02-29
304: Support GZIP for decompression

### LINKED ISSUE #304: Support [GZIP] Prefix in codec id for legacy quix packages
I'd like to extend support for GZIP compressed individual messages in QuixDeserializer. This feature https://github.com/quixio/quix-streams-dotnet/issues/30 was introduced not long ago to avoid message splitting. Overall topic bandwidth is not expected to be less, but message splitting should be drastically reduced, allowing QuisStreams 2.x to also read these messages easily.

Add Gzip decompression for incoming legacy Quix messages

peter-quix:


quixstreams/models/serializers/quix.py
```diff
@@ -221,6 +224,11 @@ def __call__(self, value: bytes, ctx: SerializationContext) -> Iterable[dict]:
                 f'Ignore Quix message with "{QModelKey.HEADER_NAME}": "{model_key}"'
             )
 
+        # Check if compressed
+        if codec_id.startswith(QCodecId.GZIP_PREFIX):
+            codec_id = codec_id[len(QCodecId.GZIP_PREFIX) :]
+            value = gzip.decompress(value)
+
```
tim-quix, line 231:
```suggestion
            if not (
                codec_id.startswith(QCodecId.GZIP_PREFIX) 
                and (codec_id := codec_id[len(QCodecId.GZIP_PREFIX) :]) in QCodecId.SUPPORTED_CODECS
            ):
                raise SerializationError(
                    f'Unsupported "{QCodecId.HEADER_NAME}" value "{codec_id}"'
                )
            value = gzip.decompress(value)
```

tim-quix: This suggestion closely matches the behavior you had before; basically if you had `[GZIP]badcodec`, would raise `... value "badcodec"`, while also raising `...value "[NOTGZIP]goodcodec"` if it didn't start with GZIP

tim-quix:
We should also prob include a unit test for this as a sanity check?

tim-quix (approved):

PR #306, opened 2024-03-01
Implement handling messages without keys

harisbotic:
#### Allow events with `null` keys overall in StreamingDataFrame.
* `null` is still a value, plus users can change the key to something else

#### Skip events with `null` keys in Windows and log warnings:
* Messages with the `null` keys can be sent to random partitions (depending on the partitioner).
* Therefore the result of the window with the `null` key can be a random value because of it
* If the events have `null` keys, they must be re-keyed before processing them in the window.

### DIFF
--- a/quixstreams/dataframe/windows/time_based.py
+++ b/quixstreams/dataframe/windows/time_based.py
@@ -206,8 +206,8 @@ def _as_windowed(
         if key is None:
             ctx = message_context()
             logger.warning(
-                f"Skipping window processing for a message because the key is None. "
-                f"Message offset: {ctx.offset}, partition: {ctx.partition}."
+                f"Skipping window processing for a message because the key is None, "
+                f"partition='{ctx.topic}[{ctx.partition}]' offset='{ctx.offset}'."
             )
             return _noop()
         with transaction.with_prefix(prefix=key):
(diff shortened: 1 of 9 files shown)


quixstreams/dataframe/windows/time_based.py
```diff
@@ -190,6 +203,13 @@ def wrapper(value: object) -> object:
             state_manager.get_store_transaction(store_name=store_name),
         )
         key = message_key()
+        if not key:
```
ghost: Using `not` is incorrect here because it can filter out keys like `0` and `False`, which are valid values.

Use `is None` check instead

ghost (changes requested):

quixstreams/dataframe/windows/time_based.py
```diff
@@ -190,6 +203,13 @@ def wrapper(value: object) -> object:
             state_manager.get_store_transaction(store_name=store_name),
         )
         key = message_key()
+        if key is None:
+            ctx = message_context()
+            logger.warning(
+                f"Skipping window processing for a message because the key is None. "
+                f"Message offset: {ctx.offset}, partition: {ctx.partition}."
```
harisbotic: Should we add a `ctx.topic` here as well?

ghost (approved):

Issue #302, opened 2024-02-29
SDK breaks when handling in array


SteveRosam:
The data being published (via the SDK) from the previous node in the pipeline is:

`[[6], [7], [8], [9], [10]]`

The error I get is:

`TypeError: Row value must be a dict, but got "<class 'list'>"`

Suggestion is: throw an error on publish so the data never enters the topic. 

Workaround: I have to delete my topic now.

Slack discussion: https://stream-processing.slack.com/archives/C033BGFD1L6/p1709210724220199

ghost:
Fixed in [v2.3.2](https://github.com/quixio/quix-streams/releases/tag/v2.3.2)

PR #307, opened 2024-03-01
Allow deserialisation of non-dict values 

harisbotic:
Allow Deserialisation of non-dict Values 

Fixes issue: https://github.com/quixio/quix-streams/issues/302


ghost (approved):
LGTM 👍

PR #310, opened 2024-03-08
Example Bugfix: Do not override state input on each update

alexmorley:
Think this might have been added accidentally in a previous commit.

It breaks the example (its not the correct type and it gets re-initialised every update).


ghost (approved):
Hi @alexmorley , thanks for the PR!
This bug has crawled into the latest update, yes.
Thanks for spotting it.

Approved and merged.

PR #312, opened 2024-03-12
A draft implementation of S3 sink and base API

ghost:
This is a draft implementation of the S3 sink and base sink API.
It needs more changes & testing before being merged.

We also plan to package it as a separate Python package to develop it on its own schedule (not done yet).

## What's new

- Added `BaseSink` and `BaseBatchingSink` classes.
  - The idea is that there're at least 2 types of sinks:
    - **Streaming** - the ones where we can write message-by-message, like MQTT, for example. Not a part of this PR yet.
    - **Batching** - the ones that require data to be batched first, like S3.
  
  - `BaseSink` contains a minimal functionality and configuration common for all types of sinks: starting & stopping the app, validating topics, handling OS signals.
  - `BaseBatchingSink` implements the actual consumer loop and provides facilities for batching messages and flushing them to the destination. This class is to be used when you need to create a new batching Sink.

- Implemented `S3Sink` as an example of a Batching Sink.  
With this sink, you can stream messages from Kafka topics to an S3 bucket in various formats and using different batching strategies.   
See the docstring of `S3Sink` class for more  details.

## Example usage
```
from quixstreams.connectors.sinks.s3 import S3Sink

# Configuring the sink app
sink = S3Sink(
    format="bytes",
    topic="topic-name",
    s3_bucket_name="bucket-name",
    s3_access_key_id="<access-key-id>",
    s3_secret_access_key="<access-key-secret>",
    batch_max_messages=100000,
)

# Running the app
sink.run()
```

# What's next
- Add tests and docs
- Resolve TODOs

### DIFF
--- a/quixstreams/connectors/sinks/base/batching.py
+++ b/quixstreams/connectors/sinks/base/batching.py
@@ -1,0 +1,133 @@
+import time
+from typing import Dict, List, Optional, Any
+
+from quixstreams.connectors.sinks.base.exceptions import BatchFullError
+
+
+__all__ = ("Batch", "BatchStore")
+
+
+# TODO: Docs
+# TODO: Wanna make it flexible, but it's probably overkill
+# - TODO: Batching by system time is not idempotent and may create duplicates
+
+
+class Batch:
+    def __init__(
+        self,
+        topic: str,
+        partition: int,
+        start_offset: int,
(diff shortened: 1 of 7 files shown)


tomas-quix:
Closing after discussion. We will use different approach.

PR #313, opened 2024-03-12
Remove need for Application.Quix()

tim-quix:
`Application` now can be used as a Quix application WITHOUT `Application.Quix()`:
- use the NEW `quix_sdk_token` argument (or set the same environment variable as before, `Quix__Sdk__Token`) to generate the same behavior as `Application.Quix()`
-  `broker_address` is now  "optional" to enable this behavior
    - `quix_sdk_token`  and `broker_address` are required, exclusive OR's:  only one MUST be set.
- `Application.Quix()` will be deprecated some time in the future

Added Deprecation warning to `Application.Quix()`
- Kept the old tests around (including fixtures...renamed them to not clash) but they are ready to be deleted after we remove it
- It will continue to work as before if you are currently using it, but we recommend changing to `Application` now.

Added new env vars, which maps to the following Application args:
- `Quix__Broker__Address` - `broker_address`
- `Quix__Consumer__Group` - `consumer_group`

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -658,8 +658,6 @@ class Application:
           is stateful and is running in Quix Cloud
         """
         # Ensure that state management is enabled if application is stateful
-        # and is running in Quix Cloud
-        logger.info("APPLICATION IS CONNECTED TO THE QUIX PLATFORM")
         if self._state_manager.stores:
             check_state_management_enabled()
 
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -91,8 +91,8 @@ class Application:
 
     def __init__(
         self,
-        consumer_group: str = "quixstreams-default",
         broker_address: Optional[str] = None,
+        consumer_group: str = "quixstreams-default",
         auto_offset_reset: AutoOffsetReset = "latest",
         auto_commit_enable: bool = True,
         partitioner: Partitioner = "murmur2",
@@ -160,10 +160,15 @@ class Application:
         :param on_producer_error: triggered when RowProducer fails to serialize
             or to produce a message to Kafka.
         """
-
+        broker_address = broker_address or environ.get("QUIXSTREAMS_BROKER_ADDRESS")
         configure_logging(loglevel=loglevel)
         self._quix_config_builder = None
         if self.is_quix_app:
+            if broker_address is not None:
+                raise ValueError(
+                    "Cannot provide a 'broker_address' when "
+                    "'Quix__Sdk__Token' is also defined in environment"
+                )
             quix_config_builder = QuixKafkaConfigsBuilder()
             self._topic_manager_class = lambda **kwargs: QuixTopicManager(
                 **kwargs, quix_config_builder=quix_config_builder
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -95,7 +95,7 @@ class Application:
         self,
         broker_address: Optional[str] = None,
         quix_sdk_token: Optional[str] = None,
-        consumer_group: str = "quixstreams-default",
+        consumer_group: Optional[str] = None,
         auto_offset_reset: AutoOffsetReset = "latest",
         auto_commit_enable: bool = True,
         partitioner: Partitioner = "murmur2",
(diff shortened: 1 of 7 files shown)


quixstreams/app.py
```diff
@@ -226,168 +255,9 @@ def __init__(
             ),
         )
 
-    def _set_quix_config_builder(self, config_builder: QuixKafkaConfigsBuilder):
-        self._quix_config_builder = config_builder
-
-    @classmethod
-    def Quix(
-        cls,
```
ghost: We need to keep `Application.Quix()` for backward compatibility reasons.

ghost (changes requested):

tim-quix:
Prob need to go fix some tests still, so lemme know if you're happy with this (or feel free to change them yourself since I know you're adding some config changes!)

quixstreams/app.py
```diff
@@ -61,10 +64,10 @@ class Application:
 
     What it Does:
 
-    - During user setup:
+    - On init:
         - Provides defaults or helper methods for commonly needed objects
-        - For Quix Platform Users: Configures the app for it
-            (see `Application.Quix()`)
+        - If `quix_sdk_token` is passed, configures the app for to work
+            with Quix Platform.
```
ghost: Please use "Quix Cloud" here instead of "Platform".

We refer to it as Quix Cloud in the new docs too.

quixstreams/app.py
```diff
@@ -108,15 +112,25 @@ def __init__(
...
+        :param quix_sdk_token: If using the Quix Platform, the SDK token to connect with.
+            Either this OR broker_address must be set to use Application (not both).
+            Linked Environment Variable: `Quix__Sdk__Token`
+            Default: None (if not run on Quix Platform)
+              >***NOTE:*** the environment variable is set for you on the Quix Platform
         :param consumer_group: Kafka consumer group.
             Passed as `group.id` to `confluent_kafka.Consumer`
-            Default - "quixstreams-default".
+            Linked Environment Variable: `Quix__Consumer__Group`
+            Default - "quixstreams-default" (post-init).
```
ghost: What does "post-init" mean here?

quixstreams/app.py
```diff
@@ -108,15 +112,25 @@ def __init__(
...
+              >***NOTE:*** Quix Applications will prefix it with the Quix workspace id.
```
ghost, line 132: What does this mean? We haven't had  this note before

quixstreams/app.py
```diff
@@ -158,8 +172,61 @@ def __init__(
...
+            quix_config_builder = QuixKafkaConfigsBuilder(quix_sdk_token=quix_sdk_token)
+
+        if broker_address and quix_config_builder:
+            raise ValueError("Cannot provide both broker address and Quix SDK Token")
+        elif not (broker_address or quix_config_builder):
+            raise ValueError("Either broker address or Quix SDK Token must be provided")
+        elif quix_config_builder:
+            # SDK Token or QuixKafkaConfigsBuilder were provided
+            logger.info(
+                f"QUIX APPLICATION DETECTED; it will connect to Quix Platform brokers"
```
ghost: 1. Quix Platform -> Quix Cloud

2. Let's expand the "detected" here so it's clear how exactly it was detected that it's a Quix app.  
E.g. "Quix SDK Token detected; the application will connect to the broker in Quix Cloud"

ghost (changes requested):

quixstreams/app.py
tim-quix: we had a similar note with Application.Quix, so I was also passing that info along. Admittedly I'm not sure if it's needed...but if someone sees their consumer group looks weird, wanted them to know why.

Whether someone is gonna inspect things to that degree is...prob unlikely?

quixstreams/app.py
tim-quix: Was referring to the fact that `None` is not the true default here...but maybe it's enough to just say it's the default?

quixstreams/app.py
ghost: Oh, my bad, I misread "prefix it with" as "**prefers** it with"  =)

quixstreams/app.py
tim-quix: Oh, didn't know we were doing that yet, I actually wrote cloud first and then reversed thinking we weren't doing it yet LOL

quixstreams/app.py
ghost: Then let's keep it

quixstreams/app.py
ghost: Let's leave just "Default - ..." or explain how this default is figured out here.

quixstreams/app.py
```diff
@@ -604,10 +655,11 @@ def _quix_runtime_init(self):
         """
         Do a runtime setup only applicable to an Application.Quix instance
         - Ensure that "State management" flag is enabled for deployment if the app
-          is stateful and is running on Quix platform
+          is stateful and is running in Quix Cloud
         """
         # Ensure that state management is enabled if application is stateful
-        # and is running on Quix platform
+        # and is running in Quix Cloud
+        logger.info("APPLICATION IS CONNECTED TO THE QUIX PLATFORM")
```
ghost: Do we need this logging code here? 
Looks like a leftover

ghost (changes requested):

quixstreams/app.py
tim-quix: I had added this one so when the app actually runs (vs just init'ing it) it lets you know it's connecting to Quix. Might be overkill though?

ghost: Oh, I see. 
I think it is overkill because it already logs the broker address here - https://github.com/quixio/quix-streams/blob/45f3e15f1515730e72ffc348bcbad86cda445c78/quixstreams/app.py#L759

ghost (approved):

PR #314, opened 2024-03-18
push a draft of CSV template, more to clean up

tim-quix:


tomas-quix:
Closing after a discussion with Daniil, we will use a different approach.

PR #320, opened 2024-03-28
added better Quix API error handling

tim-quix:
Now reveals the request body of the Quix API for non-200's to better surface what errors are actually occurring.


quixstreams/platforms/quix/api.py
```diff
@@ -63,24 +61,47 @@ def request(self, method, url, **kwargs):
...
+            try:
+                reason_text = e.response.json()
+            except requests.exceptions.JSONDecodeError:
+                reason_text = e.response.text
+
+            raise QuixApiRequestFailure(
+                {
+                    "response_error": e,  # response error
+                    "response_body": reason_text,  # error message in the API response
+                }
```
ghost: 1. Let's catch only `requests.exceptions.HTTPError` here, because this is what is raised from `raise_for_status()`
2. Instead of passing a dictionary here, let's pass an error message as a string to make it more pythonic and limit the output. For example: `{status code} Error for url "{url}": {response_body}`:  
`400 Error for URL "https://example.com": {"error": "Error message from the server"}`

quixstreams/platforms/quix/api.py
```diff
@@ -63,24 +61,47 @@ def request(self, method, url, **kwargs):
     @property
     def default_workspace_id(self) -> str:
         if not self._default_workspace_id:
-            raise self.UndefinedQuixWorkspaceId("You must provide a Quix Workspace ID")
+            raise UndefinedQuixWorkspaceId("You must provide a Quix Workspace ID")
```
ghost: Let's specify how exactly it should be provided (e.g. via env variable), because we don't accept workspace IDs in `Application`.

quixstreams/platforms/quix/api.py
```diff
@@ -63,24 +61,47 @@ def request(self, method, url, **kwargs):
...
+                    "response_body": reason_text,  # error message in the API response
+                }
+            )
+
     def _init_session(self) -> SessionWithUrlBase:
         if not (self._portal_api and self._auth_token):
             missing = filter(
                 None,
                 [self._portal_api or "portal_api", self._auth_token or "auth_token"],
             )
```
ghost: Since we're touching the code nearby, this block has issues too.
`filter()` creates a lazy object, and when you cast it to `str`, you get `'<filter object at 0x000000>'`

I suggest we remove these checks from `_init_session` and instead do them in `QuixPortalApiService.__init__`

Plus, we don't need to check portal API anymore - there's always a default value.
We only need to check the `auth_token`, and `QuixPortalApiService.__init__` is the best place to do that

ghost (changes requested):

tim-quix:
Give it one final look @daniil-quix, specifically around the auth_token handling in the init and error message formats =)

quixstreams/platforms/quix/api.py
```diff
@@ -61,16 +66,40 @@ def request(self, method, url, **kwargs):
...
+        """
+        Custom callback/hook that is called after receiving a request.Response
+
+        Catches non-200's and passes both the original exception and the Response body.
+
+        Note: *args and **kwargs expected for hook
+        """
+        try:
+            r.raise_for_status()
+        except requests.exceptions.RequestException as e:
```
ghost: As mentioned previously, let's catch only `HTTPError` here, not `RequestException`.
Only `HTTPError` exceptions are raised in `raise_for_status()`

quixstreams/platforms/quix/api.py
```diff
@@ -61,16 +66,40 @@ def request(self, method, url, **kwargs):
...
+            try:
+                reason_text = e.response.json()
+            except requests.exceptions.JSONDecodeError:
+                reason_text = e.response.text
+
+            reason_text = reason_text or "Invalid URL"
```
ghost: Where does the `"Invalid URL"` come from?

quixstreams/platforms/quix/api.py
```diff
@@ -61,16 +66,40 @@ def request(self, method, url, **kwargs):
...
+            raise QuixApiRequestFailure(
+                f'{e.response.status_code} for url "{e.response.url}": {reason_text}'
```
ghost: Let's add the word "Error " before the status code here.

ghost (changes requested):

quixstreams/platforms/quix/api.py
tim-quix: This is if you hit an endpoint that just doesn't exist period (there isn't an error message/body is empty).

I think it'll only come into play if you were to use the session object to try and hit other endpoints manually tbh (which is how I found it), but I didn't test all of them to know for sure...so I just added this as a way of clarifying it's a bad endpoint if no error was given.

ghost: I think this is related to https://github.com/quixio/quix-streams/pull/320#discussion_r1549306870

If we catch only particular exceptions, the situation you tried to handle will be eliminated (every exception will be guaranteed to have a response)

quixstreams/platforms/quix/api.py
tim-quix: IDK what happened, I had changed it but I think something got reversed when I rebased or something.

quixstreams/platforms/quix/api.py
tim-quix: Yeah, that's a diff mistake =) But what I was attempting to cover here is still applicable.

If you hit an endpoint that doesn't exist, the request has no response body (at least, for our API anyway). It still throws the error (like `HTTPError: 404 Client Error: blah blah`) That being said, IDK if this can happen with things like 403's as well, which "Invalid URL" would be inappropriate for.

I think I found something a little cleaner that shortens the error when we have more info on what failed, and otherwise just returns the original as-is if not.

ghost (approved):

PR #322, opened 2024-04-03
add clearer error messages around invalid SDF column name references

tim-quix:
Add clearer error messages around invalid SDF column name references

Thoughts:
- Also added a custom exception rather than re-using `TypeError` for non-dicts.
- Do we want to improve error messages for `StreamingSeries` column access as well? Could easily re-use this new function for `__getitem__`. Seems common enough use that it might be worthwhile
  - on the other hand, this could be part of a larger error handling improvement later


quixstreams/dataframe/series.py
```diff
@@ -9,11 +9,34 @@
...
+    :return: Nested data from column name
+    """
+    try:
+        return d[column_name]
+    except KeyError:
+        raise MissingColumn(
+            f"Column name '{column_name}' does not exist in the StreamingDataFrame"
+        )
+    except TypeError:
+        d_type = getattr(type(d), "__name__", type(d))
```
ghost: Why do you need to use `getattr` here?

`str(type(d))` looks much shorter

quixstreams/dataframe/series.py
```diff
@@ -9,11 +9,34 @@
...
+        raise InvalidColumnReference(
+            f"Cannot reference column name '{column_name}' for StreamingDataFrame "
+            f"data type '{d_type}'; a dictionary is required."
```
ghost: Suggestion here:
> 'Cannot access column "{column_name}"; the underlying value must be a mapping, got '{d_type}' instead."

ghost (changes requested):

quixstreams/dataframe/series.py
```diff
@@ -9,11 +9,34 @@
...
+    :param column_name: the column name.
+
+    :return: Nested data from column name
+    """
+    try:
+        return d[column_name]
+    except KeyError:
+        raise MissingColumn(
+            f"Column name '{column_name}' does not exist in the StreamingDataFrame"
+        )
```
ghost: Let's  rename "MissingColumn" to "ColumnDoesNotExist" (as it was in the task).
On top of it, the text says "does not exist" anyway :)

ghost (changes requested):

quixstreams/dataframe/series.py
tim-quix: the `__name__` changes it from something like `<class 'str'>` to just `'str'`, which I thought was nicer to read (and then could not assume everything thrown at it would have `__name__` defined, so I did `getattr`)

of course, it'll treat it like `str(type(d))` otherwise =)

Let me know if you still don't like it.

quixstreams/dataframe/series.py
tim-quix: one thought around this: how do you feel around the language with mentioning `StreamingDataFrame`?

I intentionally put that in there to make it clear where a mapping is expected (I know it should be relatively obvious based on this error message, but I wanted it unmistakable what's happening, haha)

tim-quix: along that line, `underlying value` may be a bit ambiguous, potentially.

Wasn't sure the best way to refer to the `StreamingDataFrame` data/value, so I went with just calling it the `StreamingDataFrame`, kinda like how you'd refer to the dataframe object in Pandas.

quixstreams/dataframe/series.py
ghost: Let's  do `obj.__class__.__name__` instead

ghost (changes requested):

ghost (approved):

PR #333, opened 2024-04-11
Checkpointing v2

ghost:
# New checkpointing

## Why?
Existing commit & recovery implementation has the following issues:
1. The delivery of produced messages (both for changelogs and output) is not guaranteed and the library doesn't check the delivery callbacks
2. The changelog offsets are calculated on the fly before they're produced, which leaves room for errors
3. Local state stores are updated for each processed message which puts more pressure on the disks

## Goals
1. Provide better state consistency guarantees in the At-Least-Once setting.
2. Optimize the state performance.
3. Build a foundation for EOS processing

## What's changed

1. Introduced a new class `Checkpoint` responsible for flushing the Producer, committing topic offsets, and flushing the state in sync.  
Now it first flushes the Producer and ensures that all outgoing messages are delivered, then synchronously commits the topic offsets, and only then does it flush the state to the disk.   

2. The Checkpoint commits on a schedule provided by the `commit_interval` Application setting.  
The default interval is 5 seconds, the same as Kafka Consumer's default autocommit interval.
Kafka Consumer autocommit is now always disabled.

3. The state updates are batched in-memory before the checkpoint is committed.  
It improves the state performance because the disks are only touched once in several seconds, although it increases memory usage.   
The batches are kept per key, so the more keys are processed during the checkpoint interval, the more memory it will require to store.
To use less memory, users may reduce the `commit_interval` value. 

4. The changelog messages now have the source topic-partition-offset info attached in the headers, and the recovery checks if the changelog belongs to the committed source topic message.  
It ensures that the state changes are not applied for the messages that are not yet committed.
The app should always* recover state consistently minimizing the chance for double-counting in the At-Least-Once setting.

5. The store changelog offsets are now taken from the delivery callbacks (no manual calculation anymore).

6. The internal RowProducer now has `enable.idempotence = True` by default to provide stronger delivery guarantees.  
This setting can be disabled by providing `producer_extra_config = {'enable.idempotence': False}` to the Application class.

8. Lots of refactoring and tests

## Caveats
1. Although the stateful processing performance increases, the apps will require more memory to batch the state updates.
2. In the At-Least-Once setting, it is still possible that unwanted changelog changes get applied.  
**Example:**
  - Checkpoint successfully produces changelog updates
  - Checkpoint fails to commit the source topic offsets
  - The user changes the application code and some of the input messages get filtered during re-processing.  
  Since the changelogs are already produced, during recovery from scratch the app will apply them to the state.  
  To mitigate this, the EOS should be used, which we will implement in the future

## Docs
I'll update the docs after the code review


quixstreams/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,165 @@
...
+        ), transaction in self._store_transactions.items():
+            offset = self._tp_offsets[(topic, partition)]
+            if transaction.failed:
+                raise StoreTransactionFailed(
+                    f'Detected a failed transaction for store "{store_name}", '
+                    f"the checkpoint is aborted"
+                )
+            transaction.prepare(processed_offset=offset)
+
+        # Step 2. Flush producer to trigger all delivery callbacks and ensure that
```
tim-quix, line 130: maybe label as 1a/1b to match with the description in the docstring =)

quixstreams/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,165 @@
...
+        produced_offsets = self._producer.offsets
+
+        # Step 3. Commit offsets to Kafka
+        offsets = [
+            TopicPartition(topic=topic, partition=partition, offset=offset + 1)
+            for (topic, partition), offset in self._tp_offsets.items()
+        ]
+        self._consumer.commit(offsets=offsets, asynchronous=False)
+
+        # Step 4. Flush state store partitions to the disk together with changelog
```
tim-quix, line 143: Same idea as above (Step 3.)

quixstreams/app.py
```diff
@@ -96,7 +100,7 @@ def __init__(
         quix_sdk_token: Optional[str] = None,
         consumer_group: Optional[str] = None,
         auto_offset_reset: AutoOffsetReset = "latest",
-        auto_commit_enable: bool = True,
+        commit_interval: float = 5.0,
```
tim-quix, line 103: missing doc string for commit interval

quixstreams/state/rocksdb/partition.py
```diff
@@ -103,28 +99,71 @@ def _changelog_recover_flush(self, changelog_offset: int, batch: WriteBatch):
...
+
         batch = WriteBatch(raw_mode=True)
-        key = changelog_message.key()
-        if value := changelog_message.value():
-            batch.put(key, value, cf_handle)
-        else:
-            batch.delete(key, cf_handle)
+        # Determine whether the update should be applied or skipped based on the
+        # latest committed offset and processed offset from the changelog message header
+        if not self._should_skip_changelog(
```
tim-quix: Minor suggestion/preference: something like `if _is_valid_changelog` reads a little nicer IMO (i.e. `if True` rather than a `if not False`)

quixstreams/state/rocksdb/partition.py
```diff
@@ -103,28 +99,71 @@ def _changelog_recover_flush(self, changelog_offset: int, batch: WriteBatch):
...
+            headers=headers, committed_offset=committed_offset
+        ):
+            key = changelog_message.key()
+            if value := changelog_message.value():
+                batch.put(key, value, cf_handle)
+            else:
+                batch.delete(key, cf_handle)
+
```
tim-quix, line 166: We already talked about this, but just documenting it since I already had it written down:

Seems like we should also batch up our changes when recovering, too.

Having a caching class like we discussed may make it easier? Could also re-use/extend/re-implement the context manager/checkpointing for this?

quixstreams/state/rocksdb/partition.py
```diff
@@ -103,28 +99,71 @@ def _changelog_recover_flush(self, changelog_offset: int, batch: WriteBatch):
...
+        :param committed_offset: latest committed offset of the source topic partition
+        :return: True if update should be skipped, else False.
+        """
+        # Parse the processed topic-partition-offset info from the changelog message
+        # headers to determine whether the update should be applied or skipped.
+        # It can be empty if the message was produced by the older version of the lib.
+        processed_offset_header = headers.get(
+            CHANGELOG_PROCESSED_OFFSET_MESSAGE_HEADER, b"null"
+        )
+        processed_offset = json_loads(processed_offset_header)
```
tim-quix, line 118: Does this need to be JSON anymore due to your previous adjustments to this? Might be a little faster to not use JSON

quixstreams/state/rocksdb/transaction.py
```diff
@@ -295,72 +326,149 @@ def failed(self) -> bool:
...
+                else serialize(prefix, dumps=self._dumps)
+            ),
+        )
+
+    def _produce_changelog(self, processed_offset: Optional[int] = None):
+        changelog_producer = self._changelog_producer
+        if changelog_producer is None:
+            return
+
+        source_topic, changelog_topic, partition = (
```
tim-quix: `source_topic` no longer used

quixstreams/state/recovery.py
```diff
@@ -140,16 +168,35 @@ class ChangelogProducer:
...
+    ):
         """
         :param changelog_name: A changelog topic name
-        :param partition_num: source topic partition number
+        :param partition: source topic partition number
         :param producer: a RowProducer (not shared with `Application` instance)
         """
         self._changelog_name = changelog_name
-        self._partition_num = partition_num
+        self._source_topic_name = source_topic_name
```
tim-quix: `source_topic_name` not used anymore from what I can tell

quixstreams/processing_context.py
```diff
@@ -0,0 +1,80 @@
...
+__all__ = ("ProcessingContext",)
+
+logger = logging.getLogger(__name__)
+
+
+class CheckpointNotInitialized(QuixException): ...
+
+
+@dataclasses.dataclass
+class ProcessingContext:
```
tim-quix, line 21: This class is fairly coupled with checkpoints, wonder if it would make sense to have them grouped together more (no functional changes, just organizational)?

quixstreams/state/recovery.py
```diff
@@ -140,16 +168,35 @@ class ChangelogProducer:
...
+        self._partition = partition
```
tim-quix, line 179: This is more just a general comment:

in case you were wondering at the extra verbosity around things like `partition` vs `partition_num`, wanted it to be more obvious at a glance whether the objects we were referring to were either `Topic` objects or `topic` names, or `Partition`-like objects (our RDB objects) or actual Kafka partition numbers.

In some ways it's a little redundant with everything typed already, but I found it helped me while reading the code sometimes since we kinda call these different objects using the same variable names within a similar, overlapping context. 

Feel free to keep your changes if you feel it's not useful, but I thought I'd expand a little in case you didn't know why I had it that way.

quixstreams/app.py
```diff
@@ -807,15 +804,26 @@ def _on_assign(self, _, topic_partitions: List[TopicPartition]):
...
+        self._processing_context.commit_checkpoint(force=True)
+
         # assigning manually here (instead of allowing it handle it automatically)
         # enables pausing them during recovery to work as expected
         self._consumer.incremental_assign(topic_partitions)
 
         if self._state_manager.stores:
             logger.debug(f"Rebalancing: assigning state store partitions")
             for tp in topic_partitions:
+                # Get the latest committed offset for the assgined topic partition
```
```suggestion
                # Get the latest committed offset for the assigned topic partition
```

quixstreams/state/rocksdb/transaction.py
```diff
@@ -187,90 +142,154 @@ def get(
...
+        if cached is not UNDEFINED:
             return True
 
         return self._partition.exists(key_serialized, cf_name=cf_name)
 
+    @_validate_transaction_status(PartitionTransactionStatus.STARTED)
+    def prepare(self, processed_offset: int):
+        """
+        Produce changelog messages to the changelog topic for all changes accumulated
+        in this transaction and prepare transcation to flush its state to the state
```
tim-quix:
```suggestion
        in this transaction and prepare transaction to flush its state to the state
```

quixstreams/state/rocksdb/transaction.py
```diff
@@ -283,7 +302,19 @@ def completed(self) -> bool:
...
         """
-        return self._completed
+        return self._status == PartitionTransactionStatus.COMPLETE
+
+    @property
+    def prepared(self) -> bool:
+        """
+        Check if the transaction is in PREPARED status.
+
+        Prepared transaction successefully flushed its changelog and cannot receive
```
tim-quix:
```suggestion
        Prepared transaction successfully flushed its changelog and cannot receive
```

tim-quix (commented):
Really nice work.

Nothing major (and some things you already addressed).

For documentation reasons, Daniil and I also discussed potentially implementing a caching class for the RDB transactions, and probably some sort of batching for recovery as well. He also has some other outstanding refactoring he'd like to do, but out of scope of these changes.

Also, don't forget to build/update any docs as necc (or maybe we just do that prior to a release?) =)

quixstreams/state/rocksdb/partition.py
ghost: I agree with the batching overall, but here we don't need any caching to do that.
Caching is mainly for reading, and here we only write.
Here something more simple is needed to batch writes.

As discussed, we'll do that separately.
I created https://github.com/quixio/quix-streams/issues/351 to track it

quixstreams/processing_context.py
ghost: What exactly do you mean by grouping `ProcessingContext` and `Checkpoint` classes?

quixstreams/checkpointing/checkpoint.py
ghost: I updated the docstring instead to make it more detailed

quixstreams/state/rocksdb/partition.py
ghost: Agree, `if not` is harder to digest in general.

quixstreams/processing_context.py
ghost: For posterity: `ProcessingContext` uses `Checkpoint` as a dependency, but they don't need to be in the same modules.

quixstreams/state/recovery.py
ghost: The changes here were made to keep things consistent across the code. 
I suggest we use "partition" when we refer to the Kafka partition numbers, "topic_partition" or "tp" when it's a `TopicPartition` instance, and "state_partition" when it's the `StatePartition` instance.

quixstreams/state/rocksdb/partition.py
ghost: The JSON is used here to simplify the debugging in case you need to examine the produced messages using some Kafka UI.
I also checked the performance, and the `orjson` lib is as fast as serializing ints to bytes using `struct`.

quixstreams/state/rocksdb/transaction.py
ghost: Agree, I removed it from the changelog classes

tim-quix (approved):
Woo, lets merge!

PR #334, opened 2024-04-12
Groupby prototype

tim-quix:
Added GroupBy as spec'ed in the story, with some additional features (to enable GroupBy functionality).

# Left before ready to merge:
- review
- discuss naming around group by topics
- add tests (besides app level)

# How it Works

Added a "branches" parameter to `StreamingDataFrame` which stores multiple `StreamingDataFrame` instances in a dict, with the `SDF`'s topic name as the key.

- When an `SDF.group_by` is called, it stores the current active `SDF` in `SDF._branches` and creates a new `SDF`, which uses a newly created GroupBy topic.
- The consumer will consume from this GroupBy topic, and the `SDF` used for processing a message depends on the topic the message came from (referenced via `SDF._branches`)

### `StreamingDataFrame`
- added `_application` parameter for Application reference since GroupBy needs info on consumer groups and needs the StateManager and TopicManager anyway
- Daniil also mentioned how he was planning on adding the Application reference anyway.

### `StreamingDataFrame.test()`
- added `mock_producer` argument that enables skipping produce while testing, 
- additionally useful for skipping groupbys and testing all the SDF branches together without Kafka.
- accomodates testing with GroupBys

### `Topic`:
-  `topic_config` argument now additionally accepts a `Callable` (which should return a `TopicConfig`) 
- This allows you to create a `Topic` that requires calling Kafka to finalize what creation config it should use (usually because it depends on another, like GroupBy topics) without doing it until creation is actually required.
  - Largely added for GroupBy `Topic` management 

### `TopicManager` cleanup
- cleanup around internal topic naming (especially with the addition of another type)

## Performance

Tested/Benchmarked with python3.8 + example app to compare performance between `main` and this branch. 

The changes that enable this GroupBy behavior do not appear to have any noticeable impact on performance.

## Other notes

- Not sold on the "branches" name yet, just left it for now.
- Considered maybe changing how GroupBy `Topics` are managed to avoid adding the `Topic.config` callable, but seems like other approaches I considered would likely be fairly convoluted. Plus, accepting a callable is not so bad (but I might improve how that's handled)


tim-quix:
closing:
- most relevant changes converted over to the new PR that uses checkpointing.
- other changes are not necc or outdated.

PR #337, opened 2024-04-19
Chore/quix topic create error

tim-quix:
Additional error handling around quix API topic creation requests when they have an `Error` status


quixstreams/platforms/quix/config.py
```diff
@@ -360,8 +359,19 @@ def _finalize_create(self, topics: Set[str], timeout: Optional[int] = None):
...
                     topics.remove(topic["id"])
+                elif topic["status"] == "Error":
+                    logger.debug(f"Topic {topic['name']} encountered an error")
+                    exceptions[topic["name"]] = (
+                        {
+                            "current_error": topic["errorStatus"],
+                            "previous_error": topic["lastError"],
+                        },
+                    )
+                    topics.remove(topic["id"])
```
ghost: Let's not complicate it here and throw an error immediately instead of collecting all of them first.

tests/test_quixstreams/test_platforms/test_quix/test_config.py
```diff
@@ -622,6 +629,45 @@ def nested():
...
+                "lastError": "a previous d error",
+            },
+        ]
+
+        cfg_factory = quix_kafka_config_factory(workspace_id="12345")
+        with patch.object(cfg_factory, "get_topics") as get_topics:
+            topics = {"12345-topic_b", "12345-topic_c", "12345-topic_d"}
+            get_topics.return_value = data
+            with pytest.raises(QuixCreateTopicFailure) as e:
+                cfg_factory._finalize_create(topics)
```
ghost, line 665: Can we re-write it so we don't test the `_finalize_create` private method?

quixstreams/platforms/quix/config.py
ghost: Also, I think the semantics of "current_error" and "previous_error" are not exactly correct here.

According to the [Swagger](https://portal-api.platform.quix.ai/swagger/index.html?urls.primaryName=Portal%20API%20V1) doc, the `"errorStatus"` field seems to indicate the stage of the topic lifecycle where the error happened.  
And the `"lastError"` field contains the error description.

ghost (changes requested):

tests/test_quixstreams/test_platforms/test_quix/test_config.py
tim-quix: Yeah, I considered that, but didn't know if we wanted to make that part of a larger overhaul?

quixstreams/platforms/quix/config.py
tim-quix: I hear you on not complicating it, but I figured we should be consistent with how non-quix `create_topics` works.

Also, it might be useful when you've maybe somehow messed up multiple topics so you dont have to debug them one by one (which is also why I had the non-quix `create_topics` collects the errors too). 

Granted, I don't know if errors really manifest in the same way there, plus Quix topics do seem to take much longer to finalize.

ghost: > I hear you on not complicating it, but I figured we should be consistent with how non-quix create_topics works.

Oh, ok, I see the logic now.

Let's keep this part consistent then

ghost (approved):

tests/test_quixstreams/test_platforms/test_quix/test_config.py
ghost: Agree, let's dedicate some time to this outside this issue's scope.

PR #338, opened 2024-04-24
Bugfix: better handle Quix "External" topics

tim-quix:
Now properly handle Quix "External" topics by querying the Quix API for the `id` name when a `Topic`-based object is created, otherwise assume it requires a `workspace_id`-appended version created.

### DIFF
--- a/quixstreams/platforms/quix/config.py
+++ b/quixstreams/platforms/quix/config.py
@@ -434,7 +434,7 @@ class QuixKafkaConfigsBuilder:
             return self.api.get_topic(topic_name, workspace_id=self.workspace_id)
         except QuixApiRequestFailure as e:
             if e.status_code == 404:
-                return None
+                return
             raise
 
     def get_topics(self) -> List[dict]:
--- a/quixstreams/platforms/quix/config.py
+++ b/quixstreams/platforms/quix/config.py
@@ -17,6 +17,7 @@ from .exceptions import (
     UndefinedQuixWorkspaceId,
     QuixCreateTopicFailure,
     QuixCreateTopicTimeout,
+    QuixApiRequestFailure,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


quixstreams/platforms/quix/config.py
```diff
@@ -417,6 +418,25 @@ def create_topics(
...
+
+        :param topic_name: name of the topic
+
+        :return: response dict of the topic info if topic found, else None
+        """
+        try:
+            return self.api.get_topic(topic_name, workspace_id=self.workspace_id)
+        except QuixApiRequestFailure as e:
+            if e.status_code == 404:
+                return None
```
ghost: You can omit `None` here

quixstreams/platforms/quix/topic_manager.py
```diff
@@ -55,12 +55,14 @@ def _create_topics(self, topics: List[Topic]):
 
     def _apply_topic_prefix(self, name: str) -> str:
         """
-        Prepend workspace ID to a given topic name
+        Prepend workspace ID to a given topic name, if required.
 
         :param name: topic name
 
         :return: name with workspace ID prepended
         """
```
ghost: I think it's time to also rename this method to something like `_resolve_topic_name()` and explain what happens inside (Quix prefixing some topics).

Because this is really what happens - we're trying to understand what the real name of the topic should be.

tests/test_quixstreams/test_platforms/test_quix/fixtures.py
```diff
@@ -43,8 +43,11 @@ def mock_quix_portal_api(api_responses: dict = None):
             api_responses = {}
         api_obj = create_autospec(QuixPortalApiService)
         api_obj.default_workspace_id = None
-        for call, data in api_responses.items():
-            api_obj.__getattribute__(call).return_value = data
+        for call, result in api_responses.items():
+            if isinstance(result, Exception):
+                api_obj.__getattribute__(call).side_effect = result
+            else:
+                api_obj.__getattribute__(call).return_value = result
```
ghost: Let's avoid doing the `__getattribute__` magic inside the factory and simply mock the methods.

It doesn't really make it easier to use but requires you to also know how this factory is implemented.

This is easier to digest and use IMO
```
portal_api = MagicMock(spec_set=QuixPortalAPIService)
portal_api.get_topic.return_value = ...
portal_api.get_topic.side_effect = ...
```

ghost (changes requested):

ghost (approved):
LGTM 👍

PR #339, opened 2024-04-24
updated to use pyproject.toml; also added a few new fields and option…

tim-quix:
Also added `keywords` (for PyPi searching), a few more URLS, and `optional-dependencies` (extras) `dev` and `test`

From what I can tell, the `PKG-INFO` is practically identical to the `setup.py` counterpart besides my additions and a few syntax changes around URLS and Author info (I guess there's updated standards around how those are packaged?).

Other files also match, though the `pyproject.toml` version also includes the `requirements-dev.txt` which we could exclude.


ghost (approved):
LGTM 👍

PR #346, opened 2024-05-02
Feature: GroupBy

tim-quix:
Added GroupBy as spec'ed in the story, with some additional features (to enable GroupBy functionality).

# What is GroupBy

See documentation for more details, but in summary:

- Change/regroup message keys using `SDF.group_by()`, which makes performing stateful aggregations easier
- `.groupby()` can accept either a column name or a custom grouping function.
- You can then perform additional operations with the new key with the same `StreamingDataFrame`.

# How it Works

Added a "branches" parameter to `StreamingDataFrame` which stores multiple `StreamingDataFrame` instances in a dict, with the `SDF`'s topic name as the key.

- When an `SDF.group_by()` is called, it stores the current active `SDF` in `SDF._branches` and creates a new `SDF`, which uses a newly created GroupBy topic.
- The consumer will consume from this GroupBy topic, and the `SDF` used for processing a message depends on the topic the message came from (referenced via `SDF._branches`)

### `StreamingDataFrame`
- now uses `TopicManager`
- surfaces the possible consumable topics using new attribute `topics_to_subscribe`

### `StreamingDataFrame.test()`
- `.test` now accepts a `topic` arg so it can handle processing from any of the new `branches`.

### `TopicManager`
- cleanup around internal topic naming (especially with the addition of another type)
- added `.repartition_topic()`

## Performance

Tested/Benchmarked with python3.8 + example app to compare performance between `main` and this branch. 

The changes that enable this GroupBy behavior do not appear to have any noticeable impact on performance.


quixstreams/dataframe/dataframe.py
```diff
@@ -338,7 +443,9 @@ def test(self, value: object, ctx: Optional[MessageContext] = None) -> Any:
         context = contextvars.copy_context()
         context.run(set_message_context, ctx)
         composed = self.compose()
-        return context.run(composed, value)
+        for k in composed:
+            value = context.run(composed[k], value)
+        return value
```
ghost: Let's not alter the existing `test()` behavior here and test `group_by()` with Kafka instead.

quixstreams/dataframe/dataframe.py
```diff
@@ -506,11 +613,17 @@ def hopping_window(
...
+        stream: Optional[Stream] = None,
+        topic: Optional[Topic] = None,
+    ) -> Self:
         clone = self.__class__(
             stream=stream,
-            topic=self._topic,
+            topic=topic or self._topic,
             processing_context=self._processing_context,
+            topic_manager=self._topic_manager,
+            branches=self._branches,
```
ghost: I suggest we do a deepcopy of `self._branches` to make sure the same dict is not changed inside multiple SDFs.

quixstreams/models/topics/manager.py
```diff
@@ -92,21 +105,33 @@ def _resolve_topic_name(self, name: str) -> str:
...
         return name
 
-    def _format_changelog_name(
-        self, consumer_group: str, topic_name: str, store_name: str
+    def _validated_topic_name_length(self, name: str):
+        if len(name) > self._max_topic_name_len:
+            raise TopicNameLengthExceeded(
+                f"Topic {name} exceeds the {self._max_topic_name_len} character limit"
+            )
+        return name
```
ghost: Looks like this method doesn't change the incoming `name`, so there's no point in returning it.

I suggest renaming it to `_validate_topic_name()` and simply raising an exception if the name is invalid

quixstreams/models/topics/manager.py
```diff
@@ -92,21 +105,33 @@ def _resolve_topic_name(self, name: str) -> str:
...
+
+    def _internal_topic_name(
+        self,
+        name: str,
+        topic_name: str,
+        store_name: Optional[str] = None,
     ):
```
ghost: Func return type is missing

quixstreams/models/topics/manager.py
```diff
@@ -92,21 +105,33 @@ def _resolve_topic_name(self, name: str) -> str:
...
+    def _validated_topic_name_length(self, name: str):
+        if len(name) > self._max_topic_name_len:
+            raise TopicNameLengthExceeded(
+                f"Topic {name} exceeds the {self._max_topic_name_len} character limit"
+            )
+        return name
+
+    def _internal_topic_name(
+        self,
+        name: str,
```
ghost: Do I understand correctly, that `name` actually represents a prefix?
If yes, it should be renamed

quixstreams/models/topics/manager.py
```diff
@@ -92,21 +105,33 @@ def _resolve_topic_name(self, name: str) -> str:
...
         :param store_name: name of storage type (default, rolling10s, etc.)
 
         :return: formatted topic name
         """
-        return f"changelog__{consumer_group}--{topic_name}--{store_name}"
+        return self._validated_topic_name_length(
+            self._resolve_topic_name(
+                f"{name}__{'--'.join(filter(None, [self._consumer_group, topic_name, store_name]))}"
+            )
+        )
```
ghost: This block should be split into multiple lines for better readability.

quixstreams/models/topics/manager.py
```diff
@@ -116,6 +141,31 @@ def _create_topics(self, topics: List[Topic]):
...
+
+        :return: a TopicConfig
+        """
+        topic_config = (
+            self._admin.inspect_topics([topic_name])[topic_name]
+            or self._non_changelog_topics[topic_name].config
+        )
+
+        # Copy only certain configuration values from original topic
+        if extras_imports:
```
ghost, line 196: `extras_imports` is not documented in this method

tests/test_quixstreams/test_app.py
```diff
@@ -525,6 +525,110 @@ def test_consumer_group_default(self):
...
+        processed_count = 0
+        user_0 = "jane"
+        user_1 = "john"
+        names = [user_0] * 3 + [user_1] * 3
+        expected_message_count = len(names)
+        total_messages = expected_message_count * 2  # groupby reproduces messages...
+        app = app_factory(
+            auto_offset_reset="earliest",
+            on_message_processed=on_message_processed,
+        )
```
ghost: Let's simplify this test to one partition and one message only.
It will make the setup and validation stages much easier to follow.

Otherwise, there's too much going on without much value for the test itself.

I think we mostly need to test that the message changed its key in the context of this test.

tests/test_quixstreams/test_app.py
```diff
@@ -525,6 +525,110 @@ def test_consumer_group_default(self):
...
+
+        app_topic_in = app.topic(
+            str(uuid.uuid4()),
+            value_deserializer=JSONDeserializer(),
+            value_serializer=JSONSerializer(),
```
ghost: `"json"` can be used instead

tests/test_quixstreams/test_app.py
```diff
@@ -525,6 +525,110 @@ def test_consumer_group_default(self):
...
+            config=app._topic_manager.topic_config(num_partitions=num_partitions),
+        )
+        app_topic_out = app.topic(
+            str(uuid.uuid4()),
+            value_serializer=JSONSerializer(),
+            value_deserializer=JSONDeserializer(),
+            config=app._topic_manager.topic_config(num_partitions=num_partitions),
```
ghost: Accessing protected attributes outside of the class.

In this case, the better option is to instantiate `TopicManager` separately, pass it to `app_factory()` and use it here too.

tests/test_quixstreams/test_models/test_topics/test_manager.py
```diff
@@ -78,24 +78,23 @@ def test_changelog_topic(self, topic_manager_factory):
...
             store_name=store_name,
-            consumer_group=group,
         )
 
         assert topic_manager.changelog_topics[topic.name][store_name] == changelog
 
-        assert changelog.name == topic_manager._format_changelog_name(
-            group, topic.name, store_name
+        assert changelog.name == topic_manager._internal_topic_name(
+            "changelog", topic.name, store_name
```
ghost: Accessing protected attributes outside of the class.

quixstreams/dataframe/dataframe.py
```diff
@@ -98,6 +111,10 @@ def stream(self) -> Stream:
     def topic(self) -> Topic:
         return self._topic
 
+    @property
+    def consumer_topics(self) -> List[Topic]:
```
ghost: Is there a better name for it?

`consumer_topics` doesn't describe these topics well enough.

quixstreams/dataframe/dataframe.py
```diff
@@ -229,6 +246,79 @@ def func(d: dict, state: State):
...
+        :param value_serializer: a serializer type for values; default - JSON
+        :param key_serializer: a serializer type for keys; default - JSON
+
+        :return: a clone with this operation added (assign to keep its effect).
+        """
+        if isinstance(key, str):
+            _gb_key_op = lambda row: row[key]
+            name = name or key
+        elif callable(key):
+            _gb_key_op = lambda row: key(row)
```
ghost: Why is this variable underscored?

ghost (changes requested):

quixstreams/dataframe/dataframe.py
tim-quix: There was still a small change I needed to make here to have this behave more consistently with the old behavior, but largely removed this.

quixstreams/models/topics/manager.py
tim-quix: Moved a few more things around (to in general further simply a few things), take a look at the new version and lemme know what you think

quixstreams/models/topics/manager.py
tim-quix: tweaked this a bit, lemme know how you like the new version

quixstreams/dataframe/dataframe.py
tim-quix: I renamed this to `app_subscription_topics`...wdyt?

tim-quix:
Last remaining thing is rebuilding the docs.

mkdocs.yml
```diff
@@ -31,6 +31,7 @@ nav:
     - Purchase Filtering: 'tutorials/purchase-filtering/tutorial.md'
   - Producing Data to Kafka: producer.md
   - Processing & Transforming Data: processing.md
+  - GroupBy Operation (re-keying): groupby.md
```
ghost: Suggestion: rename to "Grouping Data by Keys". The word "re-keying" may not be familiar to everyone, and it's actually explained within the doc

quixstreams/dataframe/dataframe.py
```diff
@@ -229,6 +251,91 @@ def func(d: dict, state: State):
         stream = self.stream.add_filter(func)
         return self._clone(stream=stream)
 
+    def _groupby_key(
+        self, key: Union[str, DataFrameFunc]
+    ) -> Callable[[object], object]:
+        """
+        PEP8 hates when you assign lambdas to variables...so this is here.
```
ghost: It doesn't seem to describe what the function is doing.
Please change it or remove it

quixstreams/dataframe/dataframe.py
```diff
@@ -229,6 +251,91 @@ def func(d: dict, state: State):
...
+            )
+        groupby_topic = self._topic_manager.repartition_topic(
+            operation=name or key,
+            topic_name=self._topic.name,
+            key_serializer=key_serializer,
+            value_serializer=value_serializer,
+            key_deserializer=key_deserializer,
+            value_deserializer=value_deserializer,
+        )
+        self._finalize_branch(self.to_topic(groupby_topic, key=self._groupby_key(key)))
```
ghost: It's better to split this line in two

ghost (changes requested):
Looks good overall, a few minor changes needed 👍

ghost:
General comment about the docs: 
Some hands-on "groupby" example is needed to showcase when you need to use it and how

mkdocs.yml
tim-quix: I thought about something similar, though I also wanted to include  `GroupBy` specifically because it's a well known term.

Wanted the section to jump out to someone who has no idea what re-keying is, but does know what `GroupBy` is, whereas "Grouping Data by Keys" may not jump out to that person as readily.

If a person does know what re-keying is, then "grouping data by keys" is also kinda verbose, maybe even confusing if they've never seen re-keying described as "grouping" (though I'd imagine it's not many people, `GroupBy` is pretty commonly used to describe that...).

Maybe theres a middle ground (though maybe a bit verbose) with something like "GroupBy (grouping data by keys)"?

IDK, not super happy with any of them, haha.

ghost: Let's simply do a "GroupBy Operation" then.

I'd like to avoid adding anything in `()`. The title should be concise

ghost (approved):

PR #347, opened 2024-05-05
Producers and Consumers should accept Topics as arguments

### LINKED ISSUE #336: Producers and Consumers should accept Topics as arguments
I find myself writing a lot of code that looks like this:

```python
    weather_topic = app.topic("weather_data")
    producer.produce(topic=weather_topic.name, ...)
```

...and the ergonomics of that aren't great. I know there's a good reason to create `Topic` objects, but to a beginner it's going to feel like they're wrapping up strings just to unwrap them again.

It's a small thing, but I think it would feel nicer to program if we changed the produce and consume APIs to accept either strings or `Topic`s:

```python

# In Producer:
def produce(topic: Union[str, Topic], ...)

# In Consumer:
def subscribe(topics: List[Union[str, Topic]], ...)
```

WDTY?

shrutimantri:
Closes #336

### DIFF
--- a/quixstreams/kafka/consumer.py
+++ b/quixstreams/kafka/consumer.py
@@ -1,7 +1,7 @@
 import functools
 import logging
 import typing
-from typing import List, Optional, Callable, Tuple
+from typing import List, Optional, Callable, Tuple, Union
 
 from confluent_kafka import (
     TopicPartition,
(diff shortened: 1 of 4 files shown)


ghost:
Closing this PR for now, more info [here](https://github.com/quixio/quix-streams/issues/336#issuecomment-2135852781)

PR #348, opened 2024-05-05
Correcting the error message to be more explicit

### LINKED ISSUE #340: More helpful configuration message
One error that can come up when configuring an `Application` is this:

```
raise ValueError("Either broker address or Quix SDK Token must be provided")
```

A very easy way to make this more useful would be to tell me what code you're actually looking for:

```
raise ValueError("Either `broker_address` or `quix_sdk_token` must be provided")
```

Same meaning, but it makes my next step crystal clear.

shrutimantri:
Closes #340

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -216,9 +216,13 @@ class Application:
             quix_config_builder = QuixKafkaConfigsBuilder(quix_sdk_token=quix_sdk_token)
 
         if broker_address and quix_config_builder:
-            raise ValueError("Cannot provide both broker address and Quix SDK Token")
+            raise ValueError(
+                'Cannot provide both "broker_address" and "quix_sdk_token"'
+            )
         elif not (broker_address or quix_config_builder):
-            raise ValueError("Either broker address or Quix SDK Token must be provided")
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


quixstreams/app.py
```diff
@@ -199,9 +199,13 @@ def __init__(
...
 
         if broker_address and quix_config_builder:
-            raise ValueError("Cannot provide both broker address and Quix SDK Token")
+            raise ValueError(
+                "Cannot provide both `broker_address` and `quix_sdk_token`"
+            )
         elif not (broker_address or quix_config_builder):
-            raise ValueError("Either broker address or Quix SDK Token must be provided")
+            raise ValueError(
+                "Either `broker_address` or `quix_sdk_token` must be provided"
```
ghost: Very minor change: could you please replace accent symbols with double-quotes`"` instead?

ghost (changes requested):
Hey @shrutimantri, thanks for the contribution!

There's only one minor change needed. 
Please have a look.

Cheers.

quixstreams/app.py
shrutimantri: Addressed the comment.

ghost (approved):

PR #353, opened 2024-05-13
Add docs covering Checkpointing

ghost:


docs/advanced/checkpointing.md
```diff
@@ -0,0 +1,108 @@
+# Checkpointing
+
+To perform stateful operations reliably and with consistent outputs, Quix Streams applications need to periodically save their state stores to disk and commit processed offsets to Kafka.
```
tim-quix: One small thing here:

when we start off the document with "to perform _stateful_ operations reliably", it might imply only stateful operations (or I guess, applications that use state) use this checkpointing, which of course the Application uses checkpointing regardless.

docs/advanced/checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+
+We call this process a “checkpointing”.
+
+The goal of checkpointing is to ensure that applications can recover after failures and reprocess records from Kafka producing the same results as if the failure never happened.
```
tim-quix, line 7: Not arguing the main goal here, but increasing processing speed is also a pretty large intention of this.

This might not immediately concern the average user, but might be worth mentioning if someone is reading this doc to understand the architecture more and wonder why we only commit every so often.

If you did want to mention it, the "how" part could probably be put elsewhere to keep this section concise.

docs/advanced/checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+- When checkpoint commits, it will:
+    1. *Produce changelog messages for every pending state update to the changelog topics.*
+
+        Changelog topics are used to back up state stores in case of failures, and they contain records for every updated state key.
+        
+    2. *Flush the Kafka Producer and verify the delivery of every outgoing message both to output and changelog topics.*
+        
+        If some messages are not delivered, the checkpoint will fail.
+        
+        In this case, the messages need to be reprocessed again starting from the previous committed offset.
```
tim-quix: Not sure if you intended this to still be here thanks to your "example failures" section. 

If so, maybe just link to that section with "see X for details"

docs/advanced/checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+During recovery, these offsets are compared with the currently committed offset of the input topic.
+
+If the processed offset in the changelog record is higher than the current committed offset (i.e., the update was produced to the changelog, but the Kafka offsets failed to commit), this
+changelog update will not be applied to the state.
+
+This way, all stateful operations will work with a consistent snapshot of the state and produce the same outputs in case of reprocessing.
+
+For more information about changelog topics, see the [**How Changelog Topics Work**](stateful-processing.md#how-changelog-topics-work) section.
+
+## Example Failure Scenarios
```
tim-quix: I feel like "examples" would be a bit more...specific or nuanced in nature than what you're outlining here.

To me these are the broad (and most common) failure states that you might expect or reasonably expect to run into (and specifically, ones that we've considered and tried to cover).

That is to say, I'd probably instead call it "Common" or "Typical" Failure Scenarios, instead...with the similar caveat that it's not an exhaustive list.

docs/advanced/checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+    
+    However, it will reduce memory usage and limit the number of potentially reprocessed messages, reducing duplicates.
+    
+- If `commit_interval` is set to `0`, the application will commit a checkpoint for every processed Kafka message.
+
+When configuring the commit interval, take into account such factors as the number of unique keys in the input topics, hardware, and infrastructure.
+
+## Limitations
+
+In the At-Least-Once setting, it is still possible that unwanted changelog changes get applied during recovery from scratch.
```
tim-quix, line 88: Might be worth mentioning what the user can do to help avoid this situation, even if it might be obvious? 

ex: "don't update code without confirming the consumer successfully committed its checkpoint".

tim-quix (approved):
Small things, make whatever changes you feel like and feel free to merge after :+1:

docs/advanced/checkpointing.md
ghost: Good call, will change to a simple "To process data reliably"

docs/advanced/checkpointing.md
ghost: It's a valid point, but in this exact case I'd rather focus on one thing - reliability - and how it's been achieved.

docs/advanced/checkpointing.md
ghost: Good call, I added a recommendation.

The best way to avoid it is, ofc, EOS

PR #364, opened 2024-05-17
Correcting the Topic parameter definition

### LINKED ISSUE #363: Topic param definition does not match its usage
**Tell us about the bug**

Topic param definition does not match its usage. 
As part of this change: https://github.com/quixio/quix-streams/pull/292, the attribute was changed from Optional to required, but the corresponding documentation change was missed.

The documentation corresponding to the parameter should have changed accordingly.

N/A

**Workaround?**
Correct the documentation

shrutimantri:
Closes #363

### DIFF
--- a/quixstreams/models/topics/topic.py
+++ b/quixstreams/models/topics/topic.py
@@ -101,11 +101,11 @@ class Topic:
     ):
         """
         :param name: topic name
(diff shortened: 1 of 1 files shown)


tim-quix (approved):
Hey Shruti, nice catch! =) Approved and merging.

PR #365, opened 2024-05-20
Don't set default topics params in Quix apps

ghost:
# Problem
When an `Application` instance works with Quix Cloud and creates a new topic, it sets its number of partitions to `1`.

When creating new topics in Quix Cloud, the `Application` instance shouldn't specify a default number of partitions and should let the cloud set the defaults.

# Solution
Use `None` for `num_partitions` when creating topics in Quix Cloud.

### DIFF
--- a/quixstreams/models/topics/manager.py
+++ b/quixstreams/models/topics/manager.py
@@ -38,11 +38,14 @@ class TopicManager:
     See methods for details.
     """
 
-    _topic_partitions = 1
-    _topic_replication = 1
+    # Default topic params
+    default_num_partitions = 1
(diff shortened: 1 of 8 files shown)


tim-quix (approved):
just so we dont forget, may be a good idea to make an issue saying to remove the requests package version limit once it's fixed.

ghost:
> just so we dont forget, may be a good idea to make an issue saying to remove the requests package version limit once it's fixed.

Yep, I'm subscribed to the original issue in requests. 
We'll see what to do when it's resolved, because we might need to update a different lib instead.

Issue #354, opened 2024-05-13
Add API to provide SASL and SSL authentication 


ghost:
# Why?
To provide SASL and SSL authentication configs, users need to do it twice for Producer and Consumer via `producer_extra_config` and `consumer_extra_config`.
It may lead to duplicating the same configs and forgetting to pass them to both places is easy.

# What?
Add a way to provide authentication configs once on the Application level.
These configs will be applied to the internal producer and consumer, as well as the ones created using `Application.get_producer()` API.

Specifically, users should be able to provide these configs via Application parameters:
- `ssl.*` - via `ssl_config: SSLConfig` param
- `sasl.*` - via `sasl_config: SASLConfig` param
- `security.protocol` - via `security_protocol: str` parameter

# How?
- Use `pydantic` to describe `SSLConfig` and `SASLConfig` models to incorporate all individual values
- Ignore the non-provided values to rely on `librdkafka`'s defaults
- Values provided via `producer_extra_config` and `consumer_extra_config` should override the above
- If `quix_sdk_token` is passed, these parameters should be ignored

tim-quix:
completed with PR #371

PR #371, opened 2024-05-23
Feature: ConnectionConfig

tim-quix:
## What

Provide a more convenient way for users to connect to Kafka clusters that require authentication.

## How

Introduce a new `ConnectionConfig` class with all possible connection settings, used primarily by `Application`, which shares it with whatever objects require it.

As such, it is also accepted by  `Producer`, `Consumer`, and `TopicAdmin` classes.

`ConnectionConfig` is a `Pydantic` class with auto-complete, type hints, and a few convenience methods (like `from_librdkafka_dict`) for easier config setup.

`producer/consumer_extra_config` will override any provided `ConnectionConfig` settings.

## Usage Example

See [new authentication documentation](https://github.com/quixio/quix-streams/blob/feature/broker-config-dict/docs/configuration.md#authentication) for more details

## Backwards Compatibility

This addition is still backwards compatible with the previous string argument approach:

```
app = Application(broker_address="localhost:9092")
```

## Other Changes

- Removed rarely used parameters from `Application` + consumer/producer (assignment strategy, partitioner)
- Added arguments for logger and callbacks for producer/consumer so users could change them
- Improved `QuixConfigBuilder` config extraction to use new API endpoint, which simplified it.


quixstreams/kafka/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+
+    model_config = SettingsConfigDict(
+        alias_generator=AliasGenerator(
+            # used during model_dumps
+            serialization_alias=lambda field_name: field_name.replace("_", "."),
+        ),
+    )
+
+    @classmethod
+    def secret_fields(cls) -> List[str]:
```
ghost: Two questions here:
1. Why does it need to be a `classmethod`? It's only used from the normal instance method.  
It looks like it could be a `_secret_fields` property instead.

2. Why is it public? I don't see it being used anywhere outside the class

quixstreams/kafka/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+        Safely obscures the any "secret" fields (passwords).
+
+        :param indent: the indent of the JSON
+
+        :return: a string-dumped JSON
+        """
+        return self.model_dump_json(by_alias=True, exclude_none=True, indent=indent)
+
+    def __str__(self) -> str:
+        return self.as_printable_json()
```
ghost: I'm unsure if returning JSON in `__str__` is appropriate here because JSON contains additional formatting and linebreaks.

I suggest using a dict instead and cast it to a string:
```
>>> str(self.model_dump(by_alias=True, exclude_none=True))
# Prints "{'bootstrap.servers': '123', 'sasl.password': SecretStr('**********')}"
```

quixstreams/kafka/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        Dump any non-empty config values as a librdkafka-formatted JSON string.
+
+        Safely obscures the any "secret" fields (passwords).
+
+        :param indent: the indent of the JSON
+
+        :return: a string-dumped JSON
+        """
+        return self.model_dump_json(by_alias=True, exclude_none=True, indent=indent)
```
ghost: Do we need this method at all? 

It may seem nice, but if we don't have a clear use for it, I'd rather remove it to reduce the surface of the public API.

quixstreams/kafka/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+    ssl_keystore_location: Optional[str] = None
+    ssl_keystore_password: Optional[SecretStr] = None
+    ssl_providers: Optional[str] = None
+    ssl_engine_location: Optional[str] = None
+    ssl_engine_id: Optional[str] = None
+    enable_ssl_certificate_verification: Optional[bool] = None
+    ssl_endpoint_identification_algorithm: Annotated[
+        Optional[Literal["none", "https"]],
+        BeforeValidator(lambda v: v.lower() if v is not None else v),
+    ] = None
```
ghost: What's the reason for putting the attributes after the methods in this class?
I know the list is big, but let's stick to the usual approach when the class attributes go before the methods

quixstreams/kafka/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+from pydantic.functional_validators import BeforeValidator
+from pydantic_settings import BaseSettings, SettingsConfigDict
+from typing_extensions import Self, Annotated
+
+__all__ = ("ConnectionConfig",)
+
+MECHANISM_ALIAS = AliasChoices("sasl_mechanism", "sasl_mechanisms")
+
+
+class ConnectionConfig(BaseSettings):
```
ghost, line 17: There's one extra thing I'd add to the class.
By default, `BaseSettings` will also pull values from the environment, which is not this class's primary goal.
The parameters are supposed to be passed directly.

To turn off this behavior, we need to add an extra method:
```
@classmethod
    def settings_customise_sources(
        cls,
        settings_cls: Type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        return (init_settings,)
```

It will make the `BaseSettings` class ignore every other source of settings except `__init__`.
Otherwise, there may be clashes if the env variable with the same name is defined

quixstreams/platforms/quix/config.py
```diff
@@ -185,6 +147,26 @@ def workspace_meta(self) -> dict:
             self.get_workspace_info()
         return self._workspace_meta
 
+    @property
+    def librdkafka_connection_config(self) -> ConnectionConfig:
+        if not self._librdkafka_connection_config:
+            self._get_librdkafka_connection_config()
+        return self._librdkafka_connection_config
```
ghost: What if we simply get the librdkafka on `__init__`?
This way we wouldn't need to update the `self._librdkafka_connection_config` attribute.

quixstreams/platforms/quix/config.py
```diff
@@ -531,84 +522,38 @@ def _set_workspace_cert(self, timeout: Optional[float] = None) -> str:
...
-
-    def get_confluent_client_configs(
-        self,
-        topics: list,
-        consumer_group_id: Optional[str] = None,
-        timeout: Optional[float] = None,
-    ) -> Tuple[dict, List[str], Optional[str]]:
+    def get_application_config(
+        self, consumer_group_id: str
+    ) -> Tuple[ConnectionConfig, dict, Optional[str]]:
```
ghost: I think we could return here some object with named properties instead, e.g. some dataclass `QuixApplicationConfig`.
It would make the result more readable.

tests/test_quixstreams/fixtures.py
```diff
@@ -364,6 +360,16 @@ def factory(workspace_id: Optional[str] = None):
         cfg_builder.get_topic.side_effect = lambda topic: {
             "id": cfg_builder.prepend_workspace_id(topic)
         }
+
+        connection = ConnectionConfig(bootstrap_servers=kafka_container.broker_address)
+        cfg_builder._librdkafka_connection_config = connection
+        cfg_builder.librdkafka_connection_config = connection
```
ghost: Do we need to patch the `_librdkafka_connection_config` here too?
Can we somehow patch only the public method?

ghost (changes requested):
I did the review, please see the list of changes.
One other thing to not forget: please add licenses for `pydantic` and `pydantic_settings` to the `LICENSES` dir.

quixstreams/platforms/quix/config.py
tim-quix: I def agree with the idea in general since we probably are using this class for that purpose 99% of the time.

Unfortunately it's a bit linked to the `get_workspace_id`, which itself could throw an exception if the token was a PAT and no Workspace ID was defined (and thus has multiple possible workspace IDs...this situation requires users to pass a topic to `get_workspace_id` manually to search for the proper workspace ID). Not that I think anyone is doing anything that way right now, but it is an issue regardless.

If we set it up in `__init__`, it wont allow creating an instance at all in that circumstance due to said exception.

TL;DR This should probably be part of a separate story to clean up this class (and its workspace management in general), as we already have noted before.

Thoughts?

quixstreams/kafka/configuration.py
tim-quix: I mostly had it that way since arguably it doesn't need an instance to use it...but it's def only gonna be used internally by an instance method, which prob means it doesn't need to be a `classmethod` itself :+1:

quixstreams/kafka/configuration.py
tim-quix: Agreed! I came up with something different to accomplish a similar thing with `as_librdkafka_dict`, feel free to take a look.

ghost (approved):

tim-quix:
closes issue #354

PR #375, opened 2024-05-27
Update the propagation of keys, timestamps and headers in StreamingDataFrame

ghost:
# Problem
Working with timestamps, keys, and headers in `StreamingDataFrames` has been complicated.

Specifically:

- To access the key or timestamp for the current message, users need to import a particular object - `MessageContext`- a global object scoped to the lifetime of a single message.
- Because of that, It's impossible to change the timestamp during message processing when 'expand=True` functions.
- Messages produced to the output topics don't preserve the timestamps of the original records.
- Results of windowed aggregations also use the current message timestamp instead of the window timestamps.
- The message headers are ignored

# What's changed

## 1. A new `Stream.compose()` API and behavior

A callable return by `Stream.compose()` now accepts value, key, and timestamp and returns `None`

This is due to changes in the composition and execution of the internal pipeline. 

Previously, all functions in the `Stream` were nested one into another. 
They returned values immediately, and the composed stream returned the result of the transformation.

Now, each function is responsible for calling the next one with the results of the transformations. 
This way, functions can control the execution flow and decide which data to pass and how.
For example, the `filter()` implementation now skips calling the following function instead of raising the `Filtered` exception.

## 2. A new type of function in `Stream` - "transform" (`Stream.add_transform()`)

"Transform" is a new kind of function in `Stream`.  
It provides an API to change not only values but keys and timestamps, too.

"Transform" functions are used in two places:

-  windowed aggregations - when the window returns the result, the window start timestamp is set as a current timestamp
- `StreamingDataFrame.set_timestamp` API - to change the timestamps

Because "Transform" can also change the records' keys, it's not exposed on the `StreamingDataFrame` level.

##  3. New `metadata` parameter in `apply()`, `update()` and `filter()` methods of `StreamingDataFrame`

When `metadata=True` is passed to any of the methods above, the callback will receive a key, timestamp, and headers along with the value.
This is now the only way to access keys, timestamps, and headers.

Example: 
```
# Accessing the current key and timestamp in the .apply() function
sdf.apply(lambda value, key, timestamp, headers: value, metadata=True)
```

## 4. Breaking changes in `MessageContext`

Attributes `key`, `timestamp`, and `headers` have been removed from the `MessageContext` class 
To access metadata, the `metadata=True` API should be used (see above).

Stateful functions and windows also take keys and timestamps from the parameters.

`MessageContext` remains to access things like the current message's topic, partition, and offset.
These attributes remain static during processing.

## 5. New API to update timestamps  - `StreamingDataFrame.set_timestamp()`
The timestamps can now be updated at any point during processing.

Example:

```
sdf = sdf.set_timestamp(lambda value, key, timestamp, headers: <new timestamp in millis>
```

## 6. `StreamingDataFrame.to_topic()` now forwards the current timestamp to the output topics.
All outgoing messages will now have the timestamps set to the current timestamp.
It will improve the data consistency across the topics. 

To keep the previous behavior, users may use'set_timestamp` and set the timestamp to the current UTC epoch:
```
# Set timestamp to a current epoch
sdf = sdf.set_timestamp(lambda value, timestamp: int(time.time() * 1000)
```

## 7. Timestamps in Windows
When the tumbling or hopping window returns the result, the window start timestamp is set as a current timestamp.  

This way, windows become independent of the message timestamps and can now be chained together.

## 9. The `expand` parameter is removed from `.final()` in Windows
Previously, the final results could be returned in batches.  
Now, it's simply not possible because each aggregation result may have a different timestamp.
Plus, this API doesn't seem to be useful at all.


quixstreams/dataframe/windows/time_based.py
```diff
@@ -139,44 +158,44 @@ def current(self, expand: bool = True) -> "StreamingDataFrame":
...
+        return self._apply_window(func=window_callback, name=self._name)
 
     def register_store(self):
         self._dataframe.processing_context.state_manager.register_windowed_store(
             topic_name=self._dataframe.topic.name, store_name=self._name
         )
 
-    def _stale(self, window_end: int, latest_timestamp: int) -> bool:
-        return latest_timestamp > window_end + self._grace_ms
+    # def _stale(self, window_end: int, latest_timestamp: int) -> bool:
```
tim-quix: commented out code

docs/processing.md
```diff
@@ -549,36 +557,82 @@ sdf = sdf.update(lambda value: pdb.set_trace())
 sdf = sdf.update(lambda value: print('Value: ', value))
 ```
 
-## Accessing Kafka Keys and Metadata
+## Updating Timestamps
+
+In Quix Streams, each processed item has a timestamp assigned.
```
tim-quix, line 562: Might not hurt to mention that it is the same as the Kafka message timestamp for those familiar with it.

quixstreams/dataframe/dataframe.py
```diff
@@ -408,21 +574,44 @@ def to_topic(
...
+                topic=topic,
+                value=value,
+                key=orig_key if key is None else key(value),
+                timestamp=timestamp,
+            ),
+            metadata=True,
         )
 
-    def _finalize_branch(self, branch: Self):
+    def set_timestamp(self, func: Callable[[Any, int], int]) -> Self:
```
tim-quix: Probably should include an example of using it in the docstring =)

quixstreams/core/stream/functions.py
```diff
@@ -120,64 +168,146 @@ class FilterFunction(StreamFunction):
...
+                func: TransformCallback = self.func,
+            ):
+                # Execute a function on a single value and return its result
+                new_value, new_key, new_timestamp = func(value, key, timestamp)
+                child_executor(new_value, new_key, new_timestamp)
+
+        return wrapper
 
 
 def compose(
```
tim-quix: There's a couple things I'll mention here:

First, I wonder if it makes sense to just have this live in the `Stream` class (along with the default `sink` function) since you generally aren't composing these functions on their own.

If so, you could potentially simplify it a bit (I didn't test it thoroughly but with a basic example this equivalently worked):

```
    def compose(
        self,
        sink: Callable[[Any, Any, int], None] = _default_sink,
        allow_filters: bool = True,
        allow_updates: bool = True,
        allow_expands: bool = True,
        allow_transforms: bool = True,
    ) -> VoidExecutor:
        tree = self.tree()
        tree.reverse()
        composed = sink
        for node in tree:
            func = node.func
            # Validate that only allowed functions are passed
            if not allow_updates and isinstance(
                    func, (UpdateFunction, UpdateWithMetadataFunction)
            ):
                raise ValueError("Update functions are not allowed")
            elif not allow_filters and isinstance(
                    func, (FilterFunction, FilterWithMetadataFunction)
            ):
                raise ValueError("Filter functions are not allowed")
            elif not allow_transforms and isinstance(func, TransformFunction):
                raise ValueError("Transform functions are not allowed")
            elif not allow_expands and func.expand:
                raise ValueError("Expand functions are not allowed")

            composed = func.get_executor(composed)

        return composed
```

That said, two things: 

1. I also mentioned to you how we should consider removing the extra functions (like the initializing lambda and having a default sink), which this new version would not work as nicely with.

2. once we start doing actual logic branching, it might need to remain something more generalized/independent of the trees for splitting, but hard to say yet what that will look like.

In either case, still might consider moving it to `Stream`...but in all it's a minor thing that may change later anyway.

quixstreams/core/stream/stream.py
```diff
@@ -44,23 +58,28 @@ def __init__(
...
+            It is expected to be wrapped into one of "Apply", "Filter", "Update" or
+            "Trasform" from `quixstreams.core.stream.functions` package.
+            Default - "ApplyFunction(lambda value: value)".
         :param parent: a parent `Stream`
         """
         if func is not None and not isinstance(func, StreamFunction):
             raise ValueError("Provided function must be a subclass of StreamFunction")
 
-        self.func = func if func is not None else ApplyFunction(lambda x: x)
+        self.func = func if func is not None else ApplyFunction(lambda value: value)
```
tim-quix, line 81: As I mentioned to you, wonder if we could consider either:

1. an `IdentityFunction` that you could filter during compose and/or tree,
2. or instead, no default function at all, which to me seems reasonable since it means you haven't added any ops yet.

tim-quix (approved):
Minor comments...easy to follow and make sense of :+1:

tim-quix:
Of course, dont forget doc rebuilds n such =)

quixstreams/core/stream/functions.py
ghost: 1. I like moving `compose` to `Stream` and simplifying it. I'll change it. Nice catch.

2. I agree that removing the placeholder functions makes sense overall, but I plan to deal with it in future versions. For example, when implementing the branching.
Changing this part doesn't seem essential for this PR.

quixstreams/core/stream/stream.py
ghost: Replied here https://github.com/quixio/quix-streams/pull/375#discussion_r1633578461

quixstreams/core/stream/functions.py
tim-quix: Agreed, I think that's reasonable :+1:

tim-quix (approved):

PR #378, opened 2024-05-31
Fix flaky tests that rely on kafka in docker

### LINKED ISSUE #380: Tests using Kafka container are flaky
**Tell us about the bug**
After implementing StreamingDataFrame.group_by(), some tests using an actual Kafka container become flaky.

Particularly this one - `TestAppGroupBy.test_group_by `.

The tests should pass all the time if the code is correct.

2.5.1

ghost:
Closes #380 

# Problem
After implementing `StreamingDataFrame.group_by()`, some tests using an actual Kafka container become flaky.

Although the root cause is not 100% clear, @tim-quix reproduced the problem with the current docker image used in tests.

# What's changed
- The tests now use the official `confluentinc/cp-kafka` Docker image.  
Changing the image fixed the issue.  
This image also uses KRaft instead of Zookeeper by default, simplifying the overall setup.
- The logging level is now parameterized via the `--log-cli-level=<loglevel>` parameter in tests.  
- Reduced the log level in CI to `ERROR`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,4 +60,4 @@ jobs:
           python -m pip install -U -r requirements.txt
       - name: Run tests
         run: |
-          python -m pytest -s -v
+          python -m pytest -v --log-cli-level=ERROR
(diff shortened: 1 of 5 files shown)


tim-quix (approved):
Looks good, ran locally and tests pass :+1:

PR #382, opened 2024-06-06
Configure the producer flush timeout with the max.poll.interval.ms

### LINKED ISSUE #381: Set timeouts in the Checkpoint
## Problem

Checkpointing synchronously calls `Producer.flush()` and `Consumer.commit()`.
Both methods do network calls and may take some time to complete.

If something goes wrong during checkpointing, the consumer may get kicked out of the consumer group while still trying to produce messages and commit.

## Solution
- Set some reasonable timeouts for flush() and commit()
- If any of the timeouts are exceeded, fail with a nice error message and abort the checkpoint.
- Asses if the default value for `session.timeout.ms` should be increased from 45s.

ghost:
closes #381 

Add a configuration option to set a default flush timeout on producers. This ensure the application is not blocked if messages can't be delivered successfully.

`consumer.commit` doesn't explicitly support a timeout parameter but rely on a globally configured timeout (probably socket.timeout, TBC).

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -50,6 +50,8 @@ MessageProcessedCallback = Callable[[str, int, int], None]
 # Enforce idempotent producing for the internal RowProducer
 _default_producer_extra_config = {"enable.idempotence": True}
 
+_default_max_poll_interval_ms = 300000
+
 
 class Application:
     """
@@ -260,6 +262,10 @@ class Application:
             broker_address=broker_address,
             extra_config=producer_extra_config,
             on_error=on_producer_error,
+            flush_timeout=consumer_extra_config.get(
+                "max.poll.interval.ms", _default_max_poll_interval_ms
+            )
+            / 1000,  # convert to seconds
         )
         self._consumer_poll_timeout = consumer_poll_timeout
         self._producer_poll_timeout = producer_poll_timeout
--- a/quixstreams/checkpointing/checkpoint.py
+++ b/quixstreams/checkpointing/checkpoint.py
@@ -12,7 +12,7 @@ from quixstreams.state import (
     DEFAULT_STATE_STORE_NAME,
 )
 from quixstreams.state.exceptions import StoreTransactionFailed
-from .exceptions import InvalidStoredOffset
+from .exceptions import InvalidStoredOffset, CheckpointProducerTimeout
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


quixstreams/app.py
```diff
@@ -260,6 +260,8 @@ def __init__(
             broker_address=broker_address,
             extra_config=producer_extra_config,
             on_error=on_producer_error,
+            flush_timeout=consumer_extra_config.get("max.poll.interval.ms", 300000)
```
ghost: Suggestion: let's put `300000` to a protected variable somewhere in the module so the constant has a name

quixstreams/checkpointing/checkpoint.py
```diff
@@ -129,7 +129,11 @@ def commit(self):
 
         # Step 2. Flush producer to trigger all delivery callbacks and ensure that
         # all messages are produced
-        self._producer.flush()
+        logger.debug("checkpoint: flushing producer")
```
ghost: Please use capital case in the log messages

quixstreams/checkpointing/checkpoint.py
```diff
@@ -129,7 +129,11 @@ def commit(self):
...
+        failed_count = self._producer.flush()
+        if failed_count:
```
ghost: I'm unsure if `failed_count` is the best name for what's happening here.
`flush()` returns the number of messages left in the producer's queue, but at this point, it's unknown if they have failed or not.

I suggest renaming it to `unproduced_count` or similar

quixstreams/checkpointing/checkpoint.py
```diff
@@ -129,7 +129,11 @@ def commit(self):
...
+            raise IncompleteFlush(failed_count)
```
ghost: A couple of notes here:
1) The name `IncompleteFlush` doesn't clearly explain to the user that the checkpoint has timed out because of the producer. Let's rename it to `ProducerFlushTimeout` or even `CheckpointProducerTimeout` for better clarity.

2) Please add some text explaining what's going on to the exception.

quixstreams/kafka/producer.py
```diff
@@ -151,11 +154,16 @@ def flush(self, timeout: Optional[float] = None) -> int:
...
-            None or -1 is infinite. Default: None
+            None use producer default or -1 is infinite. Default: None
 
         :return: number of messages remaining to flush
         """
-        return self._producer.flush(timeout=timeout if timeout is not None else -1)
+        if timeout is None:
+            timeout = self._flush_timeout
+        if timeout is None:
+            timeout = -1
```
ghost: It looks like one of the `None` checks can be moved to the `Producer.__init__` to be done once.

ghost (changes requested):
Hey @quentin-quix , looks good overall. 
There are only a few minor things to address 👍

ghost (approved):

PR #384, opened 2024-06-11
Bug: configuring broker_address take priority over quix config

### LINKED ISSUE #376: Conflicting settings with passed arguments and environment variables
This may be considered a bug, but sits on the fence, so raising as a feature request.
When I'm deploying a quixstreams application in the Quix Cloud, there are certain variables my deployment gets injected. This is done, so setup and boiler plate is reduced for most use cases. The problem I'm facing is when I have a setup where I consume from/produce to an external broker, with the other end being the environment's broker. (well, the latter part is optional for this issue)
Example:
```
app = Application(broker_address=broker_address, ...otherconfigs...) # setting up app against external source
input_topic = app.topic(input_topic_name)

producer_app = Application() # setting up app against environment broker
output_topic = producer_app.topic(output_topic_name)
```

I get the exception:
```
Traceback (most recent call last):
  File "/app/main.py", line 35, in <module>
    app = Application(broker_address=broker_address, ....otherconfigs...)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/quixstreams/app.py", line 219, in __init__
    raise ValueError(
ValueError: Cannot provide both "broker_address" and "quix_sdk_token"
```

I see two immediate solutions: 
1) prioritise argument based config over environment variables. Here I didn't pass sdk token, so it should ignore the one coming from environment variables, as my argument configuration clearly prefers broker address. Maybe log a warning/info.
2) let me pass quix_sdk_token as None, so I can turn off this behaviour for an application explicitly.  

**Additional context**
Tried with quixstreams 2.4.1, 2.5.1, same exception.

ghost:
closes https://github.com/quixio/quix-streams/issues/376

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -120,6 +120,7 @@ class Application:
             Accepts string with Kafka broker host and port formatted as `<host>:<port>`,
             or a ConnectionConfig object if authentication is required.
             Either this OR `quix_sdk_token` must be set to use `Application` (not both).
+            Takes priority over quix auto-configuration.
             Linked Environment Variable: `Quix__Broker__Address`.
             Default: `None`
         :param quix_sdk_token: If using the Quix Cloud, the SDK token to connect with.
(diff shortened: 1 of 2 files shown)


quixstreams/app.py
```diff
@@ -197,27 +198,32 @@ def __init__(
...
+            if not quix_config_builder:
+                if quix_sdk_token:
+                    quix_app_source = "Quix SDK Token"
+                    quix_config_builder = QuixKafkaConfigsBuilder(
+                        quix_sdk_token=quix_sdk_token
+                    )
+                else:
+                    raise ValueError(
+                        'Either "broker_address" or "quix_sdk_token" must be provided'
+                    )
```
ghost: I like the approach overall, i.e., for the `broker_address` to come before `quix_sdk_token` if it's defined.

On the coding side, can we reduce the number of the `if` clauses here and make the conditions more straightforward?

For example, `quix_config_builder` variable is checked in many places and maybe we can have fewer checks

ghost (changes requested):

tests/test_quixstreams/test_app.py
```diff
@@ -748,12 +748,13 @@ def get_cfg_builder(quix_sdk_token):
...
             "quixstreams.app.RowProducer"
         ) as producer_init_mock:
-            Application(
+            app = Application(
                 consumer_group=consumer_group,
                 quix_sdk_token=quix_sdk_token,
                 consumer_extra_config=extra_config,
                 producer_extra_config=extra_config,
             )
+            assert app._is_quix_app
```
ghost: Hey, please avoid calling the protected variables outside of the instance methods.

Is there another way to verify it works without checking `._is_quix_app`?

ghost (changes requested):

ghost (approved):

PR #385, opened 2024-06-11
Exactly Once Processing

### LINKED ISSUE #139: Exactly once processing semantic
In case of a restart, no message get sent twice to the output topic. 

TBD

tim-quix:
Adds "exactly once" processing and removes changelog message offsets adjustments so that `exactly-once` and `at-least-once` recovery and changelog management are handled in similar fashions (and thus you can swap between them).

This offset change does mean that recovery attempts will be made during every rebalance, though they'll often not recover anything.

There was also some general cleanup made to `RecoveryPartition` and how they are revoked since changes were needed anyway.

## How to use `exactly-once`

```python

app = Application(processing_guarantee="exactly-once")
```

Closes #139


quixstreams/app.py
```diff
@@ -180,6 +183,18 @@ def __init__(
...
+        if processing_guarantee in get_args(ExactlyOnceSemantics):
+            exactly_once = True
+        elif processing_guarantee in get_args(AtLeastOnceSemantics):
+            exactly_once = False
+        else:
+            raise ValueError(
+                f'Must provide a valid "processing_guarantee"; expected: '
+                f"{[*get_args(ExactlyOnceSemantics), *get_args(AtLeastOnceSemantics)]}"
+                f', got "{processing_guarantee}"'
+            )
```
ghost: Are you sure we need multiple values for each processing guarantee?
It complicates the code (need to call `get_args` several times) and the API (more values to choose from for no obvious reason).

For example:
```
EXACTLY_ONCE = 'exactly-once'
AT_LEAST_ONCE = 'at-least-once'

PROCESSING_GUARANTEES = [EXACTLY_ONCE, AT_LEAST_ONCE]
ProcessingGuaranteesChoice = Literal[*PROCESSING_GUARANTEES]

if value == AT_LEAST_ONCE:
    ...
elif value == EXACTLY_ONCE: 
    ...
else:
   raise ...
```

quixstreams/app.py
```diff
@@ -700,19 +721,22 @@ def run(
         """
         self._setup_signal_handlers()
 
+        guarantee = "exactly-once" if self._exactly_once else "at-least-once"
```
ghost: Can we take this string from the provided `__init__` params?

quixstreams/checkpointing/checkpoint.py
```diff
@@ -114,7 +119,9 @@ def commit(self):
         """
 
         if not self._tp_offsets:
-            # No messages have been processed during this checkpoint, return
+            logger.debug("Nothing to commit")
```
ghost: Adding a log here is a good idea, but can you make the message look similar to the others in this function?

See below:
```logger.debug("Checkpoint: <message>")```

File: quixstreams/processing_context.py [hunk 4]
```diff
@@ -70,13 +73,19 @@ def commit_checkpoint(self, force: bool = False):
 
         :param force: if `True`, commit the checkpoint before its expiration deadline.
         """
-        if not self._checkpoint.empty() and (self._checkpoint.expired() or force):
-
-            logger.debug(f"Committing a checkpoint force={force}")
+        if self._checkpoint.expired() or force:
```
ghost, line 75: Why did you remove the `not self._checkpoint.empty()` check here?

Without it, the code below will be executed after every `.poll()`, even if no messages were processed, and pollute the logs.

quixstreams/processing_context.py
```diff
@@ -70,13 +73,19 @@ def commit_checkpoint(self, force: bool = False):
...
+                f"Committed a checkpoint; forced={force}, time_elapsed={elapsed}s"
             )
             self.init_checkpoint()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.exactly_once:
+            self.producer.abort_transaction(5)
```
ghost, line 93: Do we need to abort the transactions in the `ProcessingContext` class instead of the `RowProducer` or `TransactionalProducer`?

I see this code in the `RowProducer` class:
```if self._active_transaction:
            self._producer.abort_transaction(timeout)
            self._active_transaction = False
```

So it's unclear why the ProcessingContext should trigger the abort instead of the producer doing it directly.

quixstreams/rowproducer.py
```diff
@@ -160,6 +178,82 @@ def flush(self, timeout: Optional[float] = None) -> int:
...
+                    )
+                    raise
+        raise KafkaProducerTransactionCommitFailed(
+            f"All Kafka {op_name} attempts failed; "
+            "aborting transaction and shutting down Application..."
+        )
+
+    def commit_transaction(
+        self,
+        positions: List[TopicPartition],
```
ghost, line 305: Why `positions` and not `offsets`?

quixstreams/rowproducer.py
```diff
@@ -160,6 +178,82 @@ def flush(self, timeout: Optional[float] = None) -> int:
...
+        NOTE: under normal circumstances a transaction will be open due to how
+        the Checkpoint inits another immediately after committing.
+        """
+        if self._active_transaction:
+            self._producer.abort_transaction(timeout)
+            self._active_transaction = False
+        else:
+            logger.debug("No Kafka transaction to abort")
+
+    def _retriable_commit_op(self, operation: Callable, args: list):
```
ghost: It looks like this function can be converted to a decorator and nicely wrap the `send_offsets_to_transaction` and `commit_transaction` methods

Example:

```
@retry_commit(attempts=3, backoff_sec=1)
def commit_transaction(...):
```

quixstreams/state/recovery.py
```diff
@@ -42,6 +42,11 @@ def __init__(
         self._changelog_lowwater: Optional[int] = None
         self._changelog_highwater: Optional[int] = None
         self._committed_offset = committed_offset
+        self._recovery_consume_position: Optional[int] = None
+        self._initial_offset: Optional[int] = None
+
+    def __str__(self):
+        return f"{self.changelog_name}[{self.partition_num}]"
```
ghost: Why do we need this code here? 
Do we convert instances of `RecoveryPartition` to strings somewhere?

If it's for debugging purposes, maybe `__repr__` should be used instead.

quixstreams/state/recovery.py
```diff
@@ -94,13 +108,13 @@ def update_offset(self):
...
+        if self.offset >= self._changelog_highwater:
             logger.warning(
-                f"{self.changelog_name}[{self.partition_num}] - the changelog offset "
-                f"{self.offset} in state was larger than its actual highwater "
+                f"{self} - the changelog offset "
+                f"{self.offset} in state was greater or equal to its actual highwater "
                 f"{self._changelog_highwater}, possibly due to previous Kafka or "
                 f"network issues. State may be inaccurate for any affected keys. "
-                f"The offset will now be set to {self._changelog_highwater}."
+                f"The offset will now be set to its highwater - 1."
```
ghost: Is the changelog offset adjusted somewhere when it's ahead?

ghost (changes requested):

quixstreams/processing_context.py
tim-quix: I did it there because transaction lifecycles/management is pretty much fully dictated by the Checkpointing/ProcessingContext, so it seemed like it made the most sense to be alongside it.

That said, I originally was gonna put it with the RowProducer, but it felt kinda weird to put it there when none of the other transaction tracking was managed by it.

I don't really feel strongly about it one way or the other, so whichever you prefer.

quixstreams/rowproducer.py
tim-quix: I think we discussed this already but just following the naming convention used with librdkafka.

Seems to be the term they use in general to refer to an uncommitted consumer offset read location.

ghost: Yep, I looked up the docs for `send_offsets_to_transaction,`, and it uses the name `positions.`
Let's stick to it then 👍

quixstreams/processing_context.py
ghost: Ok, that makes sense.

Besides, we cannot easily do it in `RowProducer` because `RowProducer.__exit__` triggers `.flush()`, which is managed in the Checkpoint when running with the `Application`.

quixstreams/state/recovery.py
tim-quix: it's used in a couple places as debug, and just once for info logging (informing recovery is complete for said RP, which I think `info` is correct level there since actual recovery is generally rare).

quixstreams/state/recovery.py
tim-quix: removed all offset setting

quixstreams/processing_context.py
tim-quix: Tweaked the approach to be a little closer to the original while still accommodating the transaction aborting part...let me know what you think. Tried to avoid making logging overly noisy (mostly applied to debug mode).

Also I know you're potentially playing with this with sinks anyway, so I'm sure it'll change again soon =)

quixstreams/checkpointing/checkpoint.py
tim-quix: This was adjusted based on the other change to committing.

quixstreams/app.py
tim-quix: Adjusted this a bit along with the processing_guarantee handling.

quixstreams/state/recovery.py
ghost: I think using `__repr__` here is more correct from the semantics point of view if it's used only for debugging.

ghost (approved):

PR #386, opened 2024-06-13
Check partitions status after committing in checkpoint

ghost:


quixstreams/checkpointing/checkpoint.py
```diff
@@ -145,7 +149,14 @@ def commit(self):
             for (topic, partition), offset in self._tp_offsets.items()
         ]
         logger.debug("Checkpoint: commiting consumer")
-        self._consumer.commit(offsets=offsets, asynchronous=False)
+        try:
+            commit_statuses = self._consumer.commit(offsets=offsets, asynchronous=False)
+        except KafkaException as e:
+            raise CheckpointConsumerCommitError(e.args[0]) from None
+        else:
```
ghost: Hey @quentin-quix , a couple of things here:

1. Is it necessary to have an `else:` here? Looks like we could avoid it.
2. `commit_statuses` - the `commit()` call returns a list of `TopicPartition` objects. Let's name the variable accordingly, e.g. `partitions` or `tps` (short for "topic-partitions").

ghost (changes requested):

ghost (approved):

PR #387, opened 2024-06-13
Add `StreamingDataFrame.set_headers()` API

ghost:
# What's changed

**NOTE: this PR is made to `fix/sdf-timestamp-propagation` and it waits #375  to be merged.**

1. Added a new method to set message headers -`StreamingDataFrame.set_headers()`.
2. Updated docs

Example usage:
```python
sdf = app.dataframe(...)

APP_VERSION = "v0.1.1"

# Add the value of APP_VERSION to the message headers for debugging purposes.  
# The callback receives four positional arguments: value, key, current timestamp, and headers. 
# It must return a new set of headers as a list of (header, value) tuples.

sdf = sdf.set_headers(
    lambda value, key, timestamp, headers: [('APP_VERSION', APP_VERSION.encode())]
)
```

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -16,6 +16,7 @@ from typing import (
     Dict,
     Tuple,
     Literal,
+    Collection,
 )
 
 from typing_extensions import Self
@@ -39,6 +40,7 @@ from quixstreams.models import (
     TopicManager,
     Row,
     MessageContext,
+    HeaderValue,
 )
 from quixstreams.models.serializers import SerializerType, DeserializerType
 from quixstreams.processing_context import ProcessingContext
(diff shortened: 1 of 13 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -620,6 +622,57 @@ def _set_timestamp_callback(
...
+        input_topic = app.topic("data")
+
+        sdf = app.dataframe(input_topic)
+        # Updating the record's headers based on the value and metadata
+        sdf = sdf.set_headers(lambda value, key, timestamp, headers: [('id', value['id'])])
+        ```
+
+        :param func: callable accepting the current value, key, timestamp, and headers.
+            It's expected to return a new set of headers
+            as a collection of (header, value) tuples.
```
tim-quix, line 656: May not hurt to mention valid types, like `value` can be `bytes` or `str` since it may not be super obvious given their original format mentioned in the docs.

tim-quix (approved):

PR #393, opened 2024-07-02
add drop feature

tim-quix:
Add the ability to drop a column or list of columns from a `StreamingDataFrame` via:

 `SDF.drop("col_a")` or `SDF.drop(["col_a", "col_b"])`

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -970,6 +970,10 @@ class StreamingDataFrame(BaseStreaming):
         """
         Drop column(s) from the message value (value must support `del`, like a dict).
 
+        This operation occurs in-place, meaning reassignment is entirely OPTIONAL: the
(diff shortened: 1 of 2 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -915,6 +915,35 @@ def hopping_window(
             name=name,
         )
 
+    def drop(self, columns: Union[str, List[str]]) -> Self:
```
ghost, line 969: Hey, we can use variadic params instead of `Union` here:
```
def drop(self, *columns: str) -> Self:
```

ghost (changes requested):

quixstreams/dataframe/dataframe.py
tim-quix: Discussed this and we will leave as is =)

ghost (approved):

stereosky:
Absolutely perfect timing for this as I needed it tomorrow 🙌🏽

Issue #399, opened 2024-07-15
tree() method of the core Stream class needs to be optimized for time complexity


aparnadotk:
**Tell us about the bug**
The tree() method of the Stream class inserts nodes at the beginning of a list inside a while loop, causing a potential worst-case time complexity of O(n**2), since the list.insert(0, node) operation shifts all items of the list by one position each time it is called. 

Also, the function docstring needs to be corrected to reflect the implementation ordering of parent to child nodes (vs the child to parent currently specified).

**What did you expect to see?**
For large streams, this can cause a perf degradation.

**What version of the library are you using?**
Latest source on main branch.

**Workaround?**
The implementation can be fixed by first appending to the list as we traverse the ancestors of the current node, and finally do an in-place reverse on the list to obtain the expected ordering.

**Anything else we should know?**

ghost:
Closed by #400 👍

PR #400, opened 2024-07-15
Optimize tree() method of Stream class to reduce time complexity

aparnadotk:
Modified the `tree()` method to build the list in reverse order and then reverse it at the end, making the operation more efficient.

Earlier version did an insert(0, ...) operation on the list for each parent, which can have a worst-case time complexity of O(n**2) due to shifting of list elements for each call.

Modified implementation has an O(n) time complexity, with no additional space complexity, and maintains the same node ordering and behavior as the original tree() method.

The function docstring also needed a correction which has been incorporated into this fix.

Related Issue #399

### DIFF
--- a/quixstreams/core/stream/stream.py
+++ b/quixstreams/core/stream/stream.py
@@ -234,14 +234,19 @@ class Stream:
         """
         Return a list of all parent Streams including the node itself.
 
-        The tree is ordered from child to parent (current node comes first).
+        The tree is ordered from parent to child (current node comes last).
         :return: a list of `Stream` objects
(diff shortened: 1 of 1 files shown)


ghost (approved):
Hi Aparna! 
It's a nice catch 👍 .

About the optimization itself: 
Please note that the `tree()` method is always called before any processing happens, so the overall impact of this optimization on speed may be limited.
The docstring fix is correct ✅ 

Thanks for the contribution!

aparnadotk:
Thank you, Daniil!

PR #402, opened 2024-07-16
Protobuf SerDes support

ghost:


pyproject.toml
```diff
@@ -23,6 +23,15 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 
+[project.optional-dependencies]
+all = [
+  "protobuf"
```
ghost: Do all versions of protobuf work or should we pin the minimal one at least?

quixstreams/models/serializers/protobuf.py
```diff
@@ -0,0 +1,56 @@
...
+            raise SerializationError(str(exc)) from exc
+
+
+class ProtobufDeserializer(Deserializer):
+    def __init__(
+        self,
+        msg_type: Message,
+    ):
+        """
+        Deserializer that parses protobuf data.
```
ghost: Let's add that it parses data into dictionaries instead of objects here

tests/test_quixstreams/test_models/test_serializers.py
```diff
@@ -34,6 +40,10 @@ class TestSerializers:
             (BytesSerializer(), b"abc", b"abc"),
             (JSONSerializer(), {"a": 123}, b'{"a":123}'),
             (JSONSerializer(), [1, 2, 3], b"[1,2,3]"),
+            (ProtobufSerializer(Test), {}, b""),
+            (ProtobufSerializer(Test), {"id": 3}, b"\x10\x03"),
+            (ProtobufSerializer(Test), {"name": "foo", "id": 2}, b"\n\x03foo\x10\x02"),
```
ghost, line 86: Can the byte strings be replaced here with something more readable?
E.g., using something like `to_protobuf_bytes({"name": "foo", "id": 2})` or a similar approach.

Otherwise, it's unclear what the input of this test case is.

ghost (changes requested):
Hey @quentin-quix, it looks good overall, with only a few minor comments 👍 

**Question:** How well does it work with more complex Protobuf types like Enums? E.g., Enum can be deserialized as `enum.Enum` and as an int. 
Do we cover that?

Also, please remember to add the info about Protobuf to the docs (see `docs/advanced/serialization.md`).

pyproject.toml
ghost: AFAIK all current versions should work, they all implement the necessary API.

tests/test_quixstreams/test_models/test_serializers.py
ghost: I'm not sure what you mean here. The bytes are the result of the serializer, implementing a `to_protobuf_bytes` would basically be re-implementing protobuf.

ghost:
> **Question:** How well does it work with more complex Protobuf types like Enums? E.g., Enum can be deserialized as `enum.Enum` and as an int.
> Do we cover that?

There is an argument for it that we should expose, `use_integers_for_enums`.

gwaramadze (changes requested):
I think new serdes classes need to be imported and added to `SERIALIZERS` and `DESERIALIZERS` maps in `quixstreams.models.serializers.__init__`

ghost:
> I think new serdes classes need to be imported and added to `SERIALIZERS` and `DESERIALIZERS` maps in `quixstreams.models.serializers.__init__`

We don't want to do it for protobuf since it needs an extra import. Importing it in `quixstreams.models.serializers.__init__` would raise an `ImportError` due to the possibly missing dependencies. We could catch the error and skip it but iirc we decided against it

quixstreams/models/serializers/protobuf.py
```diff
@@ -0,0 +1,100 @@
...
+            if not self._to_dict:
+                return msg
+
+            return MessageToDict(
+                msg,
+                always_print_fields_with_no_presence=True,
+                use_integers_for_enums=self._use_integers_for_enums,
+                preserving_proto_field_name=self._preserving_proto_field_name,
+            )
+        except DecodeError as exc:
```
gwaramadze, line 89: I think that only the `ParseFromString` method raises `DecodeError`. The `if` and `return` statements should not be in the try block.

pyproject.toml
ghost: Let's pin the version to be safe.

quixstreams/models/serializers/protobuf.py
```diff
@@ -0,0 +1,100 @@
...
+            raise SerializationError(str(exc)) from exc
+
+
+class ProtobufDeserializer(Deserializer):
+    def __init__(
+        self,
+        msg_type: Message,
+        use_integers_for_enums: bool = False,
+        preserving_proto_field_name: bool = False,
+        to_dict=True,
```
ghost: The type hint is missing

ghost: The docs also must mention that in the current implementation, `to_dict` affects speed.

ghost (approved):

gwaramadze (approved):

Issue #410, opened 2024-07-19
Flaky test_group_by_with_window test


gwaramadze:
Sometimes `test_group_by_with_window` fails when the full suit is run.
```
>       assert processed_count == total_messages
E       assert 0 == 2
```
https://github.com/quixio/quix-streams/actions/runs/10004949491/job/27654670279

ghost:
Relates to #390

Issue #390, opened 2024-06-26
Test ordering causes failures in very specific circumstances.


tim-quix:
## The bug
In v2.7 `exactly-once` was introduced, and some tests were parameterized to confirm consistent behavior around `ALOS` and `EOS`.

In one of those tests, `TestAppGroupBy` - `test_group_by_with_window`...the test ordering between `ALOS` and `EOS` appears to matter and cause test failures, but only in specific circumstances.

The bug occurs specifically when:
1. All of `test_app` (or the entire suite) is run
2. `test_group_by_with_window` runs ALOS first, then EOS

This will cause the EOS test to fail.

The _actual_ reason is unknown, but more details will be provided below.

## Behavior around the bug

As mentioned, this bug appears ***very*** specific.

- If you add a sufficient sleep step (2-3s) before running the app, or add additional logging that slows app processing, EOS will pass consistently.
- Switching the order to EOS, ALOS _causes both to pass consistently_.
- Running only `TestAppGroupBy` (rather than all of `test_app`) _causes EOS to pass consistently_.
- If you add multiple parameter runs like [ALOS, EOS, ALOS, EOS,...], only the first EOS fails.
- I have not encountered this behavior on any other test.

### Test failure
message `processed_count` validation fails, where actual (1) != expected (2)

### Source of failure
Basically, the app cannot seem to successfully subscribe to (the beginning of?) the groupby repartition topic, and thus cannot read the group-by'd message.

This is pretty easily surmised from the warning thrown in the stack trace:

```
[WARNING] quixstreams.kafka.consumer: OFFSET [rdkafka#consumer-139] [thrd:main]: repartition__46fd98ea-e020-4177-b444-115ad5311adc--0975fdbb-7076-4e54-a985-671692120a93--user [0]: offset reset (at offset 0 (leader epoch 0), broker 0) to offset BEGINNING (leader epoch -1): fetch failed due to requested offset not available on the broker: Broker: Offset out of range
```

What's strange is: there are NO messages on this topic to begin with, and thus there should be no offset at all...so I'm not sure how the offset is "out of range". The first (and only!) message gets put on the topic from the app itself during processing.

There's nothing else useful in the stack trace...the app otherwise appears to run normally, and basically fails due to the app hitting the app runtime timeout of `10s` we impose on it (which honestly is perhaps the most concerning part of this).

## What I've tried

Basically the stuff listed above.

[...]

PR #412, opened 2024-07-19
fix bad timestamp values in groupby tests

tim-quix:
Two `Application` tests were failing due to a "bad" timestamp value of `1000`.

Because this timestamp was being produced as the actual message timestamp to Kafka, it would once in a while get cleaned up due to retention/segment cleanup (`1000` vs `now` diff is well beyond the topic retention time), causing these tests to fail sometimes as the message they needed to consume as part of the test were removed mid-test.

### DIFF
--- a/tests/test_quixstreams/test_app.py
+++ b/tests/test_quixstreams/test_app.py
@@ -557,7 +557,7 @@ class TestAppGroupBy:
 
         processed_count = 0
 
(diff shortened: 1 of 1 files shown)


tim-quix:
This closes #410 and #390 .

gwaramadze (approved):
👏

PR #417, opened 2024-07-22
new Sinks API

### LINKED ISSUE #438: New Sinks API
Implement an API to sink data to external sources like databases, object stores, MQTT, and other.

The implementation details will be included in the individual PRs:
- #417 
- #439

ghost:
Added a new Sinks API.

Sinks can be of two types:
- Batching sinks -  batching data before writing to the external destination (e.g., databases, files)
- Streaming sinks - writing every message as it gets processed (e.g. MQTT, RabbitMQ)

**The Streaming sinks are not included in this PR.**

## API
Users need to create an instance of a sink and pass it to `StreamingDataFrame.sink()`.

**`.sink()` is a terminal operation to make sure the data in the batches is not mutated after it's sinked.**
New operations cannot be added to the same SDF after the `.sink()` is called.

Example:

```python
from quixstreams.sink.csv import CSVSink


app = Application(
    auto_offset_reset="earliest",
    consumer_group=str(uuid.uuid4()),
    loglevel="DEBUG",
    commit_every=1000,  # Commit every 1000 offsets
)
topic = app.topic("topic")

sink = CSVSink(path="sink.csv")

sdf = app.dataframe(topic=topic)

sdf.sink(sink)

app.run(sdf)
```

## Changes in this PR

- Added an ABC for all sinks - `BaseSink`.

- Added an ABC for batching sinks - `BatchingSink`.  
Users can subclass it and implement custom sinks.

- Added two implementations of batching sinks - `InfluxDBV3Sink` and `CSVSink`

## How sinks work
- For each processed message, the SDF calls `Sink.add()`, passing all message metadata. Sinks can either add a message to the batch or send it to the destination.

- On checkpoint, the app calls `Sink.flush(topic, partition)` for each topic partition with a processed offset.  
At this point, sinks can write the batches outside or acknowledge already-produced messages.
Batching sinks add a message to in-memory batches flushed on each checkpoint.

- Sinks can implement a behavior to handle backpressure when an external destination cannot accept writes due to rate limits.  
When this happens, a sink can raise the `SinkBackpressureError(retry_after, topic, partition)` error with a timeout.
The app will drop the processed batch and pause consuming of this topic partition for the timeout defined by `retry_after`.  
After that, the app will resume consuming and process data again.


quixstreams/sinks/csv.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def write(self, batch: SinkBatch):
+        is_new = not os.path.exists(self.path)
+        fieldnames = (
+            "__key",
+            "__timestamp",
+            "__value",
+            "__topic",
+            "__partition",
+            "__offset",
```
ghost: Why the leading `__` ?

quixstreams/sinks/csv.py
```diff
@@ -0,0 +1,44 @@
...
+            writer = csv.DictWriter(f, dialect=self.dialect, fieldnames=fieldnames)
+            if is_new:
+                writer.writeheader()
+
+            for item in batch:
+                writer.writerow(
+                    {
+                        "__key": json.dumps(item.key.decode()),
+                        "__timestamp": json.dumps(item.timestamp),
+                        "__value": json.dumps(item.value),
```
ghost: I'm not convinced by the `json.dumps` here, we should support serializers.

quixstreams/sinks/csv.py
```diff
@@ -0,0 +1,44 @@
...
+        with open(self.path, "a") as f:
+            writer = csv.DictWriter(f, dialect=self.dialect, fieldnames=fieldnames)
+            if is_new:
+                writer.writeheader()
+
+            for item in batch:
+                writer.writerow(
+                    {
+                        "__key": json.dumps(item.key.decode()),
+                        "__timestamp": json.dumps(item.timestamp),
```
ghost: Is this useful ? afaik it's the same as `str(item.timestamp)`.

ghost:
> The offsets won't be committed to Kafka, and the app will reprocess the same messages again.

I understand why we don't commit but can we skip reprocessing the messages ? The sink should be able to keep the failed batches in memory and retry once the remote is ready to accept data.

ghost:
> > The offsets won't be committed to Kafka, and the app will reprocess the same messages again.
> 
> I understand why we don't commit but can we skip reprocessing the messages ? The sink should be able to keep the failed batches in memory and retry once the remote is ready to accept data.

I initially did it this way, but the problem is that exactly-once transactions have strict timeouts.  
If you keep Sink data in memory, you may end up in a situation when you cannot commit the transaction anymore because the sink was backpressured for too long.

quixstreams/sinks/csv.py
ghost: The CSV sink is just an example here.
It's meant to showcase how it works, and I didn't put too much thought into it at this point :) .

The leading `__` is to highlight that these are auto-generated fields that have nothing to do with the actual data structure.
In the production case, this sink needs to support custom formatting and serialization.

tomas-quix:
I don't see timeout anywhere. That is must-have feature as again, you will get last buffer in the memory forever

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,90 @@
...
+                    else value
+                )
+                ts = value[time_key] if time_key is not None else item.timestamp
+                record = {
+                    "measurement": measurement,
+                    "tags": tags,
+                    "fields": fields,
+                    "time": ts,
+                }
+                records.append(record)
```
tomas-quix, line 194: In TRACE mode I would expect line protocol messages in log.

ghost: Hey @tomas-quix , you can pass `debug=True` to `InfluxDBV3Sink` separately to achieve that 👍 

I didn't want to blow up the logs when the loglevel is DEBUG.

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,90 @@
...
+            try:
+                self._client.write(records, write_precision=self._write_precision)
+                # TODO: Maybe use backoff and retries here
+            except influxdb_client_3.InfluxDBError as exc:
+                if exc.retry_after:
+                    # The write limit is exceeded, raise a SinkBackpressureError
+                    # to pause the partition for a certain period of time.
+                    raise SinkBackpressureError(
```
tomas-quix, line 215: I am not so sure this is enough. This error might contains errors not related to backpressure stuff.

ghost: Yep, it's a bug, the fix is there already 👍

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,91 @@
...
+        fields_keys = self._fields_keys
+        tags_keys = self._tags_keys
+        time_key = self._time_key
+        for write_batch in self._iter_batches(batch, n=self._batch_size):
+            records = []
+            for item in write_batch:
+                value = item.value
+                tags = {tag_key: value[tag_key] for tag_key in tags_keys}
+                tags["__key"] = item.key
+                tags["__offset"] = item.offset
```
tomas-quix: This is definitely not a tag. It will very quickly cause high cardinality. Also although other tags looks handy I would consider if this is turned off as default. Not strong on that though...

ghost: Okay, that makes sense. I probably misinterpreted the "tag" in the context of InfluxDB.

quixstreams/sinks/csv.py
ghost: Added basic serialization for CSV

quixstreams/sinks/csv.py
ghost: Removed dunders

quixstreams/sinks/csv.py
ghost: Yep, makes sense

quixstreams/sinks/influxdb_v3.py
ghost: Updated

quixstreams/app.py
```diff
@@ -920,23 +927,28 @@ def _on_revoke(self, _, topic_partitions: List[TopicPartition]):
...
         Dropping lost partitions from consumer and state
         """
+        logger.debug(f"Rebalancing: dropping lost partitions")
         if self._state_manager.stores:
-            logger.debug(f"Rebalancing: dropping lost state store partitions")
             for tp in topic_partitions:
                 self._state_manager.on_partition_revoke(
                     topic=tp.topic, partition=tp.partition
                 )
+                self._processing_context.on_partition_revoke(
```
ghost: This won't trigger if `self._state_manager.stores` is falsy

quixstreams/checkpointing/checkpoint.py
```diff
@@ -155,17 +162,56 @@ def commit(self):
...
+        for (topic, partition), offset in self._tp_offsets.items():
+            for sink in sinks:
+                if self._pausing_manager.is_paused(topic=topic, partition=partition):
+                    # The topic-partition is paused, skip flushing other sinks for
+                    # this TP.
+                    # Also drop the batch to reprocess it later.
+                    # Note: when flushing multiple sinks for the same TP, some
+                    # of them can be flushed before one of the sinks is backpressured.
+                    sink.on_paused(topic=topic, partition=partition)
+                    break
```
ghost: Should this be `continue` ? I think we want to trigger the `on_paused` action of all sinks ?

quixstreams/sinks/base/sink.py
```diff
@@ -0,0 +1,80 @@
...
+        if batch is None:
+            batch = SinkBatch(topic=topic, partition=partition)
+            self._batches[tp] = batch
+        batch.append(
+            value=value, key=key, timestamp=timestamp, headers=headers, offset=offset
+        )
+
+    def flush(self, topic: str, partition: int):
+        batch = self._batches.get((topic, partition))
+        if batch is not None:
```
ghost: Small nitpick but this is more readable imo, less indentation level further on in the method.
```
if batch is None:
  return
```

quixstreams/sinks/csv.py
```diff
@@ -0,0 +1,68 @@
...
+        fieldnames = (
+            "key",
+            "value",
+            "timestamp",
+            "topic",
+            "partition",
+            "offset",
+        )
+        with open(self.path, "a") as f:
+            writer = csv.DictWriter(f, dialect=self.dialect, fieldnames=fieldnames)
```
ghost, line 51: I think it makes sense to do all that in `__init__` and keep the open file around and call `f.flush()` for each batch to flush the buffers.

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Optional, Iterable, TypeVar
+
+try:
+    import influxdb_client_3
+    from influxdb_client_3 import InfluxDBClient3, WritePrecision, WriteOptions
+    from influxdb_client_3.write_client.client.write_api import WriteType
+
+    _influx_client_installed = True
+except ImportError:
+    _influx_client_installed = False
```
ghost: I don't think we should do that. We decided against it with protobuf and avro serdes. It should raise at import time if dependencies are missing and we should take care this isn't imported anywhere it's not necessary.

quixstreams/app.py
daniil-quix: Yep, I know. 
I moved these debug logs to the top in the other handlers too.

quixstreams/checkpointing/checkpoint.py
ghost: That's right, nice catch 👍 .

ghost: Fixed and added a test for that

quixstreams/sinks/csv.py
ghost: TBH, I don't think it matters much here. 
The CSV sink is meant for local debugging, so it's just easier to flush and close the file immediately after writing.

quixstreams/app.py
ghost: Yep, forgot to update `on_lost` too

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+        self._measurement = measurement
+        self._fields_keys = fields_keys
+        self._tags_keys = tags_keys or []
+        self._include_metadata_tags = include_metadata_tags
+        self._time_key = time_key
+        self._write_precision = time_precision
+        self._batch_size = batch_size
+
+    def _iter_batches(self, it: Iterable[T], n: int) -> Iterable[Iterable[T]]:
```
tim-quix: Seems to me that something like this should be something the `SinkBatch` knows how to do for you (generating batches/chunking of certain size)...I imagine this is gonna be a common pattern.

tim-quix: On a related note (and I fully admit this is more of a subjective comment) if we also renamed `SinkBatch` to `SinkBatcher`, it helps make clear that it both collects messages _and_ generates batches for you as needed. As to me, calling it just `SinkBatch` implied to me it was the entire batch, when it's not really (which will also be true in other similar implementations as mentioned in my previous comment).

Admittedly a small thing, but I thought between the two ideas it might make the collective interaction/behavior more clear/streamlined.

ghost: > Seems to me that something like this should be something the SinkBatch knows how to do for you (generating batches/chunking of certain size)...I imagine this is gonna be a common pattern.

Agree on this one, it can simplify the Influxdb sink overall

ghost: > As to me, calling it just SinkBatch implied to me it was the entire batch

It is an entire batch for the given topic-partition. You can iterate over it and get all the batched messages.
So I'd keep the name the same.

ghost: The `_iter_batches()` is just a utility method to limit the size of a single write to the destination.
E.g. if you need to write max 1000 items at once

tim-quix (approved):

quixstreams/sinks/influxdb_v3.py
```diff
@@ -0,0 +1,140 @@
...
+                ts = value[time_key] if time_key is not None else item.timestamp
+                record = {
+                    "measurement": measurement,
+                    "tags": tags,
+                    "fields": fields,
+                    "time": ts,
+                }
+                records.append(record)
+            try:
+                self._client.write(
```
tomas-quix, line 200: I am missing any sort of logging here. It looks like sink is not working because of that.

PR #419, opened 2024-07-24
Feature: Branching

tim-quix:
## What

This PR includes two different but related operations (that can be used together): "branching" and using more than 1 sdf (consuming multiple topics).

See documentation added in this PR for more details.


quixstreams/core/stream/stream.py
```diff
@@ -202,53 +206,194 @@ def add_transform(
         return self._add(TransformFunction(func, expand=expand))
 
     def diff(
-        self,
-        other: "Stream",
+        self, other: "Stream", prune: bool = False, enforce_direct_split: bool = False
```
ghost: Do we need `prune` as a parameter here?
From the usages, it seems it's always `True` in the `StreamingDataFrame.__getitem__` and `__setitem__` methods.
In this case, we may always prune the diffs until we need to do otherwise.

quixstreams/core/stream/stream.py
```diff
@@ -202,53 +206,194 @@ def add_transform(
...
+        while (parent := node.parent) and (allow_splits or len(parent.children) < 2):
+            tree_.append(parent)
             node = node.parent
 
         # Reverse to get expected ordering.
         tree_.reverse()
 
         return tree_
 
+    def prune(self, other: Self):
```
ghost: If we always prune, this method can become `_prune`.

quixstreams/core/stream/stream.py
```diff
@@ -202,53 +206,194 @@ def add_transform(
...
+        :param other: another Stream
+        :return:
+        """
+        node = self
+        while node:
+            if other in node.children:
+                node.children.remove(other)
+                return
+            node = node.parent
+        raise InvalidOperation("Cannot filter with the same SDF more than once.")
```
ghost: Are there any cases except filtering when this exception may happen?
The method is called `prune`, but the exception message specifically mentions "filtering"

quixstreams/core/stream/stream.py
```diff
@@ -202,53 +206,194 @@ def add_transform(
...
+
+    def tree_all_nodes(
```
ghost: I want to suggest another names here too:

- `tree_all_nodes` -> `full_tree`
- `tree_root_path` -> `root_path`

These two now have different meaning, and with this new feature only "full_tree" is actually a tree, while "root_path" is only a section in the tree.

quixstreams/core/stream/stream.py
```diff
@@ -202,53 +206,194 @@ def add_transform(
...
+        :return: The collection of all Streams interconnected to this one
+        """
+        if not collected_nodes:
+            collected_nodes = []
+        if not current_node:
+            current_node = self.tree_root_path()[0]
+        collected_nodes.append(current_node)
+        for child in current_node.children:
+            self.tree_all_nodes(collected_nodes, child)
+        return collected_nodes
```
ghost: What do you think if we move the recursive part into a closure here?
I.e., define some nested `_collect(nodes, current)` function and call it recursively instead of `self.tree_all_nodes` itself.

It will make it more readable, because `collected_nodes` and `current_node` variables can be defined only in the method itself, while `_collect` will do the tracing part

quixstreams/dataframe/registry.py
```diff
@@ -0,0 +1,59 @@
...
+class DataframeRegistry:
+    def __init__(self):
+        self._registry: Dict[str, Stream] = {}
+        self._topics: List[Topic] = []
+        self._branch_origins: Dict[str, str] = {}
+
+    @property
+    def registry(self) -> Dict[str, Stream]:
+        return self._registry
+
```
ghost: Can you please add docstrings for the class and its methods?

quixstreams/dataframe/dataframe.py
```diff
@@ -134,14 +134,14 @@ def __init__(
         self,
         topic: Topic,
         topic_manager: TopicManager,
+        stream_registry: DataframeRegistry,
```
ghost: The naming here is not exactly clear: the class is named `DataframeRegistry` registry, while the parameter name is `stream_registry`. 
I think it should be either one or the other :)

quixstreams/dataframe/registry.py
```diff
@@ -0,0 +1,59 @@
...
+class DataframeRegistry:
+    def __init__(self):
+        self._registry: Dict[str, Stream] = {}
+        self._topics: List[Topic] = []
+        self._branch_origins: Dict[str, str] = {}
+
+    @property
+    def registry(self) -> Dict[str, Stream]:
+        return self._registry
```
ghost: This property is not used anywhere, it probably should be removed

quixstreams/dataframe/exceptions.py
```diff
@@ -23,3 +23,6 @@ class GroupByLimitExceeded(QuixException): ...
 
 
 class DataFrameLocked(QuixException): ...
+
+
+class DuplicateStreamingDataFrame(QuixException): ...
```
ghost: The exception class should be listed in `__all__` if it's importable outside

quixstreams/dataframe/registry.py
```diff
@@ -0,0 +1,59 @@
...
+
+if TYPE_CHECKING:
+    from .dataframe import StreamingDataFrame
+
+
+class DataframeRegistry:
+    def __init__(self):
+        self._registry: Dict[str, Stream] = {}
+        self._topics: List[Topic] = []
+        self._branch_origins: Dict[str, str] = {}
```
ghost: This variable could use a better name.
Looks like it stores a mapping of <repartition_topic>: <source_topic>, but it's not clear from the name

quixstreams/dataframe/registry.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def _check_branch_count(self, current_branch: str):
+        branch_count = 0
+        while origin := self._branch_origins.get(current_branch):
+            branch_count += 1
+            current_branch = origin
+        if branch_count >= 1:
+            raise GroupByLimitExceeded(
+                "Only one GroupBy operation is allowed per StreamingDataFrame"
+            )
```
ghost: Can it be done in a more simple way?
1. Looks like it recursively search for the topics in a map, is it necessary to do so?
2. Given the `if branch_count >= 1:` block, the code can probably exit early in the loop

ghost (changes requested):

quixstreams/dataframe/registry.py
tim-quix: This seemed like the simplest approach for allowing multiple group-bys from the same topic (i.e. during a branch), but not allowing depth > 1. Could maybe do it via an object with parent attribute, but seemed overkill when I'm not doing anything else with it.

I imagine this approach could change as joins are introduced as well, so I tried not to overthink it. Open to ideas though =)

ghost (approved):

PR #420, opened 2024-07-26
Feature: Sources API

ghost:
New sources API

Sources are implemented as a second process that fetch data from somewhere (http, websocket, MQTT, kafka, ...) and send it to a kafka topic.

**Example**

Users must create an instance of a source and pass it to the `app.dataframe` method. A source is the start of a dataframe. You can't stream data from both a topic and a source.

```
    from quixstreams.sources import ValueIterableSource

    app = Application(broker_address='localhost:9092', consumer_group='group')
    source = ValueIterableSource(name="my_source", values=range(10))

    sdf = app.dataframe(source=source)
    sdf.print()

    app.run(sdf)
```


tim-quix:
Nice work Quentin, good foundation so far I'd say!

My review comments are going to be in line with the following understanding of how it works (also in the future, try to provide a high level overview like this as it'll be helpful for jumping into a code review, but no big deal!)

- User imports desired `Source` with whatever settings needed
- User inits a `SDF` with given `Source` (I see you mostly set up support for >1 `SDF` with `SourceManager`)
- User runs app with given `SDF` which inits the following under the hood:
	- register produce topic (uses `Source.default_topic()` if not provided)
	- configure/finalize setup of `Source` (hands each its producer and produce topic)
	- sets up threading of `Sources` (`SourceManager` registers each `Source` as a `SourceThread`)
	- starts all threads

- once running:
	- `Source`s produce messages
	- eventually the consumer (reading from produce topic) hits a checkpoint, where it flushes `Source` producer at the end of the checkpoint

- stopping:
	- `SourceManager` gracefully handles stopping `SourceThread`s

quixstreams/app.py
```diff
@@ -4,6 +4,7 @@
 import os
 import signal
 import warnings
+import threading
```
tim-quix: Question WRT threading (I ask this with lack of experience besides some googling):

Is it definitive that threading is the best approach here vs multiprocessing? I'm guessing we're using threading due to shared resources like the context manager and producer flushing with checkpointing (though I'm sure with the right setup we could use multiprocessing)? 

Wasn't sure if threading would hit limits quicker, as at some point I imagine we will no longer necc be I/O bound if we have enough sources running at once (maybe even just one?). I feel like multiple factors might be at play there though (including hardware related, not sure how fractional CPU in things like Kubernetes affects threading and multiprocessing)

quixstreams/app.py
```diff
@@ -657,7 +686,9 @@ def get_producer(self) -> Producer:
             extra_config=self._producer_extra_config,
         )
 
-    def get_consumer(self, auto_commit_enable: bool = True) -> Consumer:
+    def get_consumer(
```
tim-quix: Just curious as to why this was changed?

quixstreams/app.py
```diff
@@ -710,6 +741,31 @@ def clear_state(self):
...
+            topic = source.default_topic(self._topic_manager)
+
+        producer = RowProducer(
+            broker_address=self._broker_address,
+            extra_config=self._producer_extra_config,
+            flush_timeout=self._consumer_extra_config.get(
+                "max.poll.interval.ms", _default_max_poll_interval_ms
+            )
+            / 1000,  # convert to seconds
+            transactional=self._uses_exactly_once,
```
tim-quix: In this case, I think we want to use ALO regardless (or at the very least, no use the app setting), as you can't really promise EOS with Sourcing except for maybe some special circumstances, no?

Also just better to be safe as there are other configurations in place when this is set to EOS.

quixstreams/app.py
```diff
@@ -739,6 +795,9 @@ def run(
 
         :param dataframe: instance of `StreamingDataFrame`
         """
+        self._run(dataframe)
```
tim-quix, line 793: Did you have a reason for splitting this into another function? =)

quixstreams/app.py
tim-quix: To that end, might even want to warn users somehow and not give them a false impression of the processing guarantees provided with a source.

quixstreams/models/topics/manager.py
```diff
@@ -241,9 +241,9 @@ def topic_config(
     def topic(
         self,
         name: str,
-        value_deserializer: Optional[DeserializerType] = None,
+        value_deserializer: Optional[DeserializerType] = "json",
```
tim-quix: I think I'm generally on board with this default change since it's what the app topic does anyway, but trying to remember if there was a reason to have not set it up that way.

quixstreams/app.py
```diff
@@ -627,6 +642,20 @@ def stop(self, fail: bool = False):
         if self._state_manager.using_changelogs:
             self._state_manager.stop_recovery()
 
+    def _setup_excepthook(self):
+        """
+        Override threading excepthook to handle exceptions hapenning in sources.
```
```suggestion
        Override threading excepthook to handle exceptions happening in sources.
```

quixstreams/app.py
tim-quix: also extra "m" in  `consumer_group` =)

quixstreams/sources/base.py
```diff
@@ -0,0 +1,226 @@
...
+        * checkpoint
+        * stop
+        * default_topic
+
+    Helper methods
+        * serialize
+        * produce
+        * sleep
+
+    A `stopping` :class:`threading.Event` to handle gracefull shutdown and a lock to handle concurrent `produce` and `checkpoint` calls
```
tim-quix:
```suggestion
    A `stopping` :class:`threading.Event` to handle graceful shutdown and a lock to handle concurrent `produce` and `checkpoint` calls
```

tim-quix (commented):
In addition to the comments here, a couple things:

- I think some stuff could use additional doc strings to better explain what is happening overall with the flow (will help all of us in the future) =) I wasn't gonna go too deep on it this time since you mentioned you still had more to do in this regard
- It would be good to see what a pattern is going to look like where you can actually manage the a source position/commit structure...might shape things quite a bit more here. Overall it's a good foundation so far and makes sense to me.
- I made a PR #443 with some suggestions as it was simply easier to do that than leave you tons of comments to parse out and figure out what i meant =) feel free to take from it what you please!

quixstreams/models/topics/manager.py
ghost: I couldn't figure out why. In any case it seems this is only called from the `Application.topic` method that uses this default.

quixstreams/app.py
ghost: Not really. It's mostly about running sources on their own but as we discussed this clash a bit with your ongoing work.

The change makes it that `app.run` needs a sdf and if you only want to run sources you can call `app._run`. Mostly a POC and I expect this will change again with your work.

quixstreams/app.py
ghost: Yes I think we can start by having it set to ALO and maybe later some sources can leverage EOS

quixstreams/app.py
ghost: This can be reverted. Some left over of a test I forgot to remove :facepalm:

quixstreams/app.py
ghost: We could leverage multiprocessing for sources. It probably even make more sense than threading due to python constraints. We actually don't share too much data from the main thread to the source thread, especially since we are using another producer so it's be worth looking into it. 

It shouldn't be too hard to go from one to the other. Mostly, as you mentioned, handling the checkpointing.

ghost: swapped to multiprocessing

quixstreams/sources/base.py
```diff
@@ -0,0 +1,375 @@
...
+    key_serializer: Optional[SerializerType] = BytesSerializer()
+    timestamp_extractor: Optional[TimestampExtractor] = None
+
+    def asdict(self):
+        return {
+            field.name: getattr(self, field.name) for field in dataclasses.fields(self)
+        }
+
+
+class BaseSource(ABC):
```
gwaramadze, line 22: Why can't the `BaseSource` and `Source` be merged into a single ABC (called `Source`)?

Are some sources supposed to subclass from lower-level `BaseSource` and some from higher-level `Source`? If yes then what would be the use cases for this distinction?

I think for simplicity we should create and promote a single base class. Of course if possible, I may not be aware of some details.

quixstreams/dataframe/dataframe.py
```diff
@@ -48,6 +48,8 @@
 from quixstreams.processing import ProcessingContext
 from quixstreams.sinks import BaseSink
 from quixstreams.state.types import State
+from quixstreams.sources import BaseSource
+from quixstreams.rowproducer import RowProducer
```
gwaramadze: Minor nitpick: let's keep this import section sorted.

quixstreams/app.py
```diff
@@ -597,8 +602,14 @@ def dataframe(
             to be used as an input topic.
         :return: `StreamingDataFrame` object
         """
+        if not source and not topic:
+            raise TypeError("one of `source` or `topic` is required")
```
gwaramadze: I believe for this pattern, a `ValueError` is more suited.

quixstreams/app.py
```diff
@@ -710,6 +721,32 @@ def clear_state(self):
...
+        if not topic:
+            topic = self._topic_manager.source_topic(source)
+
+        producer = RowProducer(
+            broker_address=self._broker_address,
+            extra_config=self._producer_extra_config,
+            flush_timeout=self._consumer_extra_config.get(
+                "max.poll.interval.ms", _default_max_poll_interval_ms
+            )
+            / 1000,  # convert to seconds
```
gwaramadze: Readability suffers here a little bit. I suggest breaking it down into explicit stages:

```python
max_poll_interval_ms = self._consumer_extra_config.get(
    "max.poll.interval.ms", _default_max_poll_interval_ms
)
flush_timeout_seconds = max_poll_interval_ms / 1000
producer = RowProducer(
    ...,
    flush_timeout=flush_timeout_seconds,
)

File: quixstreams/app.py [hunk 12]
```diff
@@ -710,6 +721,32 @@ def clear_state(self):
         """
         self._state_manager.clear_stores()
 
+    def source(self, source: BaseSource, topic: Optional[Topic] = None) -> Topic:
```
gwaramadze: I think a more suitable name for this method is `add_source`

quixstreams/app.py
```diff
@@ -739,6 +776,17 @@ def run(
 
         :param dataframe: instance of `StreamingDataFrame`
         """
+        self._run(dataframe)
+
+    def _exception_handler(self, exc_type, exc_val, exc_tb):
+        fail = False
+
+        if exc_val is not None and exc_type is not SourceException:
```
gwaramadze, line 800: Could you explain this logic? Why is `SourceException` treated differently?

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,244 @@
...
+            process.stop()
+
+    def raise_for_error(self) -> None:
+        """
+        Raise an exception if any process has stopped with an exception
+        """
+        for process in self.processes:
+            process.raise_for_error()
+
+    def alives(self) -> bool:
```
gwaramadze: Rename to `is_alive`?

quixstreams/dataframe/dataframe.py
ghost: Thanks. Yeah I'd like to introduce a proper linter for that. Like isort or ruff

quixstreams/app.py
ghost: `SourceException` is raised in the main thread when a source fails with an exception.

In that case it's safe to checkpoint the sdf before shutting down as source and sdf are not directly related, there is a topic in between.

quixstreams/sources/base.py
ghost: iirc Daniil preference if for one BaseClass that defines the strict minimum implementation. Then we can have a bunch of additional classes that implement additional features.

gwaramadze: Let's try to find an answer to my question: Are some sources supposed to subclass from lower-level `BaseSource` and some from higher-level `Source`?

Because if the answer is no, and all the final Sources are supposed to derive from Source then I really see no point in having two base classes.

If the answer is yes, then the name `Source` is inappropriate (too generic).

I will tag @daniil-quix here too.

quixstreams/sources/base.py
```diff
@@ -0,0 +1,375 @@
...
+        :param checkpoint_interval: Time in second between each checkpoints
+        """
+        super().__init__(name, shutdown_timeout)
+
+        self._checkpoint_interval = checkpoint_interval
+        self._checkpoint_error: Optional[BaseException] = None
+
+        self.__checkpoint_timer = self.__make_timer()
+
+    def __make_timer(self):
```
gwaramadze: Why some "private" method names start with double underscores?

quixstreams/dataframe/dataframe.py
gwaramadze: Yes, we need that. For now I use isort separately here and there. We should plug it via pre-commit but it is worthy of a separate PR.

quixstreams/sources/base.py
ghost: > Let's try to find an answer to my question: Are some sources supposed to subclass from lower-level BaseSource and some from higher-level Source?

Sorry, yes `BaseSource` is all you need to implement any source. `Source` is there as an intermediary step to make it easier.

 I'd expect most sources to subclass `Source` and not `BaseSource`, that's why this name is more generic

gwaramadze: I don't support this approach. My argument against this is that the user will be forced to put extra thought into deciding which class should be the base for his/her new source. That's why I'm asking for examples of sources that would subclass from `Source` and those that would subclass from `BaseSource` directly.

Of course, this is not a blocking disagreement 😉

quixstreams/models/topics/manager.py
```diff
@@ -12,6 +12,10 @@
 )
 from .topic import Topic, TopicConfig, TimestampExtractor
 
+if TYPE_CHECKING:
+    from quixstreams.sources.base import BaseSource
+
+
```
ghost: This import is not used anywhere and may be removed

quixstreams/models/topics/manager.py
```diff
@@ -283,6 +287,10 @@ def topic(
         self._topics[name] = topic
         return topic
 
+    def register(self, topic: Topic):
+        topic.name = self._resolve_topic_name(topic.name)
+        self._topics[topic.name] = topic
```
ghost: Please add a docstring here explaining when this method is called

quixstreams/dataframe/dataframe.py
```diff
@@ -137,6 +138,7 @@ def __init__(
         processing_context: ProcessingContext,
         stream: Optional[Stream] = None,
         branches: Optional[Dict[str, Stream]] = None,
+        source: Optional[BaseSource] = None,
```
ghost: Can we make the SDF class unaware of the sources? 
It would reduce the number of dependencies and simplify the code overall

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        if super().is_alive():
+            return
+
+        if not self._exceptions:
+            while self._rpipe.poll():
+                try:
+                    self._exceptions.append(self._rpipe.recv())
+                except EOFError:
+                    return
```
ghost, line 121: The "return" lines return None instead of bool here

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        This method flush the producer.
+
+        It ensure all message are successfully delived to kafka
+
+        :param float timeout: time to attempt flushing (seconds).
+            None use producer default or -1 is infinite. Default: None
+
+        :raises CheckpointProducerTimeout: if any message fails to produce before the timeout
+        """
+        logger.debug("Flushing source")
```
ghost, line 285: Let's add the source repr* here so it's easier to identify it in logs

quixstreams/sources/base.py
ghost: I think it's okay to have the `BaseSource` as a minimal interface definition and the `Source` class, which provides some basic implementation.
We need to mention in the `BaseSource`'s docstring and docs that in most cases users need to subclass `Source` to implement their own.

This way, it's clear which base class to inherit from.

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        ```
+        """
+
+    @abstractmethod
+    def default_topic(self) -> Topic:
+        """
+        This method is triggered when the user hasn't specified a topic for the source.
+
+        The source should return a default topic configuration
+        """
```
ghost:
```suggestion
        This method is triggered when the topic is not provided to the source.

        The source must return a default topic configuration.
        """
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        You can use it to perform any kind of shutting down cleanup. For example
+        flushing the producer.
+        """
+
+    @abstractmethod
+    def stop(self) -> None:
+        """
+        This method is triggered when the application is shutting down.
+
+        The source should make sure it's `run` method completes soon.
```
ghost:
```suggestion
        The source must ensure that the `run` method is completed soon.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        The subprocess will run as long as the run method execute. You should use it to fetch data and produce it to kafka.
+        """
+
+    @abstractmethod
+    def cleanup(self, failed: bool) -> None:
+        """
+        This method is triggered once the `run` method completes.
+
+        You can use it to perform any kind of shutting down cleanup. For example
+        flushing the producer.
```
ghost:
```suggestion
        Use it to clean up the resources and shut down the source gracefully. 
        For example, flushing the producer.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+    @property
+    def producer_topic(self):
+        return self._producer_topic
+
+    @abstractmethod
+    def run(self) -> None:
+        """
+        This method is triggered in the subprocess when the source is started.
+
+        The subprocess will run as long as the run method execute. You should use it to fetch data and produce it to kafka.
```
ghost:
```suggestion
        The subprocess will run as long as the run method executes.  
        Use it to fetch data and produce it to Kafka.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+    def __init__(self):
+        self._producer: Optional[RowProducer] = None
+        self._producer_topic: Optional[Topic] = None
+        self._configured: bool = False
+
+    def configure(self, topic: Topic, producer: RowProducer) -> None:
+        """
+        This method is triggered when the source is registered to the Application.
+
+        It's used to configure the source with the kafka producer and the topic it should use
```
ghost:
```suggestion
        It configures the source's Kafka producer and the topic it will produce to.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+    """
+    BaseSource class implementation providing
+
+    Implementation for the abstract method:
+        * `default_topic`
+
+    Helper methods
+        * serialize
+        * produce
+        * flush
```
ghost: If we expect users to subclass from the `Source` class, we must also add some basic examples here.

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def serialize(
+        self,
+        key: Optional[object] = None,
+        value: Optional[object] = None,
+        headers: Optional[Headers] = None,
+        timestamp_ms: Optional[int] = None,
+    ) -> KafkaMessage:
+        """
+        Serialize data into a :class:`quixstreams.models.messages.KafkaMessage` using the producer topic serializers.
```
ghost:
```suggestion
        Serialize data to bytes using the producer topic serializers and return a :class:`quixstreams.models.messages.KafkaMessage` .
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        value: Optional[Union[str, bytes]] = None,
+        key: Optional[Union[str, bytes]] = None,
+        headers: Optional[Headers] = None,
+        partition: Optional[int] = None,
+        timestamp: Optional[int] = None,
+        poll_timeout: float = 5.0,
+        buffer_error_max_tries: int = 3,
+    ) -> None:
+        """
+        Produce data to kafka using the source topic.
```
ghost:
```suggestion
        Produce a message to the configured source topic in Kafka.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+            timestamp=timestamp,
+            poll_timeout=poll_timeout,
+            buffer_error_max_tries=buffer_error_max_tries,
+        )
+
+    def flush(self, timeout: Optional[float] = None) -> None:
+        """
+        This method flush the producer.
+
+        It ensure all message are successfully delived to kafka
```
ghost:
```suggestion
        It ensures all messages are successfully delivered to Kafka.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        logger.debug("Flushing source")
+        unproduced_msg_count = self._producer.flush(timeout)
+        if unproduced_msg_count > 0:
+            raise CheckpointProducerTimeout(
+                f"'{unproduced_msg_count}' messages failed to be produced before the producer flush timeout"
+            )
+
+    def default_topic(self) -> Topic:
+        """
+        Return a topic matching the source name.
```
ghost:
```suggestion
        Return a default topic matching the source name.  
        The default topic will not be used if the topic has already been provided to the source.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+            value_serializer="json",
+        )
+
+    def __repr__(self):
+        return self.name
+
+
+class PollingSource(Source):
+    """
+    Source implementation for polling sources
```
ghost:
```suggestion
    An implementation for Sources that actively poll data (e.g., a source polling some HTTP API).
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+
+    Periodically call the `poll` method to get a new message.
```
ghost:
```suggestion
    It periodically calls the `poll` method to get a new message.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        polling_delay: float = 1,
+        shutdown_timeout: float = 10,
+    ) -> None:
+        """
+        :param name: The source unique name. Used to generate the topic configurtion
```
ghost:
```suggestion
        :param name: A unique name for the source. It is used to generate the topic configuration.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        :param polling_delay: Time in second the source will sleep when `poll` returns `None`
```
ghost:
```suggestion
        :param polling_delay: Time in seconds the source will sleep when `poll` returns `None`.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+        :param shutdown_timeout: Time in second the application waits for the source to gracefully shutdown
```
ghost:
```suggestion
        :param shutdown_timeout: Time in seconds the application will wait for the source to shut down gracefully.
```

quixstreams/sources/base.py
```diff
@@ -0,0 +1,285 @@
...
+    @abstractmethod
+    def poll(self) -> Optional[KafkaMessage]:
+        """
+        This method is triggered when the source needs new data
+
+        You can return :
+            * a :class:`quixstreams.models.messages.KafkaMessage` to produce it
+            * `None` to make the source sleep for `polling_delay`
+
+        or raise a `StopIteration` to shutdown the source.
```
ghost:
```suggestion
        This method is called periodically when the source needs new data.

        You can return :
            * a :class:`quixstreams.models.messages.KafkaMessage` to produce it
            * `None` to make the source sleep for `polling_delay`

        or raise a `StopIteration` to shutdown the source.
```

quixstreams/sources/iterable.py
```diff
@@ -0,0 +1,163 @@
...
+__all__ = (
+    "GeneratorSource",
+    "ValueIterableSource",
+    "KeyValueIterableSource",
+)
+
+
+class ValueIterableSource(PollingSource):
+    """
+    PollingSource implementation that iterate over values with a fix key.
```
ghost:
```suggestion
    A `PollingSource` implementation that iterates over values with a fixed key.
```

quixstreams/sources/iterable.py
```diff
@@ -0,0 +1,163 @@
...
+        data = next(self._iterator)
+
+        if data is None:
+            return data
+
+        key, value = data
+        return self.serialize(key=key, value=value)
+
+
+class GeneratorSource(PollingSource):
```
ghost: Suggestion: 
Let's rename it to `IterableSource` and accept any iterable as a return result (given they return the correct types).
This way, users can also provide callables that return simple lists, and generators will continue to work as they do now.

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class SourceProcess(multiprocessing.Process):
+    """
+    An individual source process
+
+    Used to manage a source and it's process. Handles communication accross the child and parent processes,
+    lifecycle, error handling.
+
+    Part of the methods are designed to be used from the parent process and other from the child process.
```
ghost:
```suggestion
    An implementation of the Source subprocess.

    It manages a source and its subprocess, handles the communication between the child and parent processes,
    lifecycle, and error handling.

    Some methods are designed to be used from the parent process, and others from the child process.
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+
+    def __init__(self, source):
+        super().__init__()
+        self.source: BaseSource = source
+
+        self._exceptions: List[Exception] = []
+        self._stopping = False
+
+        # copy parent process loglevel to child process
```
ghost:
```suggestion
        # copy parent process log level to the child process
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def run(self) -> None:
+        """
+        Entrypoing of the child process
+
+        Responsible for:
+            * Configuring the signal handlers to properly handle shutdown
+            * Execute the source `run` method
+            * Report the source exception to the parent process
+            * Execute cleanup
```
ghost:
```suggestion
        An entrypoint of the child process.

        Responsible for:
            * Configuring the signal handlers to handle shutdown properly
            * Execution of the source `run` method
            * Reporting the source exceptions to the parent process
            * Cleanup when the source stops
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+    def _cleanup(self, failed: bool) -> None:
+        """
+        Execute post-run cleanup
+        """
+        logger.debug("Cleaning up source")
+        self.source.cleanup(failed)
+
+    def _stop(self, signum, _):
+        """
+        Ask the source to stop execution
```
ghost:
```suggestion
        Stop the source to execution.
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+
+        try:
+            self.source.stop()
+        except BaseException as err:
+            logger.exception("Error stopping source")
+            self._report_exception(err)
+
+    def _report_exception(self, err: Exception) -> None:
+        """
+        Write exception to the pipe so the parent process can access it.
```
ghost:
```suggestion
        Write an exception to the pipe so the parent process can access it.
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+    # --- PARENT PROCESS METHODS --- #
+
+    def start(self) -> "SourceProcess":
+        logger.info("Starting source %s", self.source)
+        return super().start()
+
+    def raise_for_error(self) -> bool:
+        """
+        If the child process has terminated with an exception raises a
+        :class:`quixstreams.sources.manager.SourceException`.
```
ghost:
```suggestion
        Raise a :class:`quixstreams.sources.manager.SourceException`
        if the child process was terminated with an exception.
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,255 @@
...
+        if self.exitcode not in (0, -signal.SIGTERM):
+            raise SourceException(self)
+
+    def stop(self):
+        """
+        Handle shutdown of the source and the associated process
+
+        First try a graceful shutdown by sending a SIGTERM and waiting up to
+        `source.shutdown_timeout` seconds for the process to exit. If the process
+        is still alive force kill it with a SIGKILL/
```
ghost:
```suggestion
        Handle shutdown of the source and its subprocess.

        First, it tries to shut down gracefully by sending a SIGTERM and waiting up to
        `source.shutdown_timeout` seconds for the process to exit. If the process
        is still alive, it will kill it with a SIGKILL.
```

quixstreams/sources/base.py
ghost: THe logger is already configured to display the source name

quixstreams/app.py
```diff
@@ -745,6 +790,19 @@ def run(
 
         :param dataframe: instance of `StreamingDataFrame`
         """
+        self._run(dataframe)
+
+    def _exception_handler(self, exc_type, exc_val, exc_tb):
+        fail = False
+
+        # Sources and the application are independant, if a source fails
+        # the application can shutdown gracefully.
```
ghost:
```suggestion
        # Sources and the application are independent.
        # If a source fails, the application can shutdown gracefully.
```

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,241 @@
...
+        An entrypoint of the child process.
+
+        Responsible for:
+            * Configuring the signal handlers to handle shutdown properly
+            * Execution of the source `run` method
+            * Reporting the source exceptions to the parent process
+        """
+        self._setup_signal_handlers()
+        configure_logging(self._loglevel, str(self.source), pid=True)
+        logger.info("Source started")
```
ghost, line 59: Let's add the source name here too.
By default I see this:
```[2024-09-04 16:22:18,951] [INFO] [quixstreams] : Source started```

Comment (ghost) [reply quixstreams/sources/manager.py:59]: Do you have any logging configuration ? I get
```
[2024-09-04 16:32:19,389] [INFO] [my_source] [165511] : Source started
```

File: quixstreams/sources/manager.py [hunk 48]
```diff
@@ -0,0 +1,241 @@
...
+
+        try:
+            self.source.start()
+        except BaseException as err:
+            logger.exception(f"Error in source")
+            self._report_exception(err)
+
+        logger.info("Source completed")
```
ghost, line 67: Please add a source name here too

quixstreams/sources/manager.py
```diff
@@ -0,0 +1,241 @@
...
+        `source.shutdown_timeout` seconds for the process to exit. If the process
+        is still alive, it will kill it with a SIGKILL.
+        """
+        if self.is_alive():
+            logger.info("Stopping source %s", self.source)
+            self.terminate()
+            self.join(self.source.shutdown_timeout)
+
+        if self.is_alive():
+            logger.info("Force stopping source %s", self.source)
```
ghost: This should be `logger.error`, and we should also say why it's force-stopped.

quixstreams/sources/manager.py
ghost: same as :point_up:

quixstreams/sources/manager.py
ghost: Nope, just a bare minimum app:
```app = Application(broker_address="localhost:9092", auto_offset_reset="earliest")```

ghost (approved):
@quentin-quix merge away 👍

PR #434, opened 2024-07-30
Allow to ignore missing columns when calling "SDF.drop()"

ghost:
Added a new parameter `errors="raise" | "ignore"` to the `StreamingDataFrame.drop()` method to ignore potentially missing columns.

The default behavior is to fail if one of the columns to drop is missing.

The parameter is named after `pandas.DataFrame.drop()` for consistency.

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -966,7 +966,11 @@ class StreamingDataFrame(BaseStreaming):
             name=name,
         )
 
-    def drop(self, columns: Union[str, List[str]]) -> Self:
+    def drop(
+        self,
+        columns: Union[str, List[str]],
+        errors: Literal["ignore", "raise"] = "raise",
+    ) -> Self:
         """
         Drop column(s) from the message value (value must support `del`, like a dict).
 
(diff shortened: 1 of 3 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -966,7 +966,11 @@ def hopping_window(
             name=name,
         )
 
-    def drop(self, columns: Union[str, List[str]]) -> Self:
+    def drop(
+        self,
+        columns: Union[str, List[str]],
+        errors: Literal["ignore", "raise"] = "raise",
```
gwaramadze, line 972: That looks like a boolean. Do you expect there might ever be more than 2 options? Maybe for simplicity it is worth exposing literally `ignore_missing: bool = False`?

ghost: Hey @gwaramadze , the API is made to match [the same method in Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html)

>          
    errors:  {‘ignore’, ‘raise’}, default ‘raise’
        If ‘ignore’, suppress error and only existing labels are dropped.

gwaramadze (approved):

PR #443, opened 2024-08-09
suggestions/ideas for sources

tim-quix:
The big thing I adjusted here was having to pass around as much stuff, and reducing the overall responsibilities of the various objects.

Basically, since `Application.source()` is setting up a bunch of stuff anyway, went ahead and put a few more operations there and remove the need for `Source` objects to manage those assets as well.

- `SourceManager` now expects "configured" sources during registration, which now happens in `Application.source()`

- `SourceThread` now tracks less things and sticks to managing threads for a source.

- Users don't need to be aware of `TopicManager` now when making a Source

Granted, I haven't tested these changes and maybe sources that have commits will negate some of these changes, but this was just something that popped out to me.

Use whatever you think makes sense, and feel free to ask me questions =)

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -609,7 +609,7 @@ class Application:
         if not source and not topic:
             raise TypeError("one of `source` or `topic` is required")
         elif source and not topic:
-            topic = source.default_topic(self._topic_manager)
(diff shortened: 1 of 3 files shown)


tim-quix:
Quentin has added changes as needed.

PR #445, opened 2024-08-15
fix(docs): correcting the output showcased for word count with other minor corrections

shrutimantri:
The output is a tuple, and not a list. Hence, the correction.

### DIFF
--- a/docs/tutorials/word-count/tutorial.md
+++ b/docs/tutorials/word-count/tutorial.md
@@ -14,7 +14,7 @@ You'll learn how to:
 
 ## Outline of the Problem
 
-Imagine we are a company with various products. 
+Imagine we are a company with various products.
 
 We want to process text reviews for our various products in real time 
 and see what words are the most popular across all of them.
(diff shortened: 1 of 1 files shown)


docs/tutorials/word-count/tutorial.md
```diff
@@ -202,8 +202,8 @@ In the end we would produce 5 messages in total, like so:
 
 ```python
 # two shown here...
-{kafka_key: "bob", kafka_value: ["bob", 1]}
-{kafka_key: "likes", kafka_value: ["likes", 2]}
+{kafka_key: "bob", kafka_value: ("bob", 1)}
+{kafka_key: "likes", kafka_value: ("likes", 2)}
```
tim-quix: So it turns out the original is technically correct (though in reality it's `{kafka_key: b"bob", kafka_value: '["bob", 1]'}`, etc..., but I wrote it like this to make it a little more reader-friendly =) )

It actually converts the tuples into a list when it does `JSON` serialization, but I admit that's not super obvious behavior. Rather than changing this, we could instead mention that below it as like, a note.

tim-quix (commented):
Nice cleanup! All looks good besides the commented change; will approve once that's changed. 

Thanks!

docs/tutorials/word-count/tutorial.md
shrutimantri: Ack. My bad.. this is how they were appearing on printing. But the values in Kafka show them with [].
I have reverted that change, and added a note instead.

tim-quix (approved):

PR #447, opened 2024-08-16
Schema Registry 1: Serialization Context

gwaramadze:
This change is required in the context of upcoming Schema Registry support. The SerializationContext.field attribute is expected for the default subject name strategy.

More on subject name strategies:
https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#subject-name-strategy

This may be considered a breaking change for two reasons:

1. The new SerializationContext.field attribute is required. There is no way to resolve a default value; it must be passed from a higher scope.
2. The SerializationContext.to_confluent_ctx method has been removed. Since we must accept the field attribute, maintaining this method no longer makes sense.

### DIFF
--- a/quixstreams/models/topics/topic.py
+++ b/quixstreams/models/topics/topic.py
@@ -191,7 +191,7 @@ class Topic:
             value_deserialized = (
                 None
                 if value_bytes is None
-                else self._value_deserializer(value=message.value(), ctx=ctx)
+                else self._value_deserializer(value=value_bytes, ctx=ctx)
             )
         except IgnoreMessage:
             # Ignore message completely if deserializer raised IgnoreValueError.
--- a/quixstreams/models/topics/topic.py
+++ b/quixstreams/models/topics/topic.py
@@ -147,7 +147,10 @@ class Topic:
         # Try to serialize the key only if it's not None
         # If key is None then pass it as is
         # Otherwise, different serializers may serialize None differently
-        key_serialized = None if key is None else self._key_serializer(key, ctx=ctx)
+        if key is None:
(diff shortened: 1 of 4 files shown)


quixstreams/models/topics/topic.py
```diff
@@ -144,18 +145,24 @@ def row_serialize(self, row: Row, key: Any) -> KafkaMessage:
             raise SerializerIsNotProvidedError(
                 f'Value serializer is not provided for topic "{self.name}"'
             )
+
+        ctx = partial(SerializationContext, topic=self.name, headers=row.headers)
```
ghost: Hey, as a small optimization, I suggest not using `partial` and initializing the `SerializationContext` directly here.

It's more verbose, but deserialization happens for each message, and partials do additional attribute lookups internally and are slightly slower than a normal function call.

quixstreams/models/topics/topic.py
```diff
@@ -178,33 +185,34 @@ def row_deserialize(
             )
 
         headers = message.headers()
-        ctx = SerializationContext(topic=message.topic(), headers=headers)
+        ctx = partial(SerializationContext, topic=message.topic(), headers=headers)
```
ghost: Same here for partials

ghost (changes requested):
It looks good except a small thing around using `partial` in the critical path.

quixstreams/models/topics/topic.py
gwaramadze: Fixed!

ghost (approved):

PR #448, opened 2024-08-16
Kafka Replicator Source & Quix Environment Source

ghost:


quixstreams/sources/__init__.py
```diff
@@ -2,3 +2,4 @@
 from .manager import SourceException
 from .multiprocessing import multiprocessing
 from .csv import CSVSource
+from .kafka import KafkaSource, QuixEnvironmentSource
```
daniil-quix, line 5: Let's move the `QuixEnvironmentSource` to a separate `quix` package so it's easier to follow the structure.

ghost: `QuixEnvironmentSource` should be imported from the `quix` package

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def produce_message(self, msg: Message):
+        topic_name, partition, offset = msg.topic(), msg.partition(), msg.offset()
+        self._checkpoint.store_offset(topic_name, partition, offset)
+        logger.log(
+            5,
+            "producing message to partition %s with offset %s",
+            partition,
+            offset,
+        )
```
ghost: Why to use `logger.log(5...)` here instead of `logger.debug` or similar?

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["KafkaSource"]
+
+
+class KafkaSource(Source):
+    """
+    Source implementation that replicate a topic from a kafka cluster to your application cluster.
```
ghost:
```suggestion
    A Source implementation that replicates a topic from a separate Kafka broker to your Application broker.
    
    It can copy messages for development and testing without risking producing them back or affecting the consumer groups.
```

quixstreams/sources/kafka/quix.py
```diff
@@ -0,0 +1,110 @@
...
+if TYPE_CHECKING:
+    from quixstreams.app import ApplicationConfig
+
+
+__all__ = ["QuixEnvironmentSource"]
+
+
+class QuixEnvironmentSource(KafkaSource):
+    """
+    Source implementation that replicate a topic from a quix workspace to your application cluster.
```
ghost:
```suggestion
    A Source implementation that replicates a topic from an environment in Quix Cloud to your application cluster.  
    
    It can copy messages for development and testing without risking producing them back or affecting the consumer groups.
```

quixstreams/sources/kafka/kafka.py
ghost: Because it's very very spamy but useful when doing in depth debugging of the source. 5 is lower than debug.

quixstreams/sources/kafka/kafka.py
ghost: > It can copy messages for development and testing without risking producing them back or affecting the consumer groups.

This should should be fine for production use. I don't think this is necessary.

quixstreams/sources/__init__.py
ghost: Both the KafkaSource and QuixEnvSource are coming from the `kafka` package. I don't think we should split it since QuixEnvSource inherit from KafkaSource

quixstreams/sources/kafka/kafka.py
ghost: In that case, how is it going to be visible if most people use `DEBUG`, `INFO` and similar values to set a log level?

ghost: It's not. That's the point. We don't want to have one log per message in `DEBUG`. I'd rather remove the log if that's an issue.

ghost: If there's no easy way to print this log, then let's remove it 👍

mkdocs.yml
```diff
@@ -52,6 +52,8 @@ nav:
     - Sources:
         - 'connectors/sources/README.md'
         - CSV Source: connectors/sources/csv-source.md
+        - Kafka Source: connectors/sources/kafka-source.md
```
ghost:
```suggestion
        - Kafka Replicator Source: connectors/sources/kafka-source.md
```

docs/connectors/sources/kafka-source.md
```diff
@@ -0,0 +1,35 @@
...
+
+  app.run(sdf)
+
+if __name__ == "__main__":
+  main()
+```
+
+## Topic
+
+The Kafka Replicator source only deal with bytes. It read the remote keys and values as bytes and produce them directly as bytes. You can configure the key and value deserializer used by the Streaming Dataframe with the `key_deserializer` and `value_deserializer` paramaters.
```
ghost:
```suggestion
The Kafka Replicator source only deals with bytes. It reads the remote keys and values as bytes and produces them directly as bytes.
You can configure the key and value deserializer used by the Streaming Dataframe with the `key_deserializer` and `value_deserializer` paramaters.
```

docs/connectors/sources/quix-source.md
```diff
@@ -0,0 +1,34 @@
+# Quix Environment Source
+
+A specialised [Kafka Source](kafka-source.md) that simplify copying data from a Quix environment.
+
+## How to use the Quix Environment Source
+
+To use a Quix Environment source, you need to create an instance of `QuixSource` and pass it to the `app.dataframe()` method.
```
ghost:
```suggestion
A specialised [Kafka Source](kafka-source.md) that simplify copying data from a Quix environment.

## How to use the Quix Environment Source

To use a Quix Environment source, you need to create an instance of `QuixEnvironmentSource` and pass it to the `app.dataframe()` method.
```

quixstreams/kafka/consumer.py
```diff
@@ -556,3 +556,13 @@ def __enter__(self):
...
+
+    # support pickling by dropping the inner consumer
+    def __getstate__(self) -> object:
+        state = self.__dict__.copy()
+        state.pop("_inner_consumer", None)
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._inner_consumer = None
```
ghost: Do we still need to pickle Consumers?
If not, let's remove this code

quixstreams/sources/kafka/checkpoint.py
```diff
@@ -0,0 +1,84 @@
...
+        if self._exactly_once:
+            self._producer.abort_transaction()
+
+    def commit(self):
+        """
+        Commit the checkpoint.
+
+        This method will:
+         2. Flush the producer to ensure everything is delivered.
+         3. Commit topic offsets.
```
ghost:
```suggestion
         1. Flush the producer to ensure everything is delivered.
         2. Commit topic offsets.
```

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,340 @@
...
+            self._producer_topic.name,
+            target_topic_config,
+        )
+
+        if source_topic_config.num_partitions > target_topic_config.num_partitions:
+            raise ValueError("Source topic has more partitions than destination topic")
+        elif source_topic_config.num_partitions < target_topic_config.num_partitions:
+            logger.warning("Source topic has less partitions than destination topic")
+
+    def _target_cluster_offsets(self, partitions) -> Dict[int, int]:
```
ghost:
```suggestion
    def _target_cluster_offsets(self, partitions: List[TopicPartition]) -> Dict[int, int]:
```

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+            for partition in partitions
+        ]
+        partitions_commited = self._target_cluster_consumer.committed(
+            partitions, timeout=self._config.request_timeout
+        )
+        a = {partition.partition: partition.offset for partition in partitions_commited}
+        return a
+
+    def on_assign(self, _, source_partitions) -> None:
```
ghost:
```suggestion
    def on_assign(self, _, source_partitions: List[TopicPartition]) -> None:
```

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,340 @@
...
+                    partition.topic,
+                    partition.partition,
+                )
+
+            self._source_cluster_consumer.incremental_assign(source_partitions)
+        except Exception as exc:
+            logger.exception("Error assigning partitions")
+            self._error = exc
+
+    def on_revoke(self, _, partitions) -> None:
```
ghost:
```suggestion
    def on_revoke(self, *_) -> None:
```

quixstreams/sources/kafka/checkpoint.py
```diff
@@ -0,0 +1,84 @@
...
+class Checkpoint(BaseCheckpoint):
+    """
+    Checkpoint implementation used by the KafkaReplicatorSource
+    """
+
+    def __init__(
+        self,
+        producer: RowProducer,
+        producer_topic: Topic,
+        consumer: RowConsumer,
```
ghost: The type hint should be `consumer: Consumer` probably

quixstreams/sources/kafka/kafka.py
```diff
@@ -0,0 +1,323 @@
...
+        auto_offset_reset: AutoOffsetReset = "latest",
+        consumer_extra_config: Optional[dict] = None,
+        consumer_poll_timeout: Optional[float] = None,
+        shutdown_timeout: float = 10,
+        on_consumer_error: Optional[ConsumerErrorCallback] = default_on_consumer_error,
+        value_deserializer: DeserializerType = "json",
+        key_deserializer: DeserializerType = "bytes",
+    ) -> None:
+        """
+        :param name: The source unique name. Used to generate the default topic name.
```
ghost:
```suggestion
        :param name: The source unique name.  
            It is used to generate the default topic name and consumer group name on the source broker.
            Running multiple instances of `KafkaReplicatorSource` with the same name connected 
            to the same broker will make them share the same consumer group.
```

ghost (approved):

PR #451, opened 2024-08-20
Schema Registry 3: Avro

gwaramadze:
Schema Registry integration with AvroSerializer and AvroDeserializer

### DIFF
--- a/quixstreams/kafka/configuration.py
+++ b/quixstreams/kafka/configuration.py
@@ -1,13 +1,11 @@
 from typing import Literal, Optional, Callable, Tuple, Type
 
-from pydantic import AliasGenerator, SecretStr, AliasChoices, Field
+from pydantic import SecretStr, AliasChoices, Field
 from pydantic.functional_validators import BeforeValidator
-from pydantic_settings import (
-    BaseSettings,
-    SettingsConfigDict,
-    PydanticBaseSettingsSource,
-)
+from pydantic_settings import PydanticBaseSettingsSource
 from typing_extensions import Self, Annotated
+
+from quixstreams.utils.settings import BaseSettings
 
 __all__ = ("ConnectionConfig",)
 
@@ -22,13 +20,6 @@ class ConnectionConfig(BaseSettings):
 
     Also obscures secrets and handles any case sensitivity issues.
     """
-
-    model_config = SettingsConfigDict(
-        alias_generator=AliasGenerator(
-            # used during model_dumps
-            serialization_alias=lambda field_name: field_name.replace("_", "."),
-        ),
-    )
 
     bootstrap_servers: str
     security_protocol: Annotated[
@@ -122,7 +113,7 @@ class ConnectionConfig(BaseSettings):
             config = {k: v for k, v in config.items() if k in valid_keys}
         return cls(**config)
 
-    def as_librdkafka_dict(self, plaintext_secrets=True) -> dict:
+    def as_librdkafka_dict(self, plaintext_secrets: bool = True) -> dict:
         """
         Dump any non-empty config values as a librdkafka dictionary.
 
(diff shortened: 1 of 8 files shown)


quixstreams/kafka/configuration.py
```diff
@@ -131,12 +122,4 @@ def as_librdkafka_dict(self, plaintext_secrets=True) -> dict:
...
-        dump = self.model_dump(by_alias=True, exclude_none=True)
-        if plaintext_secrets:
-            for field, value in dump.items():
-                if isinstance(value, SecretStr):
-                    dump[field] = dump[field].get_secret_value()
-        return dump
-
-    def __str__(self) -> str:
-        return str(self.as_librdkafka_dict(plaintext_secrets=False))
+        return self.as_dict((plaintext_secrets))
```
ghost: Looks like a typo here:
`self.as_dict((plaintext_secrets)) == self.as_dict(plaintext_secrets)`

quixstreams/models/serializers/avro.py
```diff
@@ -31,14 +40,30 @@ def __init__(
...
+            )
+            self._schema_registry_serializer = _AvroSerializer(
+                schema_registry_client=SchemaRegistryClient(client_config),
+                schema_str=json.dumps(schema),
+                conf=schema_registry_config.serialization.as_dict(),
+            )
 
     def __call__(self, value: Any, ctx: SerializationContext) -> bytes:
-        data = BytesIO()
+        if self._schema_registry_serializer:
```
ghost: Can you replace it with `is None` check? 
It's more to the point and avoids calling the `__bool__` for the `_schema_registry_serializer` object.

quixstreams/models/serializers/avro.py
```diff
@@ -85,19 +111,40 @@ def __init__(
...
+            )
+            self._schema_registry_deserializer = _AvroDeserializer(
+                schema_registry_client=SchemaRegistryClient(client_config),
+                schema_str=json.dumps(schema) if schema else None,
+            )
 
     def __call__(
         self, value: bytes, ctx: SerializationContext
     ) -> Union[Iterable[Mapping], Mapping]:
+        if self._schema_registry_deserializer:
```
ghost: Same here for `is None`

quixstreams/schema_registry/configuration.py
```diff
@@ -0,0 +1,49 @@
...
+
+class SchemaRegistryConfig(BaseSettings):
+    """
+    Provides an interface to integrate with a Schema Registry.
+    """
+
+    client: SchemaRegistryClientConfig
+    serialization: SchemaRegistrySerializationConfig = (
+        SchemaRegistrySerializationConfig()
+    )
```
ghost: Is it necessary to nest the settings this way? It makes the initialization more complicated and requires more imports, etc.
The attributes in these objects don't intersect, so why don't we have a single object with documented initialization params?

ghost (changes requested):

quixstreams/schema_registry/configuration.py
ghost: Also, let's add docstrings for individual params too, so it's clear from the code what needs to be passed

gwaramadze: I did that for two reasons:

1. Even if they don't intersect, client and serializer will complain if it receives a bigger dict:  [here](https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/schema_registry_client.py#L118) and [here](https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/avro.py#L228). This can be mitigated with a bit more verbose `SchemaRegistryConfig.as_dict` call with specific fields to include.
2. Hypothetically someone could have a single `SchemaRegistryClientConfig` config but several distinct `SchemaRegistrySerializationConfig` instances to pass down to multiple distinct serializers. In this case some attributes in flat `SchemaRegistryConfig` would have to be duplicated. But all in all, maybe that problem is both, not very probable and not very severe.

Given those considerations, let me know what your preference would be?

ghost: What if we simply have only two objects then: one for schema registry urls, and another for schema registry params. 
And Serializers will accept two params instead of one.

In this case, you at least don't need a third class two wrap another two.

quixstreams/models/serializers/avro.py
gwaramadze: Good point! But if you don't mind I will keep the original checks in the __init__ functions where performance is not so important.

ghost (approved):

PR #458, opened 2024-08-22
Schema Registry 5: Protobuf

gwaramadze:
Schema Registry integration with ProtobufSerializer and ProtobufDeserializer


quixstreams/utils/settings.py
```diff
@@ -13,14 +15,16 @@ class BaseSettings(_BaseSettings):
...
     )
 
-    def as_dict(self, plaintext_secrets: bool = False) -> dict:
+    def as_dict(
+        self, plaintext_secrets: bool = False, include: Optional[Set[str]] = None
+    ) -> dict:
         """
         Dump any non-empty config values as a dictionary.
 
         :param plaintext_secrets: whether secret values are plaintext or obscured (***)
```
ghost, line 24:
```suggestion
        :param plaintext_secrets: whether secret values are plaintext or obscured (***)
        :param include: optional list of fields to be included in the dictionary
```

quixstreams/models/serializers/schema_registry.py
```diff
@@ -57,9 +62,25 @@ class SchemaRegistrySerializationConfig(BaseSettings):
...
     normalize_schemas: bool = False
     use_latest_version: bool = False
     subject_name_strategy: SubjectNameStrategy = topic_subject_name_strategy
+
+    # Protobuf-only atrributes
+    skip_known_types: bool = False
+    reference_subject_name_strategy: ReferenceSubjectNameStrategy = (
+        reference_subject_name_strategy
+    )
+    use_deprecated_format: bool = False
```
ghost, line 86: If these are protobuf-only parameters, should we maybe prefix them with `protobuf_` ?
We could map the names with pydantic easily

ghost (approved):
Nicely done, approved.
There's only one minor comment regarding the parameter naming in `SchemaRegistrySerializationConfig`, please have a look.

quixstreams/models/serializers/schema_registry.py
gwaramadze: I am against this, my arguments: 

We mimic the Confluent config 1 to 1, which they did not do on their end. It will confuse our users when they port their configs to Quix.

I will give you an example. Imagine a user with a Protobuf use-case, they don't even know what Avro or JSON Schema is (hypothetically). They'd initialize the config like this:

```python
ser_conf = SchemaRegistrySerializationConfig(
    auto_register_schemas=False,
    normalize_schemas=False,
    subject_name_strategy=some_func,
    protobuf_skip_known_types=True,
    protobuf_reference_subject_name_strategy=some_other_func,
)
```

Some params are prefixed, others not, which is rather inconsistent.

PR #470, opened 2024-08-27
Application config API

ghost:
Provide an API to export the application configuration

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -4,10 +4,13 @@ import logging
 import os
 import signal
 import warnings
-from typing import Optional, List, Callable, Union, Literal, get_args
+from typing import Optional, List, Callable, Union, Literal, Tuple, Type
 
 from confluent_kafka import TopicPartition
 from typing_extensions import Self
+from pathlib import Path
+from pydantic import Field, AliasGenerator
+from pydantic_settings import PydanticBaseSettingsSource, SettingsConfigDict
 
 from .context import set_message_context, copy_context
 from .dataframe import StreamingDataFrame
@@ -41,8 +44,9 @@ from .sinks import SinkManager
 from .state import StateStoreManager
 from .state.recovery import RecoveryManager
 from .state.rocksdb import RocksDBOptionsType
+from .utils.settings import BaseSettings
 
-__all__ = ("Application",)
+__all__ = ("Application", "ApplicationConfig")
 
 logger = logging.getLogger(__name__)
 ProcessingGuarantee = Literal["at-least-once", "exactly-once"]
@@ -155,7 +159,7 @@ class Application:
         :param producer_extra_config: A dictionary with additional options that
             will be passed to `confluent_kafka.Producer` as is.
         :param state_dir: path to the application state directory.
-            Default - `".state"`.
+            Default - `"state"`.
         :param rocksdb_options: RocksDB options.
             If `None`, the default options will be used.
         :param consumer_poll_timeout: timeout for `RowConsumer.poll()`. Default - `1.0`s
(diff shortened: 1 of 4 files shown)


quixstreams/app.py
```diff
@@ -1,4 +1,5 @@
 import contextlib
+import dataclasses
```
ghost: Hey @quentin-quix , we're using `pydantic` already at some places for configuration.
Can we use it here for consistency too?

quixstreams/app.py
```diff
@@ -263,85 +253,102 @@ def __init__(
...
+                **producer_extra_config,
+            },
+            consumer_extra_config=consumer_extra_config,
+            processing_guarantee=processing_guarantee,
+            consumer_poll_timeout=consumer_poll_timeout,
+            producer_poll_timeout=producer_poll_timeout,
+            auto_create_topics=auto_create_topics,
+            on_consumer_error=on_consumer_error or default_on_consumer_error,
+            on_processing_error=on_processing_error or default_on_processing_error,
+            on_producer_error=on_producer_error or default_on_producer_error,
```
ghost: I don't know if callbacks belong to this Config object.
There is no sense in re-using them elsewhere except in the `Application` class.

Let's limit the scope of this class only to those values that should be shared.

ghost (changes requested):
Looks good overall. 
Only a few minor changes are needed, please see the comments.

quixstreams/app.py
ghost: I'm not really a fan of pydantic and I don't see what it provide in this case. We don't have a very complex validation and I don't think we want to let those setting be easily overrided by environment variable.

quixstreams/app.py
```diff
@@ -964,3 +968,69 @@ def _on_sigint(self, *_):
...
+                f'{get_args(ProcessingGuarantee)}, got "{self.processing_guarantee}"'
+            )
+
+        if self.auto_offset_reset not in get_args(AutoOffsetReset):
+            raise ValueError(
+                f'Must provide a valid "auto_offset_reset"; expected one of: '
+                f'{get_args(AutoOffsetReset)}, got "{self.auto_offset_reset}"'
+            )
+
+    def copy(self, **kwargs) -> "ApplicationConfig":
```
ghost: `-> "ApplicationConfig"` can be replaced with `-> Self` from the `typing_extensions` module.

quixstreams/app.py
```diff
@@ -263,85 +251,101 @@ def __init__(
...
-            exactly_once=self._uses_exactly_once,
+            exactly_once=self._config.exactly_once,
             sink_manager=self._sink_manager,
             pausing_manager=self._pausing_manager,
         )
 
     @property
-    def is_quix_app(self) -> bool:
-        return self._is_quix_app
+    def config(self):
```
ghost: Missing type hint `-> ApplicationConfig` here

quixstreams/app.py
```diff
@@ -964,3 +968,69 @@ def _on_sigint(self, *_):
...
+    auto_offset_reset: AutoOffsetReset = "latest"
+    commit_interval: float = 5.0
+    commit_every: int = 0
+    producer_extra_config: dict = dataclasses.field(default_factory=dict)
+    consumer_extra_config: dict = dataclasses.field(default_factory=dict)
+    processing_guarantee: ProcessingGuarantee = "at-least-once"
+    consumer_poll_timeout: float = 1.0
+    producer_poll_timeout: float = 0.0
+    auto_create_topics: bool = True
+    on_message_processed: Optional[MessageProcessedCallback] = None
```
ghost: The `on_message_processed` callback should not be here too

quixstreams/app.py
ghost: > We don't have a very complex validation and I don't think we want to let those setting be easily overrided by environment variable.

I see your point, but there's a `validate()` method.
The env variables overrides can also be disabled; the `ConnectionConfig` class has an example.

ghost (approved):

PR #474, opened 2024-08-29
added fix for Series manipulations attempted across branches

tim-quix:
Exceptions are now thrown when Series operations with multiple SDFs are attempted.


tim-quix:
Side note: there _might_ be some really contrived example where an `id()` reference could somehow be re-used (see: [`id()` uniqueness](https://stackoverflow.com/questions/52096582/how-unique-is-pythons-id)), but I couldn't think of way of it occurring in a way that would affect this implementation, especially since assigning a column value requires an object to exist.

I think we'd otherwise have to generate a UUID and match on that instead (and pass it around in SDF). Seems overkill?

PR #489, opened 2024-09-09
Optimize get_window_ranges function

gwaramadze:
deque.appendleft is roughly 10x faster than list.insert(0, ...)

Code snippet to test it

```python
import timeit
from collections import deque

# Function to test list.insert(0, ...)
def list_insert():
    my_list = []
    for _ in range(10000):
        my_list.insert(0, 1)

# Function to test deque.appendleft(1)
def deque_appendleft():
    my_deque = deque()
    for _ in range(10000):
        my_deque.appendleft(1)

# Timing the list.insert(0, ...) operation
list_insert_time = timeit.timeit(list_insert, number=100)

# Timing the deque.appendleft operation
deque_appendleft_time = timeit.timeit(deque_appendleft, number=100)

print(f"list.insert(0, ...): {list_insert_time:.6f} seconds")
print(f"deque.appendleft(): {deque_appendleft_time:.6f} seconds")
```

Results:
```
list.insert(0, ...): 1.535651 seconds
deque.appendleft(): 0.015165 seconds
```

### DIFF
--- a/tests/test_quixstreams/test_dataframe/test_windows/test_base.py
+++ b/tests/test_quixstreams/test_dataframe/test_windows/test_base.py
@@ -16,12 +16,10 @@ class TestGetWindowRanges:
         ],
     )
     def test_get_window_ranges_with_step(self, timestamp, duration, step, expected):
-        assert (
-            get_window_ranges(
-                timestamp_ms=timestamp, duration_ms=duration, step_ms=step
-            )
-            == expected
+        result = get_window_ranges(
+            timestamp_ms=timestamp, duration_ms=duration, step_ms=step
         )
+        assert list(result) == expected
 
     @pytest.mark.parametrize(
         "timestamp, duration, expected",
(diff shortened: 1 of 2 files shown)


tests/test_quixstreams/test_dataframe/test_windows/test_base.py
```diff
@@ -16,12 +18,9 @@ class TestGetWindowRanges:
...
     def test_get_window_ranges_with_step(self, timestamp, duration, step, expected):
-        assert (
-            get_window_ranges(
-                timestamp_ms=timestamp, duration_ms=duration, step_ms=step
-            )
-            == expected
-        )
+        assert get_window_ranges(
+            timestamp_ms=timestamp, duration_ms=duration, step_ms=step
+        ) == deque(expected)
```
ghost: Let's rather do `list(get_window_ranges(...))` here so we can freely change the return type as long as it's iterable without changing the test

gwaramadze: done

ghost (approved):

PR #494, opened 2024-09-10
Source API documentation

ghost:


docs/connectors/sources/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Quix streams provide a source out of the box.
+
+* [CSVSource](./csv-source.md): A source that reads data from a single CSV file.
+
+You can also implement your own, have a look at [Creating a Custom Source](custom-sources.md) for documentation on how to do that.
+
+## Multiprocessing
+
+For good performances, sources are run in a separate process. Setting up, monitoring, and tearing down the process is handled by Quix streams. For multiplatform support, the [spawn](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) start method is used to create the source process and as a side-effect, the source object needs to be pickleable. If a source needs to handle un-pickleable objects, it's best to initialize those in the source subprocess (in the source `start` method).
```
ghost:
```suggestion
For good performance, each source runs in a subprocess.  
Quix Streams automatically manages the subprocess's setting up, monitoring, and tearing down.  

For multiplatform support, Quix Streams starts the source process using the [spawn](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) approach.  
As a side effect, each Source instance must be pickleable. If a source needs to handle unpickleable objects, it's best to initialize those in the source subprocess (in the source `start` method).
```

docs/connectors/sources/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Topics
+
+Sources work by sending data to a Kafka topic. Then the StreamingDataframe consume from that topic.
+Each source provides a default topic based on it's configuration. You can override the topic used by
+specyfying a topic in the `app.dataframe()` method.
```
ghost:
```suggestion
Sources work by sending data to Kafka topics. 
Then StreamingDataFrames consume these topics
.
Each source provides a default topic based on its configuration. You can override the default topic by
specifying a topic using the `app.dataframe()` method.
```

docs/connectors/sources/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+For example, a topic with 4 partitions
```
ghost:
```suggestion
**Example**

Provide a custom topic with four partitions to the source. 
```

docs/connectors/sources/csv-source.md
```diff
@@ -0,0 +1,49 @@
+# CSV Source
+
+A basic source that reads data from a single CSV file.
```
ghost, line 3: It would be nice also to mention that the file is processed only once when the source starts.
I.e., it doesn't watch the file for new updates, and on every restart it will produce new messages to the topic.

docs/connectors/sources/csv-source.md
```diff
@@ -0,0 +1,49 @@
...
+foo7,bar7
+```
+
+## Key and value format
+
+By default the CSV source expect the `key` is a string and the `value` a json object. You can configure the deserializers using the `key_deserializer` and `value_deserializer` paramaters.
+
+## Topic
+
+The default topic used for the CSV source will use the `path` as name and expect keys are strings and values are json objects.
```
ghost:
```suggestion
The default topic used for the CSV source will use the `path` as a name and expect keys to be strings and values to be JSON objects.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
+# Creating a custom source
+
+Quix streams also provides aset of classes made to help users implement additional sources.
```
ghost:
```suggestion
Quix Streams also provides a set of classes to help users implement custom sources.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+
+* [`quixstreams.sources.base.BaseSource`]: This is the base class for all other sources. It defines the must have methods.
+* [`quixstreams.sources.base.Source`]: A subclass of `BaseSource` implementing a serie of method useful for writing sources. We recommend subclassing `Source` instead of `BaseSource`.
```
ghost:
```suggestion
* [`quixstreams.sources.base.Source`]: A subclass of `BaseSource` that implements some methods that are helpful in writing sources. We recommend subclassing `Source` instead of `BaseSource`.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Source
+
+The recomended parent class to write a new source. It handles configuring, starting, stopping the source as well as implement a serie of helpers.
```
ghost:
```suggestion
The recommended parent class to create a new source.   
It handles configuring, starting, and stopping the source, as well as implementing a number of helpers.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+
+To get started implement the `run` method and return when `self.running` is `False`.
```
ghost:
```suggestion
To get started, implement the `run` method and return when `self.running` is `False`.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+                    value=line.strip(),
+                )
+
+                self.produce(
+                    key=msg.key,
+                    value=msg.value,
+                )
+```
+
+For more information see `quixstreams.sources.base.Source` docstrings.
```
ghost:
```suggestion
For more information, see `quixstreams.sources.base.Source` docstrings.
```

ghost: Here, we can link directly to the `Source` class' API docs

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+
+## BaseSource
+
+This is the base class for all sources. It handles configuring the source and define 3 must have methods.
```
ghost:
```suggestion
This is the base class for all sources.  
It handles configuring the source and requires the definition of three must-have methods.
```

docs/connectors/sources/custom-sources.md
```diff
@@ -0,0 +1,49 @@
...
+
+* `start`: This method is called, in the subprocess, when the source is started.
+* `stop`: This method is called, in the subporcess, when the application is shutting down.
+* `default_topic`: This method is called, in the main process, when a topic is not provided with the source.
+
+For more information see `quixstreams.sources.base.BaseSource` docstrings.
```
ghost: Same here, it is better to link to the API docs so readers don't need to open the code.

docs/connectors/sources/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+* [CSVSource](./csv-source.md): A source that reads data from a single CSV file.
+
+You can also implement your own, have a look at [Creating a Custom Source](custom-sources.md) for documentation on how to do that.
+
+## Multiprocessing
+
+For good performance, each source runs in a subprocess. Quix Streams automatically manages the subprocess's setting up, monitoring, and tearing down. 
+
+For multiplatform support, Quix Streams starts the source process using the [spawn](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) approach. As a side effect, each Source instance must be pickleable. If a source needs to handle unpickleable objects, it's best to initialize those in the source subprocess (in the source `start` method).
```
ghost:
```suggestion
For multiplatform support, Quix Streams starts the source process using the [spawn](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) approach.   
As a side effect, each Source instance must be pickleable.   
If a source needs to handle unpickleable objects, it's best to initialize those in the source subprocess (in the `BaseSource.start` or `Source.run` methods).  
```

ghost (approved):

PR #495, opened 2024-09-11
Delay starting sources until the consumer get assigned partitions

ghost:
Only start the sources once the consumer is assigned. Otherwise a source can produce data before the consumer starts. If that happens on a new consumer with `auto_offset_reset` set to `latest` the consumer will not get the source data.


gwaramadze (approved):
LGTM, but I'd run the comment through some LLM. It often suggests changes that boost clarity and style on top of any grammar corrections.

PR #503, opened 2024-09-18
Merging POC: multi-parent streams

tim-quix:
POC of merging based on multi-parent streams.

Now appropriately handles groupby and state store registration.


tim-quix:
Made a new branch that is rebased on main

PR #504, opened 2024-09-18
Merging POC: sdf-level

tim-quix:
Another way of handling merging: tracking happens at the SDF level, where operations called on the current SDF get called on the merged SDFs as well.

Obviously the operation would make more sense as a decorator overall, but didn't wanna mess with that for now (and probably would mess up the typing like the locking did).


tim-quix:
I think we can do away with this approach and just use multi-parent since I've got that working better now...

tim-quix:
Super outdated and prob not the right approach anyway.

PR #505, opened 2024-09-19
Add an example of custom websocket source

ghost:
Added a basic WebSocket Source example that reads data from the Coinbase WS API.
The main goal is to show how to implement a source with a minimal setup and configuration.

The file layout is different from the existing examples.
I want to refactor the other two in the future to make them more contained.

### DIFF
--- a/examples/custom_websocket_source/main.py
+++ b/examples/custom_websocket_source/main.py
@@ -65,22 +65,22 @@ class CoinbaseSource(Source):
 
             # Parse the message and send only "ticker" data to the topic
             msg_json = json.loads(msg)
-            if msg_json["type"] == "ticker":
-                key = msg_json["product_id"]
-                dt = isoparse(msg_json["time"])
-                timestamp = int(dt.timestamp() * 1000)
+            if msg_json["type"] != "ticker":
+                continue
 
-                # Serialize message data to bytes to produce to Kafka
-                kafka_msg = self.serialize(
-                    key=key, value=msg_json, timestamp_ms=timestamp
-                )
+            key = msg_json["product_id"]
+            dt = isoparse(msg_json["time"])
+            timestamp = int(dt.timestamp() * 1000)
 
-                # Produce a serialized message to the Kafka topic
-                self.produce(
-                    value=kafka_msg.value,
-                    key=kafka_msg.key,
-                    timestamp=kafka_msg.timestamp,
-                )
+            # Serialize message data to bytes to produce to Kafka
+            kafka_msg = self.serialize(key=key, value=msg_json, timestamp_ms=timestamp)
+
+            # Produce a serialized message to the Kafka topic
+            self.produce(
+                value=kafka_msg.value,
+                key=kafka_msg.key,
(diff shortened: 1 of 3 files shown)


examples/custom_websocket_source/main.py
```diff
@@ -0,0 +1,109 @@
...
+                msg = ws_conn.recv(timeout=1)
+            except TimeoutError:
+                continue
+            except ConnectionClosedOK:
+                logger.info("Websocket connection closed")
+                break
+
+            # Parse the message and send only "ticker" data to the topic
+            msg_json = json.loads(msg)
+            if msg_json["type"] == "ticker":
```
gwaramadze: Minor: maybe reverse the condition and save one indentation level
```python
if msg_json["type"] != "ticker":
    continue
```

examples/custom_websocket_source/main.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Initialize an Application with Kafka configuration
+app = Application(
+    broker_address="localhost:9092",  # Specify your Kafka broker address here
+    auto_offset_reset="earliest",
+)
+
+# Configure the CoinbaseSource instance
+coinbase_source = CoinbaseSource(
+    name="coinbase-source",  # Pick the unique name for the source. It will
```
gwaramadze: the comment is missing continuation

gwaramadze (changes requested):

gwaramadze (approved):

examples/custom_websocket_source/main.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+        ws_conn.send(json.dumps(subscribe_payload))
+
+        # Start the reading loop
+        while self.running:
+            try:
+                # Receive a message from the websocket
+                msg = ws_conn.recv(timeout=1)
+            except TimeoutError:
+                continue
```
ghost, line 61: I think we should poll the producer here

examples/custom_websocket_source/main.py
```diff
@@ -0,0 +1,111 @@
...
+    product_ids=[
+        "ETH-BTC",
+    ],
+)
+
+# Connect the CoinbaseSource to a StreamingDataFrame
+sdf = app.dataframe(source=coinbase_source)
+
+# Print the incoming messages from the source
+sdf.print()
```
ghost: This would be better in a `main` function

examples/custom_websocket_source/main.py
ghost: Shouldn't it happen automatically on each `.produce()` call?

ghost: It won't happen if you don't receive anything from coinbase

ghost: Currently, there's no public method for `.poll()`

ghost: I'm going to merge it because the source is working.
I created #511  to address the gaps in Producer polling in the future

PR #506, opened 2024-09-19
sources: Ensure process was started before checking for error

ghost:
If the first assignment a consumer get is empty the sources won't be started. In that situation the app run loop will check for error in sources and fail.

### DIFF
--- a/quixstreams/sources/manager.py
+++ b/quixstreams/sources/manager.py
@@ -211,7 +211,8 @@ class SourceManager:
         Raise an exception if any process has stopped with an exception
         """
         for process in self.processes:
-            process.raise_for_error()
+            if process.started:
+                process.raise_for_error()
 
     def is_alive(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


quixstreams/sources/manager.py
```diff
@@ -211,7 +211,8 @@ def raise_for_error(self) -> None:
         Raise an exception if any process has stopped with an exception
         """
         for process in self.processes:
-            process.raise_for_error()
+            if process.started:
```
gwaramadze: Can we move this check to the `SourceProcess.raise_for_error`?

quixstreams/sources/manager.py
```diff
@@ -117,6 +117,9 @@ def raise_for_error(self) -> None:
         Raise a `quixstreams.sources.manager.SourceException`
         if the child process was terminated with an exception.
         """
+        if not self.started:
+            return
```
gwaramadze, line 121: I think that line has no coverage, adding a test should be relatively easy, wdyt?

gwaramadze (approved):

PR #513, opened 2024-09-23
Branching rebase

tim-quix:


quixstreams/app.py
```diff
@@ -788,12 +789,17 @@ def run(
...
-        :param dataframe: instance of `StreamingDataFrame`
         """
-        self._run(dataframe)
+        if dataframe is not None:
+            warnings.warn(
+                "Application.run() received a `dataframe` argument which is "
+                "no longer used (StreamingDataFrames are now tracked automatically); "
+                "the argument should be removed.",
+                DeprecationWarning,
+            )
```
ghost, line 801: Should we just remove it all together if we are doing a major version bump ?

daniil-quix: Agree with that, but let's do it in a separate PR

tim-quix:
opened this just to sanity check everything was good. Merging with #419

PR #514, opened 2024-09-24
Update README.md

ghost:
Add conda install instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ Use Quix Streams to build simple Kafka producer/consumer applications or leverag
 ### Install Quix Streams
 
 ```shell
+# PyPI
 python -m pip install quixstreams
+
+# or conda
+conda install -c conda-forge quixio::quixstreams
 ```
 
 #### Requirements


README.md
```diff
@@ -27,7 +27,11 @@ Use Quix Streams to build simple Kafka producer/consumer applications or leverag
 ### Install Quix Streams
 
 ```shell
+# PyPI
 python -m pip install quixstreams
```
gwaramadze, line 31: hey, lets use this opportunity to simplify it to `pip install quixstreams`?

gwaramadze (approved):

README.md
ghost: This is how [pip recommends it](https://pip.pypa.io/en/stable/cli/pip_install/)

ghost: There was a [discussion](https://bugs.python.org/issue22295) about it too.
I don't remember all the differences, but `python -m pip` seems to be more compatible with different OS and ensures the package is installed to the correct interpreter

PR #520, opened 2024-09-25
Fix consumer group used in QuixEnvironmentSource

### LINKED ISSUE #519: QuixEnvironmentSource fails when piping data from one Quix env to another Quix env because
**Tell us about the bug**
QuixEnvironmentSource fails when piping data from one Quix env to another Quix env because of the consumer group prefixing.
(Quix -> other Kafka works).

I expect Quix-> Quix scenario to work with`QuixEnvironmentSource`

2.11.0

ghost:
Fixes https://github.com/quixio/quix-streams/issues/519

### DIFF
--- a/quixstreams/sources/kafka/kafka.py
+++ b/quixstreams/sources/kafka/kafka.py
@@ -122,11 +122,11 @@ class KafkaReplicatorSource(Source):
         self._target_cluster_admin: Optional[TopicAdmin] = None
 
     @property
-    def source_customer_group(self):
+    def source_consumer_group(self):
         return f"source-{self.name}"
 
     @property
-    def target_customer_group(self):
+    def target_consumer_group(self):
         consumer_group = f"source-{self.name}-offsets"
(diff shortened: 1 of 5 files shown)


quixstreams/sources/kafka/kafka.py
```diff
@@ -121,10 +121,20 @@ def __init__(
...
 
+    @property
+    def source_customer_group(self):
+        return f"source-{self.name}"
+
+    @property
+    def target_customer_group(self):
+        if self._config.consumer_group_prefix:
+            return f"{self._config.consumer_group_prefix}-source-{self.name}-offsets"
+        return f"source-{self.name}-offsets"
```
ghost:
```suggestion
        consumer_group = f'source-{self.name}-offsets'
        if self._config.consumer_group_prefix:
            consumer_group = f"{self._config.consumer_group_prefix}-{consumer_group}"
        return consumer_group
```

ghost (approved):

PR #530, opened 2024-09-27
bugfix for topic creation with quix-based external broker

tim-quix:
Fix two bugs: 

1. the wrong topic name is used during topic creation attempt for Quix external broker topic that already exists: 
  - the external topic name (ex:` "my_ext_topic"`), which already exists, would be correctly identified and NOT prepend the workspace-id (so it stays `"my_ext_topic"`), as desired
  - PROBLEM: Then, during topic creation step per `QuixConfigBuilder`, the workspace_id would be prepended and used to check whether the topic already exists (ex: `"my_ws-my_ext_topic"`), but the actual topic name is `"my_ext_topic"`
  - It looks like the topic needs created, so it tries to create it, sans prefix (`"my_ext_topic"`), as desired.
  - The creation correctly raises an exception that the topic already exists. This should be caught and ignored gracefully, BUT....
  
  2. When attempting to create a Quix topic that already exists, it was looking for an `HTTPError` to catch, but the error is bundled as `QuixApiRequestFailure` instead. So, it raises said exception instead of gracefully ignoring it.

The fixes:
1. whenever topic creation requests are made by the `QuixTopicManager`, it passes a flag to `QuixConfigBuilder.create_topics` to NOT prepend the workspace_id during creation since it's already handled all the naming resolutions.

2. Handle the correct exception.

### DIFF
--- a/quixstreams/platforms/quix/config.py
+++ b/quixstreams/platforms/quix/config.py
@@ -371,6 +371,7 @@ class QuixKafkaConfigsBuilder:
         topics: List[Topic],
         timeout: Optional[float] = None,
         finalize_timeout: Optional[float] = None,
+        prepend: bool = True,
     ):
         """
         Create topics in a Quix cluster.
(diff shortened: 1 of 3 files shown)


gwaramadze (approved):
LGTM

PR #537, opened 2024-10-02
Remove InfluxDB3Sink import from core

ghost:
Remove `InfluxDB3Sink` from `sinks.core` imports to fix failing if `influxdb3-python` is not installed

### DIFF
--- a/quixstreams/sinks/__init__.py
+++ b/quixstreams/sinks/__init__.py
@@ -1,6 +1,5 @@
 from .base import BatchingSink, SinkBatch, BaseSink, SinkBackpressureError, SinkManager
 from .core.csv import CSVSink
-from .core.influxdb3 import InfluxDB3Sink
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


quixstreams/sinks/__init__.py
```diff
@@ -1,13 +1,9 @@
 from .base import BatchingSink, SinkBatch, BaseSink, SinkBackpressureError, SinkManager
-from .core.csv import CSVSink
```
gwaramadze, line 2: why is this one removed? for consistency?

quixstreams/sinks/__init__.py
```diff
@@ -1,13 +1,9 @@
...
-from .core.influxdb3 import InfluxDB3Sink
```
gwaramadze, line 3: I assume we don't want to do `try`/`except ImportError` at all?

gwaramadze (approved):

PR #544, opened 2024-10-03
app: Add option to select store backend

ghost:


quixstreams/state/manager.py
```diff
@@ -130,7 +141,10 @@ def _setup_changelogs(
             )
 
     def register_store(
-        self, topic_name: str, store_name: str = DEFAULT_STATE_STORE_NAME
+        self,
+        topic_name: str,
+        store_name: str = DEFAULT_STATE_STORE_NAME,
+        store_type: Optional[StoreTypes] = None,
```
ghost, line 145: Suggestion here:

What if we use a `typing.Union` here instead of enum?
Seems a bit cleaner

```
from typing import Union, Type, get_args
from quixstreams.state.rocksdb.store import Store

STORE_TYPES = get_args(Union[Type[RocksDBStore]])

>>> RocksDBStore in STORE_TYPES 
True
```

ghost (approved):

quixstreams/state/manager.py
ghost: Not sure I totally understand.

We can't type using `STORE_TYPES` and it feels weird to type `Store` while you can't use any `Store` implementation, only the select supported ones.

We can do a `typing.Literal["memory", "rocksdb"]` if you want.

PR #548, opened 2024-10-04
state: implement memory store

ghost:
Implement a second type of store. The MemoryStore keeps all the data in memory and need to be fully recovered from changelog on each restart


quixstreams/app.py
```diff
@@ -628,6 +628,17 @@ def get_consumer(self, auto_commit_enable: bool = True) -> Consumer:
...
+        self, topic: str, store_name: str = DEFAULT_STATE_STORE_NAME
+    ) -> Store:
+        """
+        Get a store for given name and topic
+        :param topic: topic name
+        :param store_name: store name
+        :return: instance of `Store`
+        """
+        return self._state_manager.get_store(topic, store_name)
+
```
ghost: This method doesn't seem to be used anywhere except in tests. 
If it's not needed, let's remove it from the App to avoid increasing the number of public methods.

quixstreams/state/memory/partition.py
```diff
@@ -0,0 +1,142 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = ("MemoryStorePartition",)
+
+
+class MemoryStorePartition(StorePartition):
+    """
+    Class to access in-memory state.
+
+    Responsabilities:
```
ghost:
```suggestion
    Responsibilities:
    
```

quixstreams/state/memory/partition.py
```diff
@@ -0,0 +1,142 @@
...
+    def write(
+        self,
+        cache: PartitionTransactionCache,
+        processed_offset: Optional[int],
+        changelog_offset: Optional[int],
+    ) -> None:
+        """
+        Write data to the state
+
+        :param data: The modified data
```
ghost:
```suggestion
        :param cache: The partition update cache
```

quixstreams/state/memory/partition.py
```diff
@@ -0,0 +1,142 @@
...
+    def exists(self, key: bytes, cf_name: str = "default") -> bool:
+        """
+        Check if a key is present in the store.
+
+        :param key: a key encoded to `bytes`.
+        :param cf_name: rocksdb column family name. Default - "default"
+        :return: `True` if the key is present, `False` otherwise.
+        """
+        if self._closed:
+            raise RuntimeError("partition is closed")
```
ghost: In what scenarios the Store gets closed, but the methods like `exists` get invoked?
We don't have these checks in `RocksDBStore`

ghost: If the store is closed any read/write operation should fail. We do have the same in `RocksDBStore` but it's handled directly by `RocksDB`, once the underlying `Rdict` is closed most operation on it fails.

ghost: Can you please re-write it to a decorator then? Like it's done in `PartitionTransaction`

quixstreams/state/memory/store.py
```diff
@@ -0,0 +1,48 @@
...
+from .partition import MemoryStorePartition
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("MemoryStore",)
+
+
+class MemoryStore(Store):
+    """
+    In-memory state store
```
ghost:
```suggestion
    In-memory state store.
```

quixstreams/state/memory/store.py
```diff
@@ -0,0 +1,48 @@
...
+
+    It keeps track of individual store partitions and provides access to the
+    partitions' transactions.
+
+    Requires a full state recovery for each partition on assignement.
```
ghost:
```suggestion
    Requires a full state recovery for each partition on assignment.
```

quixstreams/state/memory/store.py
```diff
@@ -0,0 +1,48 @@
...
+        self,
+        name: str,
+        topic: str,
+        changelog_producer_factory: Optional[ChangelogProducerFactory] = None,
+    ) -> None:
+        """
+        :param name: a unique store name
+        :param topic: a topic name for this store
+        :param changelog_producer_factory: a ChangelogProducerFactory instance
+            if using changelogs
```
ghost:
```suggestion
        :param changelog_producer_factory: a ChangelogProducerFactory instance
            if using changelog topics.
```

quixstreams/state/memory/partition.py
```diff
@@ -0,0 +1,142 @@
...
+            dumps=json_dumps,
+            loads=json_loads,
+            changelog_producer=changelog_producer,
+        )
+        self._processed_offset: Optional[int] = None
+        self._changelog_offset: Optional[int] = None
+        self._state: Dict[str, Dict[bytes, Dict[bytes, Any]]] = {
+            "default": {},
+            METADATA_CF_NAME: {},
+        }
```
ghost: The type hint seems to be wrong here: 
Looks like it should be `Dict[str, Dict[bytes, Any]]` instead

ghost (approved):

PR #558, opened 2024-10-08
Refactor WindowedRocksDBPartitionTransaction.get_windows

gwaramadze:
* Make _get_windows a public method
* ~~Refactor get_windows into a generator~~
* ~~Refactor expire_windows into a generator~~
* Enable backwards iteration from get_windows
* Introduce pytest.mark.timeit and the --timeit argument

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,3 +81,5 @@ addopts = "--log-disable=urllib3.connectionpool --log-disable=parso --log-disabl
 log_cli = true
 log_cli_level = "INFO"
 log_cli_format = "[%(levelname)s] %(name)s: %(message)s"
+# Custom markers
+markers = ["timeit"]
--- a/quixstreams/state/rocksdb/windowed/state.py
+++ b/quixstreams/state/rocksdb/windowed/state.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, List, Tuple, TYPE_CHECKING
+from typing import Any, Generator, Optional, List, Tuple, TYPE_CHECKING
 
 from quixstreams.state.types import WindowedState
 
@@ -86,7 +86,7 @@ class WindowedTransactionState(WindowedState):
 
     def get_windows(
         self, start_from_ms: int, start_to_ms: int
-    ) -> List[Tuple[Tuple[int, int], Any]]:
+    ) -> Generator[Tuple[Tuple[int, int], Any], None, None]:
         return self._transaction.get_windows(
             start_from_ms=start_from_ms,
             start_to_ms=start_to_ms,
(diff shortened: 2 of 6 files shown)


gwaramadze:
@daniil-quix This is an alternative drop-in alternative implementation for `get_windows` method.
It is approx. 13% slower than the first proposition, but still up to 30% faster than the original `_get_windows` method.

```python
    def get_windows(
        self,
        start_from_ms: int,
        start_to_ms: int,
        prefix: bytes,
        backwards: bool = False,
    ) -> Generator[Tuple[Tuple[int, int], Any], None, None]:
        """
        Get all windows that start between "start_from_ms" and "start_to_ms"
        within the specified prefix.

        This function also checks the update cache for any updates not yet
        committed to RocksDB.

        :param start_from_ms: The minimal window start time, exclusive.
        :param start_to_ms: The maximum window start time, inclusive.
        :param prefix: The key prefix for filtering windows.
        :param backwards: If True, yields windows in reverse order.
        :return: A generator that yields sorted tuples in the format `((start, end), value)`.
        """
        seek_from = max(start_from_ms, 0)
        seek_from_key = encode_window_prefix(prefix=prefix, start_ms=seek_from)

        # Add +1 to make the upper bound inclusive
        seek_to = start_to_ms + 1
        seek_to_key = encode_window_prefix(prefix=prefix, start_ms=seek_to)

        # Set iterator bounds to reduce IO by limiting the range of keys fetched
        read_opt = ReadOptions()
        read_opt.set_iterate_lower_bound(seek_from_key)
        read_opt.set_iterate_upper_bound(seek_to_key)

        cache = self._update_cache.get("default", {}).get(prefix, {}).items()
        cached_windows = sorted(cache, reverse=True)
        db_windows = self._partition.iter_items(
            read_opt=read_opt, from_key=seek_from_key
        )

        def _next_cached_window() -> Tuple[Optional[bytes], Optional[bytes]]:
            try:
                return cached_windows.pop()
            except IndexError:
                return None, None

        def _process_window(
            key: bytes, value: bytes
        ) -> Generator[Tuple[Tuple[int, int], Any], None, None]:
            if value is DELETED:
                return
            _, start, end = parse_window_key(key)
            if start_from_ms < start <= start_to_ms:
                yield (start, end), self._deserialize_value(value)

        def _get_windows() -> Generator[Tuple[Tuple[int, int], Any], None, None]:
            """
            This internal generator function iterates over database and cache entries.
            Cached entries take precedence over database entries with the same timestamp,
            and non-deleted cache entries are yielded in the appropriate order.
            """
            cached_key, cached_value = _next_cached_window()
            for db_key, db_value in db_windows:
                while cached_key and cached_key < db_key:
                    yield from _process_window(cached_key, cached_value)
                    cached_key, cached_value = _next_cached_window()

                if cached_key != db_key:
                    yield from _process_window(db_key, db_value)

            while cached_key:
                yield from _process_window(cached_key, cached_value)
                cached_key, cached_value = _next_cached_window()

        yield from reversed(list(_get_windows())) if backwards else _get_windows()
```

ghost (approved):

Issue #559, opened 2024-10-08
File sink


tomas-quix:
Ability to store data in topic in files on disk in various formats.

ghost:
Released in 3.2.0

PR #560, opened 2024-10-08
local file sink

tomas-quix:
Only JSON tested.

Issue: https://github.com/quixio/quix-streams/issues/559


quixstreams/sinks/community/file/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+_UNSAFE_CHARACTERS_REGEX = re.compile(r"[^a-zA-Z0-9 ._]")
+
+
+class InvalidFormatError(Exception):
+    """
+    Raised when the format is specified incorrectly.
+    """
+
+
+class FileSink(BatchingSink):
```
tomas-quix: Is it normal to keep class in init file? I would not expect it here.

quixstreams/sinks/community/file/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+
+        If `append` is `True` and an existing file is found, data will be appended to
+        the last file. Otherwise, a new file is created based on the batch's starting
+        offset.
+
+        :param batch: The batch of data to write.
+        """
+
+        # Generate directory based on topic and partition
+        directory = Path(self._output_dir)
```
tomas-quix: This is something that should be in separate method so it is easy to overload.

quixstreams/sinks/community/file/__init__.py
gwaramadze: There is technically nothing wrong with it, but I will move it to file.py and import the class in __init__.py

quixstreams/sinks/community/file/sink.py
```diff
@@ -0,0 +1,167 @@
...
+        If `append` is `True` and an existing file is found, data will be appended to
+        the last file. Otherwise, a new file is created based on the batch's starting
+        offset.
+
+        :param batch: The batch of data to write.
+        """
+
+        # Generate directory based on topic and partition
+        # Generate filename based on a padded offset
+        # Returns existing file path if append is True
```
ghost: Looks like the comment refers to something different here

gwaramadze: Yeah the wording is poor here. I will skip this comment since there is a big on in the method itself.

tim-quix (dismissed):
Tested using my File Source Connector, works :+1:

tim-quix:
One last thing we may wanna add is a docs page before merging

gwaramadze:
Docs page added: https://quixdocsdev.blob.core.windows.net/pr432/quix-streams/connectors/sinks/file-sink.html

ghost (approved):

PR #576, opened 2024-10-14
Abstract away the state update cache

ghost:
- Moved the StateTransaction's update cache logic to a new object - `PartitionTransactionCache`
- Updated the structure of this cache: "update" and "delete" operations are now stored in separate data structures. The "updates" are stored in a dictionary (as before), and "deletes" are stored in a set.
- On the side, optimized the window keys serialization by using a single `struct.pack/unpack` call instead of 2 calls.

### DIFF
--- a/quixstreams/state/base/transaction.py
+++ b/quixstreams/state/base/transaction.py
@@ -76,16 +76,14 @@ class PartitionTransactionCache:
         :param: cf_name: column family name
         """
         # Check if the key has been deleted
-        deleted = key in self._deleted[cf_name]
-        if deleted:
+        if key in self._deleted[cf_name]:
             # The key is deleted and the store doesn't need to be checked
             return DELETED
 
         # Check if the key has been updated
         # If the key is not present in the cache, we need to check the store and return
         # UNDEFINED to signify that
-        cached = self._updated[cf_name][prefix].get(key, UNDEFINED)
-        return cached
+        return self._updated[cf_name][prefix].get(key, UNDEFINED)
 
     def set(self, key: bytes, value: bytes, prefix: bytes, cf_name: str = "default"):
         """
@@ -120,11 +118,11 @@ class PartitionTransactionCache:
         """
         return self._empty
 
-    def get_column_families(self) -> list[str]:
+    def get_column_families(self) -> Set[str]:
         """
         Get all update column families.
         """
-        return list(set(self._updated.keys()) | set(self._deleted.keys()))
+        return set(self._updated.keys()) | set(self._deleted.keys())
 
     def get_updates(self, cf_name: str = "default") -> Dict[bytes, Dict[bytes, bytes]]:
         """
@@ -419,7 +417,7 @@ class PartitionTransaction(ABC):
             }
 
             updates = self._update_cache.get_updates(cf_name=cf_name)
-            for _, prefix_update_cache in updates.items():
+            for prefix_update_cache in updates.values():
                 for key, value in prefix_update_cache.items():
                     self._changelog_producer.produce(
                         key=key,
--- a/quixstreams/state/base/transaction.py
+++ b/quixstreams/state/base/transaction.py
@@ -1,21 +1,11 @@
 import enum
-import logging
 import functools
-from typing import (
-    Any,
-    Optional,
-    Dict,
-    Tuple,
-    Union,
-    TYPE_CHECKING,
-)
-
+import logging
 from abc import ABC
+from collections import defaultdict
+from typing import Any, Optional, Dict, Tuple, Union, TYPE_CHECKING, Set
 
-from quixstreams.state.exceptions import (
-    StateTransactionError,
-    InvalidChangelogOffset,
-)
+from quixstreams.state.exceptions import StateTransactionError, InvalidChangelogOffset
 from quixstreams.state.metadata import (
     CHANGELOG_CF_MESSAGE_HEADER,
     CHANGELOG_PROCESSED_OFFSET_MESSAGE_HEADER,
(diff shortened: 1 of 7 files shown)


quixstreams/state/base/transaction.py
```diff
@@ -25,24 +15,131 @@
...
+        Return True if any changes have been made (updates or deletes), otherwise
+        return False.
+        """
+        return self._empty
+
+    def get_column_families(self) -> list[str]:
+        """
+        Get all update column families.
+        """
+        return list(set(self._updated.keys()) | set(self._deleted.keys()))
```
gwaramadze: I don't think you need to cast it to list, let's return set.

quixstreams/state/base/transaction.py
```diff
@@ -25,24 +15,131 @@
...
+        self._updated[cf_name][prefix].pop(key, None)
+        self._deleted[cf_name].add(key)
+        self._empty = False
+
+    def is_empty(self) -> bool:
+        """
+        Return True if any changes have been made (updates or deletes), otherwise
+        return False.
+        """
+        return self._empty
```
gwaramadze, line 119: Did you consider `return not self._updated and not self._deleted`?

We wouldn't have to maintain self._empty at all. `is_empty` is called less time than `set` and `delete` where `self._empty` assignment happens.

It may be slightly faster, but would be good to check.

quixstreams/state/base/transaction.py
```diff
@@ -25,24 +15,131 @@
...
+        If the key is not present in the cache, returns "UNDEFINED sentinel
+        (we need to check the store).
+
+        :param: key: key as bytes
+        :param: prefix: key prefix as bytes
+        :param: cf_name: column family name
+        """
+        # Check if the key has been deleted
+        deleted = key in self._deleted[cf_name]
+        if deleted:
```
gwaramadze: `if key in self._deleted[cf_name]:` 😉

quixstreams/state/base/transaction.py
```diff
@@ -25,24 +15,131 @@
...
+            # The key is deleted and the store doesn't need to be checked
+            return DELETED
+
+        # Check if the key has been updated
+        # If the key is not present in the cache, we need to check the store and return
+        # UNDEFINED to signify that
+        cached = self._updated[cf_name][prefix].get(key, UNDEFINED)
+        return cached
```
gwaramadze: Same here `return self._updated[cf_name][prefix].get(key, UNDEFINED)`

Variable assignment takes time, let's not do it if not necessary.

quixstreams/state/base/transaction.py
```diff
@@ -25,24 +15,131 @@
...
+
+        :param: cf_name: column family name
+        """
+        return self._updated.get(cf_name, {})
+
+    def get_deletes(self, cf_name: str = "default") -> Set[bytes]:
+        """
+        Get all deleted keys (excluding updated) as a set.
+        """
+        return self._deleted[cf_name]
```
gwaramadze, line 140: Shouldn't this also be `self._deleted.get(cf_name, {})`? Why are you sure that cf_name key is in the dict at this point?

quixstreams/state/base/transaction.py
```diff
@@ -310,21 +409,32 @@ def _prepare(self, processed_offset: int):
...
+
+        for cf_name in column_families:
             headers = {
                 CHANGELOG_CF_MESSAGE_HEADER: cf_name,
                 CHANGELOG_PROCESSED_OFFSET_MESSAGE_HEADER: source_tp_offset_header,
             }
-            for _, prefix_update_cache in cf_update_cache.items():
+
+            updates = self._update_cache.get_updates(cf_name=cf_name)
+            for _, prefix_update_cache in updates.items():
```
gwaramadze: `for prefix_update_cache in updates.values():`

quixstreams/state/rocksdb/partition.py
```diff
@@ -113,32 +112,37 @@ def _recover_from_changelog_message(
...
+
         # Iterate over the transaction update cache
-        for cf_name, cf_update_cache in data.items():
+        column_families = cache.get_column_families()
+        for cf_name in column_families:
             cf_handle = self.get_column_family_handle(cf_name)
-            for _prefix, prefix_update_cache in cf_update_cache.items():
+
+            updates = cache.get_updates(cf_name=cf_name)
+            for _, prefix_update_cache in updates.items():
```
gwaramadze: `for prefix_update_cache in updates.values():`

quixstreams/state/rocksdb/windowed/serialization.py
```diff
@@ -1,15 +1,22 @@
...
     int_to_int64_bytes,
 )
 
 __all__ = ("parse_window_key", "encode_window_key", "encode_window_prefix")
 
 _TIMESTAMP_BYTE_LENGTH = len(int_to_int64_bytes(0))
 _PREFIX_SEPARATOR_LENGTH = len(PREFIX_SEPARATOR)
+_TIMESTAMPS_SEGMENT_LEN = _TIMESTAMP_BYTE_LENGTH * 2 + _PREFIX_SEPARATOR_LENGTH
+
+
```
gwaramadze: nit: remove one empty line

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -203,38 +203,40 @@ def get_windows(
...
+        for key, value in itertools.chain(db_windows, updated_windows):
+            if key not in deleted_windows:
                 merged_windows[key] = value
 
-        final_windows = []
-        for key in sorted(merged_windows, reverse=backwards):
-            _, start, end = parse_window_key(key)
+        # Sort and deserialize windows merged from the cache and store
+        result = []
+        for db_key, db_value in sorted(
```
gwaramadze: merged_windows are no longer tied to either db or cache, so it's best to rename variables to just key and value

quixstreams/state/base/transaction.py
ghost: Because `_deleted` and `_updated` are `defaultdicts`.
If the key doesn't exist, it will be added automatically

quixstreams/state/base/transaction.py
ghost: I wanted to do it this way at first, but to say whether `_updated` is empty or not, you need to iterate over it.
Due to `defaultdict`, `.get()` will create a key on the first prefix level, but the actual key may not be set.

Using a `self._empty` was easier because if there's at least one "delete" or "update", the cache is not empty anymore (even if the updated key was deleted).

gwaramadze (approved):

PR #591, opened 2024-10-24
Calculate window watermarks by individual keys instead of partitions

ghost:
Calculate window watermarks by individual message keys instead of partitions.

This way, windows can capture more data when processing misaligned data streams (e.g., physical devices sending data at different frequencies), and fewer updates will be declared "late".

### DIFF
--- a/quixstreams/state/rocksdb/windowed/metadata.py
+++ b/quixstreams/state/rocksdb/windowed/metadata.py
@@ -2,5 +2,4 @@ LATEST_EXPIRED_WINDOW_TIMESTAMP_KEY = b"__expired_start_gt__"
 
 LATEST_EXPIRED_WINDOW_CF_NAME = "__expiration-index__"
 LATEST_TIMESTAMPS_CF_NAME = "__latest-timestamps__"
-
-LATEST_TIMESTAMP_KEY = b"__topic_latest_timestamp__"
+LATEST_TIMESTAMP_KEY = b"__latest_timestamp__"
--- a/quixstreams/state/rocksdb/windowed/metadata.py
+++ b/quixstreams/state/rocksdb/windowed/metadata.py
@@ -1,5 +1,6 @@
 LATEST_EXPIRED_WINDOW_TIMESTAMP_KEY = b"__expired_start_gt__"
 
 LATEST_EXPIRED_WINDOW_CF_NAME = "__expiration-index__"
+LATEST_TIMESTAMPS_CF_NAME = "__latest-timestamps__"
 
 LATEST_TIMESTAMP_KEY = b"__topic_latest_timestamp__"
--- a/quixstreams/state/rocksdb/windowed/partition.py
+++ b/quixstreams/state/rocksdb/windowed/partition.py
@@ -5,17 +5,14 @@ from rocksdict import RdictItems, ReadOptions, WriteBatch  # type: ignore
 
 from quixstreams.state.base import PartitionTransactionCache
 from quixstreams.state.recovery import ChangelogProducer
-from .metadata import (
-    LATEST_EXPIRED_WINDOW_CF_NAME,
-    LATEST_TIMESTAMPS_CF_NAME,
-)
-from .transaction import WindowedRocksDBPartitionTransaction
-from quixstreams.state.serialization import int_from_int64_bytes, int_to_int64_bytes
 
 from ..exceptions import ColumnFamilyDoesNotExist
 from ..partition import RocksDBStorePartition
 from ..types import RocksDBOptionsType
-from .metadata import LATEST_EXPIRED_WINDOW_CF_NAME, LATEST_TIMESTAMPS_CF_NAME, LATEST_TIMESTAMP_KEY
+from .metadata import (
+    LATEST_EXPIRED_WINDOW_CF_NAME,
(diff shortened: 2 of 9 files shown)


quixstreams/state/rocksdb/windowed/metadata.py
```diff
@@ -1,5 +1,5 @@
 LATEST_EXPIRED_WINDOW_TIMESTAMP_KEY = b"__expired_start_gt__"
 
 LATEST_EXPIRED_WINDOW_CF_NAME = "__expiration-index__"
-
-LATEST_TIMESTAMP_KEY = b"__topic_latest_timestamp__"
+LATEST_TIMESTAMPS_CF_NAME = "__latest-timestamps__"
+LATEST_TIMESTAMP_KEY = b"__latest_timestamp__"
```
gwaramadze, line 5: There no usage of this constant anymore

quixstreams/state/rocksdb/windowed/partition.py
```diff
@@ -81,19 +72,9 @@ def write(
     ):
         batch = WriteBatch(raw_mode=True)
 
-        if latest_timestamp_ms is not None:
```
gwaramadze, line 84: `latest_timestamp_ms` is not used anymore

quixstreams/state/rocksdb/windowed/partition.py
```diff
@@ -81,19 +72,9 @@ def write(
     ):
         batch = WriteBatch(raw_mode=True)
```
gwaramadze: I know it was already like this, but that looks strange to ignore the batch argument and always initialize it from scratch. Could you check if everything is alright there?

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -51,15 +55,18 @@ def as_state(self, prefix: Any = DEFAULT_PREFIX) -> WindowedTransactionState:
...
-                f"Invalid window duration: window end {end_ms} is smaller or equal "
-                f"than window start {start_ms}"
-            )
+        stored_ts = self.get(
+            key=LATEST_TIMESTAMP_KEY,
+            prefix=prefix,
+            cf_name=LATEST_TIMESTAMPS_CF_NAME,
+        )
+        self._latest_timestamps[prefix] = stored_ts
+        return stored_ts or 0
```
gwaramadze: I think we can do `self.get(..., default=0)` here and remove `... or 0`. This way we will end up with 0 in the `_latest_timestamps` cache which is better than useless `None`.

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -240,3 +218,46 @@ def get_windows(
...
+    def _set_latest_timestamp(self, prefix: bytes, timestamp_ms: int):
+        self._latest_timestamps[prefix] = timestamp_ms
+        self.set(
+            key=LATEST_TIMESTAMP_KEY,
+            value=timestamp_ms,
+            prefix=prefix,
+            cf_name=LATEST_TIMESTAMPS_CF_NAME,
+        )
+
+    def _get_last_expired_timestamp(self, prefix: bytes) -> Optional[int]:
```
gwaramadze, line 229: Get and set methods for latest timestamp and last expired are very similar. Did you think about some abstraction here?

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -240,3 +218,46 @@ def get_windows(
...
+        if cached_ts is not None:
+            return cached_ts
+
+        stored_ts = self.get(
+            key=LATEST_EXPIRED_WINDOW_TIMESTAMP_KEY,
+            prefix=prefix,
+            cf_name=LATEST_EXPIRED_WINDOW_CF_NAME,
+        )
+        self._last_expired_timestamps[prefix] = stored_ts
+        return stored_ts
```
gwaramadze, line 240: We don't want to return 0 here like for latest_timestamp?

quixstreams/state/rocksdb/windowed/partition.py
ghost: The whole `write()` method looks like a leftover of the previous implementation.
We don't need to override it anymore

quixstreams/state/rocksdb/windowed/transaction.py
ghost: Nope, not in this case.
The previous code expected `None`, so I don't want to change it.

quixstreams/state/rocksdb/windowed/transaction.py
ghost: They are, but I don't think we need to abstract anything here. They are very simple

quixstreams/state/rocksdb/windowed/partition.py
ghost: Same here https://github.com/quixio/quix-streams/pull/591#discussion_r1825610733

quixstreams/state/rocksdb/windowed/transaction.py
ghost: Agree, updated 👍

gwaramadze (approved):

PR #594, opened 2024-10-24
Chore/quix topics management change

tim-quix:
# NON-Quix APPLICATION BEHAVIOR IS ENTIRELY UNCHANGED, AND THIS IS NON-BREAKING.

All topic and state store names remain the same. 

This has been fully tested in Quix with both Quix and External brokers using all our internal topic naming schemes to confirm no namings have changed.

This changes only how Quix topics are managed internally in the Application and when Quix topics are created. 

Topic validation step remains unchanged.

## Recap of topic naming in Quix Cloud
Quix has two identifiers for each topic: a topic "name" and topic "ID". 

The name is typically what the user will provide (i.e. app.topic("NAME")) 

The ID is the true cluster topic name (and what corresponds to typical topic use for the SDK), which usually has some sort of workspace or other identifier prepended to the "name" since Quix is multi-tenant. 

The user will typically be unaware and/or not need to use the ID version.

## Why our current approach is problematic

We currently manage the name resolution ourselves (including prepending/appending WS id's ourselves as needed) in most cases, which depends on the naming behavior in the platform to correspond to how we handle it. Basically we have redundant logic going on, and we ideally should let the API handle as much of the platform-related behaviors as possible.

## Why does the current approach exist
1. we wanted topic creation behavior steps to be the same regardless of Quix App vs Non-Quix App 
2. the API could not correctly resolve receiving the Topic ID version of a topic (but it now can).

## NEW CHANGES/APPROACH

Essentially, the SDK now fully relies on the API to return the correct topic ID's based on the names the user hands the application with app.topic(<NAME>).

This is accomplished by three main changes: 

1. topics being get/created in a Quix App when app.topic(<NAME>) is called, allowing us to get/use the topic ID immediately and directly throughout the App (all behavior afterward is then is like a non-Quix App).

2. The "name" from the API call is now stored and used for "internal topic" name constructions, which don't want to have the WS name in them multiple times (again, this results in the same topic names we were previously constructing ourselves).

3. During the original topic creation step, we now instead confirm the "Ready" status (originally, we both created AND waited for "Ready" status here).

### DIFF
--- a/quixstreams/platforms/quix/config.py
+++ b/quixstreams/platforms/quix/config.py
@@ -3,12 +3,12 @@ import dataclasses
 import logging
 import time
 from copy import deepcopy
-from typing import List, Optional, Set
+from typing import List, Optional
 
 from requests import HTTPError
 
 from quixstreams.kafka.configuration import ConnectionConfig
-from quixstreams.models.topics import Topic
+from quixstreams.models.topics import Topic, TopicConfig
 
 from .api import QuixPortalApiService
 from .exceptions import (
@@ -159,6 +159,29 @@ class QuixKafkaConfigsBuilder:
             "metadata.max.age.ms": QUIX_METADATA_MAX_AGE_MS,
         }
 
+    @classmethod
+    def convert_topic_response(cls, api_response: dict) -> Topic:
+        """
+        Converts a GET or POST ("create") topic API response to a Topic object
+
+        :param api_response: the dict response from a get or create topic call
+        :return: a corresponding Topic object
+        """
+        topic_config = api_response["configuration"]
+        return Topic(
+            name=api_response["id"],
+            config=TopicConfig(
+                num_partitions=topic_config["partitions"],
+                replication_factor=topic_config["replicationFactor"],
+                extra_config={
+                    "retention.ms": topic_config["retentionInMinutes"] * 60 * 1000,
+                    "retention.bytes": topic_config["retentionInBytes"],
+                    # TODO: uncomment or remove this once Emanuel confirms API behavior
+                    # "cleanup.policy": true_cfg["cleanupPolicy"],
+                },
+            ),
+        )
+
     def strip_workspace_id_prefix(self, s: str) -> str:
         """
         Remove the workspace ID from a given string if it starts with it.
@@ -301,11 +324,12 @@ class QuixKafkaConfigsBuilder:
             ):
                 return ws
 
-    def _create_topic(self, topic: Topic, timeout: Optional[float] = None):
+    def create_topic(self, topic: Topic, timeout: Optional[float] = None):
         """
-        The actual API call to create the topic
+        The actual API call to create the topic.
 
         :param topic: a Topic instance
+        :param timeout: response timeout (seconds); Default 30
         """
         cfg = topic.config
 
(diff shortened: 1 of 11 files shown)


quixstreams/platforms/quix/config.py
```diff
@@ -321,12 +323,43 @@ def _create_topic(self, topic: Topic, timeout: Optional[float] = None):
...
+        marked as "Ready" (and thus ready to produce to/consume from).
+        """
+        try:
+            return self.get_topic(topic_name=topic.name, timeout=timeout)
+        except QuixApiRequestFailure as get_error:
+            if get_error.status_code == 404:
+                try:
+                    return self._create_topic(topic, timeout=timeout)
+                except QuixApiRequestFailure as create_error:
+                    if create_error.status_code == 404:
```
ghost: Are you sure about this ? Create fails with a 404 when the topic already exist ?

quixstreams/platforms/quix/config.py
```diff
@@ -348,9 +381,13 @@ def _finalize_create(
...
             # Each topic seems to take 10-15 seconds each to finalize (at least in dev)
             time.sleep(1)
-            for topic in [
-                t for t in self.get_topics(timeout=timeout) if t["id"] in topics.copy()
-            ]:
+            for topic in (
+                checks := [
+                    t
+                    for t in self.get_topics(timeout=timeout)
+                    if t["id"] in topics.copy()
```
ghost: Small nitpick, this is doing a copy of topics for each `t`

quixstreams/platforms/quix/config.py
```diff
@@ -359,119 +396,61 @@ def _finalize_create(
...
-        self._finalize_create(
-            finalize, timeout=timeout, finalize_timeout=finalize_timeout
+        logger.debug("Confirming all topics are ready in Quix Cloud...")
+        topic_ids = [topic.name for topic in topics]
+        self._confirm_topic_ready_statuses(
+            {
+                t["id"]
+                for t in self.get_topics(timeout=timeout)
+                if t["id"] in topic_ids and t["status"] != "Ready"
+            },
```
ghost: I think it's best to pass all `topics_ids` to `_confirm_topic_ready_statuses`.

`_confirm_topic_ready_statuses` will do the check and return if they are all ready, no need to do a first filtering here.

quixstreams/platforms/quix/config.py
```diff
@@ -359,119 +396,61 @@ def _finalize_create(
...
         if topics:
             raise QuixCreateTopicTimeout(
-                f"Creation succeeded, but waiting for 'Ready' status timed out "
-                f"for topics: {[self.strip_workspace_id_prefix(t) for t in topics]}"
+                f"Waiting for 'Ready' status timed out for Quix "
+                f"topics: {[t['name'] for t in checks if t['id'] in topics]}"
             )
 
-    def create_topics(
+    def confirm_topic_ready_statuses(
```
ghost: I'd prefer a more descriptive method name. This can block for a while. something like `wait_for_topics_ready_statuses`

ghost (changes requested):
I'm surprised no changes are needed for `QuixEnvironmentSource`. If it's still working, I guess we should do a second pass and make the source also rely on the API information.

quixstreams/platforms/quix/config.py
tim-quix: Giving it a second look, it's just a 400 with creation if it already exists (it IS a 404 with GET topic though if it doesn't exist)....pretty sure it returns a 400 with a failure message for anything BUT a successful topic creation (authentication would still be a 403 though).

With that, I opted to no longer inspect the response if it raises an error and just try the next calls regardless. The stack trace will include ALL the exceptions if all attempted calls fail so the information will manifest correctly regardless, and then we're not dependent on inspecting error messages or status codes perfectly.

ghost: I think testing for `404` on the first get make sense. We only want to create on a 404.

Any error on create should be raised, we can't recover and do anything about it.

tim-quix: Well, with creating, it might get a 404 on it because another app tried to create the topic at the same time, which is why I do the final GET attempt. 

I could still do a 404 check on the first get which will catch things like a 403 without making extra unnecc calls

ghost (approved):

ghost (approved):

PR #601, opened 2024-11-04
local file source

tim-quix:
Supports Parquet, Jsonlines with gzip compression, and replaying messages at original produce speed.


quixstreams/sources/community/file/file.py
```diff
@@ -0,0 +1,175 @@
...
+        f'Invalid format name "{formatter}". '
+        f"Allowed values: {allowed_formats}, "
+        f"or an instance of a subclass of `Format`."
+    )
+
+
+def _file_finder(filepath: Path) -> Generator[Path, None, None]:
+    if filepath.is_dir():
+        for i in sorted(filepath.iterdir(), key=lambda x: x.name):
+            print(i)
```
gwaramadze: let's use logger 😉

EDIT: actually I think you meant for this to be removed. I assume this because you don't print filepath if `is_dir` is false

gwaramadze (approved):
LGTM, just one small comment.

gwaramadze (changes requested):
I forgot something. You need to do this:

* include `connectors/sources/file-source.md` under `- Sources:` section in `mkdocs.yml`.
* Include _at least_ `quixstreams.sources.community.file.file` under `"sources.md"` section in `docs/build/build.py`

`FileSink` is merged, rebase first.

quixstreams/sources/community/file/file.py
tim-quix: good catch, thanks =)

gwaramadze (approved):

PR #602, opened 2024-11-05
New version of the CSVSource

ghost:
A new version of CSVSource to make it more flexible:
- No need to pack the message value to a field. Every column is produced to Kafka now
- Added optional "key_extractor" and "timestamp_extractor" callables to specify Kafka keys and timestamps based on the row
- Added optional `delay` param to sleep between the rows for simulation

Breaking change:
- The "name" parameter is required now (previously the filename was used as a name)
- `key_serializer` and `value_serializer` params have been removed.

### DIFF
--- a/quixstreams/sources/core/csv.py
+++ b/quixstreams/sources/core/csv.py
@@ -13,8 +13,8 @@ logger = logging.getLogger(__name__)
 class CSVSource(Source):
     def __init__(
         self,
-        name: str,
         path: Union[str, Path],
+        name: str,
         key_extractor: Optional[Callable[[dict], AnyStr]] = None,
         timestamp_extractor: Optional[Callable[[dict], int]] = None,
         delay: float = 0,
@@ -25,8 +25,9 @@ class CSVSource(Source):
         A base CSV source that reads data from a CSV file and produces rows
         to the Kafka topic in JSON format.
 
-        :param name: The source unique name. Used to generate the topic configuration.
-        :param path: path to the CSV file
+        :param path: a path to the CSV file.
+        :param name: a unique name for the Source.
+            It is used as a part of the default topic name.
         :param key_extractor: an optional callable to extract the message key from the row.
             It must return either `str` or `bytes`.
             If empty, the Kafka messages will be produced without keys.
@@ -35,6 +36,8 @@ class CSVSource(Source):
             It must return time in milliseconds as `int`.
             If empty, the current epoch will be used.
             Default - `None`
+        :param delay: an optional delay after producing each row for stream simulation.
+            Default - `0`.
         :param shutdown_timeout: Time in second the application waits for the source to gracefully shut down.
         :param dialect: a CSV dialect to use. It affects quoting and delimiters.
             See the ["csv" module docs](https://docs.python.org/3/library/csv.html#csv-fmt-params) for more info.
--- a/quixstreams/sources/core/csv.py
+++ b/quixstreams/sources/core/csv.py
@@ -1,74 +1,81 @@
 import csv
-import json
-from typing import Any, Callable, Optional
+import logging
+import time
+from pathlib import Path
+from typing import AnyStr, Callable, Optional, Union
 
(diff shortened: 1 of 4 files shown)


quixstreams/sources/core/csv.py
```diff
@@ -1,74 +1,81 @@
...
+logger = logging.getLogger(__name__)
+
 
 class CSVSource(Source):
     def __init__(
         self,
-        path: str,
-        dialect: str = "excel",
-        name: Optional[str] = None,
+        name: str,
```
gwaramadze: Why name required now?

quixstreams/sources/core/csv.py
```diff
@@ -1,74 +1,81 @@
...
+        path: Union[str, Path],
+        key_extractor: Optional[Callable[[dict], AnyStr]] = None,
+        timestamp_extractor: Optional[Callable[[dict], int]] = None,
+        delay: float = 0,
```
gwaramadze, line 20: Missing in docstring

quixstreams/sources/core/csv.py
```diff
@@ -1,74 +1,81 @@
...
-                    value=msg.value,
-                    timestamp=msg.timestamp,
-                    headers=msg.headers,
-                )
+                # Publish the data to the topic
+                self.produce(timestamp=msg.timestamp, key=msg.key, value=msg.value)
+
+                # If the delay is specified, sleep before producing the next row
+                if self.delay > 0:
+                    time.sleep(self.delay)
```
gwaramadze, line 81: If we know the timestamp of a record, delay may be dynamic. You can reuse what @tim-quix implemented [here](https://github.com/quixio/quix-streams/blob/main/quixstreams/sources/community/file/file.py#L92).

quixstreams/sources/core/csv.py
```diff
@@ -1,74 +1,81 @@
...
+                msg = self.serialize(key=message_key, value=row, timestamp_ms=timestamp)
 
-                self.produce(
-                    key=msg.key,
-                    value=msg.value,
-                    timestamp=msg.timestamp,
-                    headers=msg.headers,
-                )
+                # Publish the data to the topic
+                self.produce(timestamp=msg.timestamp, key=msg.key, value=msg.value)
```
gwaramadze, line 77: Not related to this PR, but would be good if `produce` could just accept `KafkaMessage`...

quixstreams/sources/core/csv.py
```diff
@@ -1,74 +1,81 @@
...
+
 
 class CSVSource(Source):
     def __init__(
         self,
-        path: str,
-        dialect: str = "excel",
-        name: Optional[str] = None,
+        name: str,
+        path: Union[str, Path],
```
gwaramadze, line 16: Reordering arguments is also a breaking change, why `path` can't stay on position 1 and put name on position 2?

quixstreams/sources/core/csv.py
ghost: For consistency and flexibility.

There may be cases when you simply want to change the file name or stream data from a different file while keeping the same topic.
And picking the source name is relatively easy.

Also, file paths are not exactly unique (e.g., `./path.csv`  and `path.csv` can lead to the same file).
And if we use "path" as a name, the relative file paths cannot be used.

quixstreams/sources/core/csv.py
ghost: That's reasonable, yep

quixstreams/sources/core/csv.py
ghost: I'd keep the current delay behavior because it's easy to reason about until `FileSource` receives more testing.

quixstreams/sources/core/csv.py
ghost: That would reduce the boilerplate, yes.
Maybe we should simply do the serialization inside `produce()`

gwaramadze (approved):

PR #615, opened 2024-11-07
Stateful sources implementation

ghost:
Enhance source connectors by allowing access to a kafka backed state

* Implement stateful sources
* Added support for creating changelog topics without a related "data" topic
  
  Stateful sources don't have an input topic on which to associate the changelog topic. The topic and recovery manager need to support a changelog topic with a topic set to `None`.

* Added support for custom changelog topic configuration (instead of relying on the "data" topic configuration)

  When creating the changelog topic we use a configuration similar to the data topic configuration (replication factor, number of partitions, ...). As the data topic is in-existant for stateful sources we implement a way to create a changelog topic with a passed configuration.

### DIFF
--- a/quixstreams/sources/base/manager.py
+++ b/quixstreams/sources/base/manager.py
@@ -12,7 +12,7 @@ from quixstreams.state.memory import MemoryStore
 
 from .exceptions import SourceException
 from .multiprocessing import multiprocessing
-from .source import BaseSource, StatefullSource
+from .source import BaseSource, StatefulSource
 
 logger = logging.getLogger(__name__)
 
@@ -76,7 +76,7 @@ class SourceProcess(multiprocessing.Process):
 
         configuration = {"topic": self.topic, "producer": self._producer}
 
-        if isinstance(self.source, StatefullSource):
+        if isinstance(self.source, StatefulSource):
             try:
                 configuration["store_partition"] = self._recover_state(self.source)
             except BaseException as err:
@@ -102,7 +102,7 @@ class SourceProcess(multiprocessing.Process):
             "s" if threadcount > 1 else "",
         )
 
-    def _recover_state(self, source: StatefullSource) -> StorePartition:
+    def _recover_state(self, source: StatefulSource) -> StorePartition:
         """
         Recover the state from the changelog topic and return the assigned partition
 
--- a/quixstreams/sources/base/manager.py
+++ b/quixstreams/sources/base/manager.py
@@ -5,8 +5,10 @@ from pickle import PicklingError
 from typing import List
 
 from quixstreams.logging import LOGGER_NAME, configure_logging
-from quixstreams.models import Topic
+from quixstreams.models import Topic, TopicAdmin
 from quixstreams.models.topics import TopicConfig
+from quixstreams.rowconsumer import RowConsumer
+from quixstreams.rowproducer import RowProducer
 from quixstreams.state import RecoveryManager, StateStoreManager, StorePartition
 from quixstreams.state.memory import MemoryStore
 
@@ -27,10 +29,17 @@ class SourceProcess(multiprocessing.Process):
     Some methods are designed to be used from the parent process, and others from the child process.
     """
 
-    def __init__(self, source, topic, producer, consumer, topic_manager):
+    def __init__(
+        self,
+        source: BaseSource,
+        topic: Topic,
+        producer: RowProducer,
+        consumer: RowConsumer,
+        topic_manager: TopicAdmin,
+    ):
         super().__init__()
         self.topic = topic
-        self.source: BaseSource = source
+        self.source = source
 
         self._exceptions: List[Exception] = []
         self._started = False
@@ -48,7 +57,7 @@ class SourceProcess(multiprocessing.Process):
         self._rpipe, self._wpipe = multiprocessing.Pipe(duplex=False)
 
     @property
-    def started(self):
+    def started(self) -> bool:
         return self._started
 
     # --- CHILD PROCESS METHODS --- #
@@ -124,7 +133,7 @@ class SourceProcess(multiprocessing.Process):
             store_type=MemoryStore,
             topic_config=TopicConfig(
                 num_partitions=source.store_partitions_count,
-                replication_factor=1,
+                replication_factor=self._topic_manager.default_replication_factor,
             ),
         )
 
@@ -240,7 +249,7 @@ class SourceManager:
         producer,
         consumer,
         topic_manager,
-    ):
+    ) -> SourceProcess:
         """
         Register a new source in the manager.
 
--- a/quixstreams/sources/base/manager.py
+++ b/quixstreams/sources/base/manager.py
@@ -4,8 +4,10 @@ import threading
 from pickle import PicklingError
 from typing import List
 
+from confluent_kafka import OFFSET_BEGINNING
+
 from quixstreams.logging import LOGGER_NAME, configure_logging
-from quixstreams.models import Topic, TopicAdmin
+from quixstreams.models import Topic, TopicManager
 from quixstreams.models.topics import TopicConfig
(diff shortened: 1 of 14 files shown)


quixstreams/sources/base/source.py
```diff
@@ -320,3 +314,140 @@ def default_topic(self) -> Topic:
...
+        self._store_partition: Optional[StorePartition] = None
+        self._store_transaction: Optional[PartitionTransaction] = None
+
+    def configure(
+        self,
+        topic: Topic,
+        producer: RowProducer,
+        *,
+        store_partition: StorePartition,
+        **kwargs,
```
ghost: The type checker yells at this method:
```Signature of method 'StatefulSource. configure()' does not match signature of the base method in class 'BaseSource' ```

File: quixstreams/sources/base/source.py [hunk 2]
```diff
@@ -320,3 +314,140 @@ def default_topic(self) -> Topic:
...
+
+        It ensures the state is published to the changelog topic and all messages are successfully delivered to Kafka.
+
+        :param float timeout: time to attempt flushing (seconds).
+            None use producer default or -1 is infinite. Default: None
+
+        :raises CheckpointProducerTimeout: if any message fails to produce before the timeout
+        """
+        if self._store_transaction:
+            self._store_transaction.prepare(None)
```
ghost, line 452: This method expects `int` only

ghost:
I propose changing how the State object is accessed in the StatefulSource.

Whoever implements the `StatefulSource` interface would need to be aware of the object's lifecycle returned by `self.state()` and re-create it after each `flush()`. 
If you forget to do that, it will most likely crash with some internal `StateTransactionError` that is not a part of the public API.

Instead, we can wrap the `State` object in some proxy class and update the state of this proxy when the transaction is flushed.

For example:
```python
class StateProxy(State):
    def __init__(self, real_state: State):
        self._real_state = real_state

    def get(self, key: Any, default: Any = None) -> Optional[Any]:
        return self._real_state.get(...)

    def set(self, key: Any, value: Any):
        return self._real_state.set(...)

    def delete(self, key: Any): ...

    def exists(self, key: Any) -> bool: ...

    def __replace_real_state__(self, real_state: State):
        # This method is called by StatefulSource and replaces the State object with
        # a new one after Transaction is committed.
        self._real_state = real_state
```

This `StateProxy` will always point to the correct `State` instance.
We can also make `StatefulSource.state()` a property returning this `StateProxy` to make it easier to access.

tim-quix:
To add to Daniil's thoughts, I basically wrote something along a similar line for `KinesisSource` to handle that workflow...so we prob do need something to streamline it:


```python
class SourceCheckpointer():
    def __init__(self, stateful_source: StatefulSource, commit_interval: float = 5.0):
        self._source = stateful_source
        self._last_committed_at = time.monotonic()
        self._commit_interval = commit_interval
        self._state: Optional[State] = None

    @property
    def last_committed_at(self) -> float:
        return self._last_committed_at

    def begin(self):
        if not self._state:
            self._state = self._source.state()

    def get(self, key: str) -> Optional[str]:
        return self._state.get(key)

    def set(self, key: str, value: str):
        self._state.set(key, value)

    def commit(self, force: bool = False):
        if (
            (now := time.monotonic()) - self._last_committed_at > self._commit_interval
        ) or force:
            self._source.flush()
            self._last_committed_at = now
            self._state = None
```

quixstreams/sources/base/source.py
ghost: It's not complaining here. Can you share the full message ? Might be easier to address it in the mypy MR

ghost: I get this message in Pycharm,

The `store_partition` argument is required in `StatefulSource.configure`, which makes `BaseSource.configure` and `StatefulSource.configure` incompatible

quixstreams/sources/base/manager.py
```diff
@@ -24,22 +29,35 @@ class SourceProcess(multiprocessing.Process):
...
     """
 
-    def __init__(self, source):
+    def __init__(
+        self,
+        source: BaseSource,
+        topic: Topic,
+        producer: RowProducer,
+        consumer: RowConsumer,
+        topic_manager: TopicAdmin,
```
ghost: This should be `topic_manager: TopicManager`

quixstreams/sources/base/manager.py
```diff
@@ -170,20 +242,31 @@ class SourceManager:
...
         Register a new source in the manager.
 
         Each source need to already be configured, can't reuse a topic and must be unique
         """
-        if not source.configured:
-            raise ValueError("Accepts configured Source only")
-        if source.producer_topic in self.topics:
-            raise ValueError(f"topic '{source.producer_topic.name}' already in use")
+        if topic in self.topics:
+            raise ValueError(f"topic '{topic.name}' already in use")
```
ghost:
```suggestion
            raise ValueError(f'Topic name "{topic.name}" is already in use')
```

quixstreams/sources/base/manager.py
```diff
@@ -170,20 +242,31 @@ class SourceManager:
...
         elif source in self.sources:
             raise ValueError(f"source '{source}' already registered")
```
ghost:
```suggestion
            raise ValueError(f'Source "{source}" is already registered')
```

quixstreams/sources/base/manager.py
```diff
@@ -79,6 +111,46 @@ def run(self) -> None:
...
+        recovery_manager = RecoveryManager(
+            consumer=self._consumer,
+            topic_manager=self._topic_manager,
+        )
+
+        state_manager = StateStoreManager(
+            producer=self._producer, recovery_manager=recovery_manager
+        )
+
+        store_name = f"source-{source.store_name}"
```
ghost: Can this code be moved to the `StatefulSource.store_name` property?
Otherwise the property doesn't really return a store name but only a part of it

quixstreams/sources/base/manager.py
```diff
@@ -79,6 +111,46 @@ def run(self) -> None:
...
+            ),
+        )
+
+        self._topic_manager.create_all_topics()
+        self._topic_manager.validate_all_topics()
+
+        store_partitions = state_manager.on_partition_assign(
+            topic=None,
+            partition=source.assigned_store_partition,
+            committed_offset=0,
```
ghost: Why `committed_offset` is `0` here?
`0` means that there's one processed message

quixstreams/sources/base/source.py
```diff
@@ -320,3 +314,144 @@ def default_topic(self) -> Topic:
...
+    """
+
+    def __init__(self, name: str, shutdown_timeout: float = 10) -> None:
+        """
+        :param name: The source unique name. It is used to generate the topic configuration.
+        :param shutdown_timeout: Time in second the application waits for the source to gracefully shutdown.
+        """
+        super().__init__(name, shutdown_timeout)
+        self._store_partition: Optional[StorePartition] = None
+        self._store_transaction: Optional[PartitionTransaction] = None
```
ghost, line 376:
```suggestion
        self._store_transaction: Optional[PartitionTransaction] = None
        self._store_state:Optional[State] = None
```

quixstreams/sources/base/source.py
```diff
@@ -320,3 +314,144 @@ def default_topic(self) -> Topic:
...
+        """
+        return self.name
+
+    @property
+    def state(self) -> State:
+        """
+        Create an instance of the `State` protocol that can be used by the source.
+
+        The instance is only valid as long as the transaction is open. After a `flush` a new state
+        MUST be created.
```
ghost:
```suggestion
Get an instance of the `State` to access the state inside the source.
The instance is cached until the next `.flush()` call.

Important: after each `.flush()` call, the returned instance is invalidated and cannot be used.  
This property must be called again after `.flush()`.
```

ghost (approved):

PR #616, opened 2024-11-07
New Sink: GCP Pub/Sub

gwaramadze:


quixstreams/sinks/community/pubsub.py
```diff
@@ -0,0 +1,79 @@
...
+            _timestamp=str(timestamp),
+            **dict(headers),  # non-unique header keys are overwritten by the last one
+        )
+        self._futures[(topic, partition)].append(future)
+
+    def flush(self, topic: str, partition: int) -> None:
+        """
+        Wait for all publish operations to complete.
+        """
+        if futures := self._futures.pop((topic, partition), None):
```
ghost, line 118: Here, we need to make sure that the futures are removed from the dict* if committing this TP fails

quixstreams/sinks/community/pubsub.py
```diff
@@ -0,0 +1,79 @@
...
+
+from quixstreams.models.types import HeaderValue
+from quixstreams.sinks.base import BaseSink
+
+logger = logging.getLogger(__name__)
+
+TopicPartition = tuple[str, int]
+
+
+class PubSubSinkTopicNotFoundError(Exception):
```
ghost: Do we need the word `Sink` here?
```suggestion
class PubSubTopicNotFoundError(Exception):
```

quixstreams/sinks/community/pubsub.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, project_id: str, topic_id: str, **kwargs) -> None:
+        """
+        Initialize the PubSubSink.
+
+        :param project_id: GCP project ID
+        :param topic_id: Pub/Sub topic ID
+        :param kwargs: Additional keyword arguments passed to PublisherClient
+            Most notably, `credentials` can be used to specify credentials to use
+            for authentication.
```
ghost: Can we pull the most common credentials from `kwargs` to parameters?

quixstreams/sinks/community/pubsub.py
```diff
@@ -0,0 +1,79 @@
...
+        headers: list[tuple[str, HeaderValue]],
+        topic: str,
+        partition: int,
+        offset: int,
+    ) -> None:
+        """
+        Publish a message to Pub/Sub.
+        """
+        data = value if isinstance(value, bytes) else str(value).encode()
+        key = key if isinstance(key, bytes) else str(key)
```
ghost: I think we need to have `value_serializer` and `key_serializer` here for non-bytes

quixstreams/sinks/community/pubsub.py
gwaramadze: That is a `.pop` operation, so we're good. Is there a gotcha I don't see? 🤔

quixstreams/sinks/community/pubsub.py
gwaramadze: Actually, no, I will remove this highlight from the documentation and leave it to the user. I thought that credentials expected here is a str type (path to service account JSON file) but it is `google.auth.credentials.Credentials`. It's optional and if not provided, the client will take the path from `GOOGLE_APPLICATION_CREDENTIALS` env var.

quixstreams/sinks/community/pubsub.py
```diff
@@ -0,0 +1,126 @@
...
+        flush_timeout: int = 5,
+        **kwargs,
+    ) -> None:
+        """
+        Initialize the PubSubSink.
+
+        :param project_id: GCP project ID.
+        :param topic_id: Pub/Sub topic ID.
+        :param service_account_json: a JSON string with service account credentials
+            to connect to Pub/Sub.
```
ghost:
```suggestion
        :param service_account_json: an optional JSON string with service account credentials
            to connect to Pub/Sub.
            The internal `PublisherClient` will use the Application Default Credentials if not provided.
            See https://cloud.google.com/docs/authentication/provide-credentials-adc for more info.
            Default - `None`.
```

ghost (dismissed):

ghost (approved):

PR #621, opened 2024-11-12
New Sink: BigQuery

ghost:
The new BigQuery Sink.

## How to install 
```
pip install quixstreams[bigquery]
```

## Things to know
1. The data is ingested using `google-cloud-bigquery`, which uses REST API under the hood
2. The sink creates a dataset and a table if it doesn't exist.
3. It supports only `dicts`
4. It adds new columns on the fly with the types inferred from the values' Python types. Most of the widely used types are supported (ints, bytes, datetimes, lists, dicts ,etc.).   
It can be turned off if `schema_auto_update=False` is passed.
The mapping can be found in the `_BIGQUERY_TYPES_MAP` variable.
5. The underlying `google-cloud-bigquery` transforms some date types before sending them. For example, `bytes` values are encoded as base64.
6. The backpressure support has not been implemented in this PR yet.

## How to use
```python
# Create an Application.
from quixstreams.sinks.community.bigquery import BigQuerySink

app = Application(
    auto_offset_reset="earliest",
    consumer_group="consumer-group",
    loglevel="DEBUG",
)

topic = app.topic("topic-name")

# Read service account credentials from the JSON file
with open("service-account.json", "r") as f:
    service_account_json = f.read()

# Initialize a sink
bigquery_sink = BigQuerySink(
    project_id="<project ID>",
    location="<location>",
    dataset_id="<dataset ID>",
    table_name="<table name>",
    service_account_json=service_account_json,
)

sdf = app.dataframe(topic)
sdf.sink(bigquery_sink)
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -50,6 +50,7 @@ nav:
     -  Sinks:
         - 'connectors/sinks/README.md'
         - Apache Iceberg Sink: connectors/sinks/apache-iceberg-sink.md
+        - BigQuery Sink: connectors/sinks/bigquery-sink.md
         - CSV Sink: connectors/sinks/csv-sink.md
         - File Sink: connectors/sinks/file-sink.md
         - Google Cloud Pub/Sub Sink: connectors/sinks/google-cloud-pubsub-sink.md
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -52,7 +52,7 @@ nav:
         - Apache Iceberg Sink: connectors/sinks/apache-iceberg-sink.md
         - CSV Sink: connectors/sinks/csv-sink.md
         - File Sink: connectors/sinks/file-sink.md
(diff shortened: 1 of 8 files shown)


ghost (dismissed):

mkdocs.yml
```diff
@@ -53,6 +53,7 @@ nav:
         - CSV Sink: connectors/sinks/csv-sink.md
         - InfluxDB v3 Sink: connectors/sinks/influxdb3-sink.md
         - File Sink: connectors/sinks/file-sink.md
+        - Google BigQuery Sink: connectors/sinks/bigquery-sink.md
```
gwaramadze: maybe `GCP BigQuery Sink`?

gwaramadze (dismissed):

mkdocs.yml
gwaramadze: obsolete comment

gwaramadze (approved):

PR #643, opened 2024-11-20
CI: Implement mypy pre-commit check

ghost:
* Add mypy check in pre-commit
* Fix typing in various places
* Split consumer into `BaseConsumer`/`Consumer`/`RowConsumer` to support accepting different topic type (`str` and `Topic`) in `subscribe`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install requirements
         run: |
           python -m pip install -U pip
-        python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
+          python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
       - uses: pre-commit/action@v3.0.1
 
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,12 +25,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.12
-      - name: Update pip
-        run: |
-          python -m pip install -U pip
       - name: Install requirements
         run: |
-          python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
+          python -m pip install -U pip
+        python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
       - uses: pre-commit/action@v3.0.1
 
   test:
@@ -50,11 +48,9 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
-      - name: Update pip
-        run: |
-          python -m pip install -U pip
       - name: Install requirements
         run: |
+          python -m pip install -U pip
           python -m pip install -U -r requirements.txt -r tests/requirements.txt
       - name: Run tests
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,13 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.12
+      - name: Update pip
+        run: |
+          python -m pip install -U pip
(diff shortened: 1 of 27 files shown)


quixstreams/app.py
```diff
@@ -1009,8 +1023,8 @@ def copy(self, **kwargs) -> Self:
         return copy
 
     @property
-    def flush_timeout(self) -> float:
-        return (
+    def flush_timeout(self) -> int:
+        return int(
```
ghost: I'm unsure if everything should be int or float. Currently most timeout expect an int

requirements-mypy.txt
```diff
@@ -0,0 +1,5 @@
+mypy==1.13.0
```
gwaramadze, line 1: Shouldn't mypy be installed by pre-commit?

ghost: We need pre-commit to use the system environment for mypy. Mypy needs to run on a environment with all the dependencies installed. There is no good way to have a pre-commit environment that does that, using the system environment is easier (system mean the current active environment outside of pre-commit not the system base python install)

.github/workflows/ci.yml
```diff
@@ -24,7 +24,13 @@ jobs:
...
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.12
+      - name: Update pip
+        run: |
+          python -m pip install -U pip
+      - name: Install requirements
+        run: |
+          python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
```
gwaramadze: I think we can zip both steps together

```suggestion
      - name: Install requirements
        run: |
          python -m pip install -U pip
          python -m pip install -U -r requirements.txt -r tests/requirements.txt -r requirements-mypy.txt
```

gwaramadze: If you choose to do it, you can do the same for `test` job.

requirements-mypy.txt
gwaramadze: Okay, I see that you use `language: system` so we need to install it outside of pre-commit env

.pre-commit-config.yaml
```diff
@@ -11,3 +11,10 @@ repos:
         entry: python conda/requirements.py
         language: python
         files: ^(requirements\.txt|pyproject\.toml)$
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.13.0
+    hooks:
+    -   id: mypy
+        args: []
```
gwaramadze, line 18: I think this may be removed
```suggestion
```

pyproject.toml
```diff
@@ -96,3 +96,39 @@ log_cli_level = "INFO"
...
+[[tool.mypy.overrides]]
+module = "quixstreams.sources.community.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "quixstreams.state.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "quixstreams.rowproducer.*"
```
gwaramadze: This should do

```suggestion
module = [
    "quixstreams.core.*",
    "quixstreams.dataframe.*",
    "quixstreams.models.*",
    "quixstreams.platforms.*",
    "quixstreams.sinks.community.*",
    "quixstreams.sources.community.*",
    "quixstreams.state.*",
    "quixstreams.rowproducer.*"
]
```

gwaramadze: So many ignored 🙈
Too much work to fix those?

.pre-commit-config.yaml
ghost: It's needed to override the one [set in the recipe repo](https://github.com/pre-commit/mirrors-mypy/blob/main/.pre-commit-hooks.yaml#L7C5-L7C64)

I prefer to configure mypy using pyproject.toml so we can easily run it manually

pyproject.toml
ghost: Too much work for this PR. The goal is to fix them at one point.

gwaramadze (dismissed):

gwaramadze (approved):

PR #649, opened 2024-11-21
Pass AWS credentials parameters to KinesisSink

gwaramadze:


tim-quix:
What if we just defaulted the KinesisSource/Sink cred arguments to os.getenv("AWS boto var name")? Any downsides besides Amazon changing them later?

gwaramadze:
> What if we just defaulted the KinesisSource/Sink cred arguments to os.getenv("AWS boto var name")? Any downsides besides Amazon changing them later?

I think boto will do that implicitly, we don't need to.

ghost (approved):

PR #651, opened 2024-11-25
Prefix names for auto-generated source topics

ghost:
Added the `source__` prefix to the automatically generated Source topics for better visibility across the cluster.

The prefix is added within the `Application.add_source` method to ensure it's always present.
I also considered prefixing at the `Source` level, but the implementation can override it.


quixstreams/app.py
```diff
@@ -665,7 +665,13 @@ def add_source(self, source: BaseSource, topic: Optional[Topic] = None) -> Topic
             Default: the source default
         """
         if not topic:
-            topic = self._topic_manager.register(source.default_topic())
+            # Prefix the name of the default topic generated by the Source
+            # for visibility across all the topics.
+            default_topic = source.default_topic()
+            default_topic = default_topic.__clone__(
```
gwaramadze, line 671: why clone? is Topic immutable?

ghost: Technically, it's not immutable, but it seems cleaner overall. It's done similarly in `QuixTopicManager._finalize_topic`.

gwaramadze (approved):

PR #653, opened 2024-11-26
Source: files (s3 and local)

tim-quix:
Not 100% done yet but here is a fairly refined draft. Just a little more cleanup to do and a tiny bit more work left on Azure, and then documentation assuming we like the approach.

Currently extends `FileSource` (which itself had some minor adjustments to better accommodate extension).


docs/connectors/sources/aws-s3-file-source.md
```diff
@@ -0,0 +1,175 @@
...
+The resulting messages can be produced in "replay" mode, where the time between record 
+producing is matched as close as possible to the original. (per topic partition only).
+
+
+## How To Install
+
+Install Quix Streams with the following optional dependencies:
+
+```bash
+pip install quixstreams[aws]
```
gwaramadze: We kept separation, so you can use [s3]
```suggestion
pip install quixstreams[s3]
```

quixstreams/sources/community/file/origins/s3.py
```diff
@@ -0,0 +1,84 @@
...
+        f"Package {exc.name} is missing: "
+        'run "pip install quixstreams[aws]" to use S3FileOrigin'
+    ) from exc
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("S3FileOrigin",)
+
+
+class S3FileOrigin(FileOrigin):
```
gwaramadze: Maybe just `S3Origin(Origin)`.

quixstreams/sources/community/file/origins/local.py
```diff
@@ -0,0 +1,36 @@
...
+                yield from self.file_collector(i)
+        else:
+            yield filepath
+
+    def get_folder_count(self, folder: Path) -> int:
+        return len([f for f in folder.iterdir()])
+
+    def get_raw_file_stream(self, filepath: Path) -> BytesIO:
+        with open(filepath, "rb") as f:
+            return BytesIO(f.read())
```
gwaramadze: I believe this should work, but it's hard to be 100% confident without tests 😅 

```suggestion
        return BytesIO(filepath.read_bytes())
```

quixstreams/sources/community/file/origins/local.py
```diff
@@ -0,0 +1,36 @@
...
+        self.root_location = "/"
+
+    @property
+    def client(self):
+        return
+
+    def file_collector(
+        self, filepath: Optional[Path] = None
+    ) -> Generator[Path, None, None]:
+        if filepath.is_dir():
```
gwaramadze, line 12: Type hint says `filepath` may be None at this point.

quixstreams/sources/community/file/origins/local.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(
+        self,
+    ):
+        self._client = None
+        self._credentials = {}
+        self.root_location = "/"
+
+    @property
+    def client(self):
+        return
```
gwaramadze: This and `self._client = None` and both unused. Seems redundant 😅

What I'm trying to say is that maybe base class doesn't need to be aware that some implementations _may_ need a client?

quixstreams/sources/community/file/origins/s3.py
```diff
@@ -0,0 +1,84 @@
...
+        data = self.client.get_object(Bucket=self.root_location, Key=str(filepath))[
+            "Body"
+        ].read()
+        return BytesIO(data)
+
+    def get_folder_count(self, folder: Path) -> int:
+        resp = self.client.list_objects(
+            Bucket=self.root_location, Prefix=str(folder), Delimiter="/"
+        )
+        self._client = None
```
gwaramadze: This is very strange. Mistake?

quixstreams/sources/community/file/origins/base.py
```diff
@@ -0,0 +1,53 @@
...
+    Provides methods for navigating folders and retrieving/opening raw files.
+    """
+
+    _client: Any
+    _credentials: Union[dict, str]
+    root_location: Union[str, Path]
+
+    @property
+    @abstractmethod
+    def client(self): ...
```
gwaramadze: Do we need this property and does it have to be _public_? Maybe we can get away with only `_client` attribute and on class implementations. The base class does not need to be aware of all the additional shit its subclasses might need to do its job :)

quixstreams/sources/community/file/origins/s3.py
tim-quix: Unfortunately no. For some reason in was running into "S3 client is unpicklable" errors when I had it initialized prior to running it...but I need to query for the number of files to get partition count. 

Not sure the way around it other than user says how many partitions there are, or maybe make a separate client to query that with (i think it only fails when a query is made before it's run? I'd have to double check)

quixstreams/sources/community/file/origins/base.py
tim-quix: I definitely see your argument and considered similar...though I figured that it seems likely that only "local" will not use one? 

As such, I concluded it seemed more reasonable to account for the majority rather than a special case. Wdyt?

quixstreams/sources/community/file/origins/local.py
tim-quix: Works on my machine 🤣

quixstreams/sources/community/file/origins/s3.py
gwaramadze: So weird. Let's drop a comment then 😉

quixstreams/sources/community/file/origins/base.py
gwaramadze: I don't fully agree but it's not a blocker either 😉 So feel free to resolve or you can ask Daniil what he thinks.

One alternative would be to introduce an intermediate class like `ExternalOrigin(Origin)` and then `S3Origin(ExternalOrigin)`. I didn't think about it before, I could do the same thing for destinations 🤔

quixstreams/sources/community/file/origins/local.py
gwaramadze: (this is being discussed in another thread)

quixstreams/sources/community/file/origins/base.py
tim-quix: Another thought that crossed my mind 🙂 wasnt trying to over engineer it since the only thing different is client, but it would be a small change. Seems like a good middle ground 👍

docs/connectors/sources/aws-s3-file-source.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Configuration
+
+Here are some important configurations to be aware of (see [File Source API](../../api-reference/sources.md#filesource) for all parameters).
+
+### Required:
+
+`S3FileOrigin`:
+
+- `aws_s3_bucket`: The S3 URI with bucket name only (ex: 's3://your-bucket').
```
gwaramadze: Lets lose the `s3://` prefix. User should pass only the bucket name.

quixstreams/sources/community/file/origins/s3.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("S3FileOrigin",)
+
+
+class S3FileOrigin(FileOrigin):
+    def __init__(
+        self,
+        aws_s3_bucket: str,
```
gwaramadze: Maybe just `bucket`? :)

quixstreams/sources/community/file/origins/azure.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    def client(self):
+        if not self._client:
+            blob_client = BlobServiceClient.from_connection_string(self._credentials)
+            container_client = blob_client.get_container_client(self.root_location)
+            self._client: ContainerClient = container_client
+        return self._client
+
+    def file_collector(self, folder: Path) -> Generator[str, None, None]:
```
gwaramadze: Abstract method return type hints at `Iterable[Path]` but this method yields strings.

quixstreams/sources/community/file/origins/s3.py
tim-quix: Changed the approach (and added a comment to clarify the behavior). Let me know what you think!

docs/connectors/sources/aws-s3-file-source.md
tim-quix: Didn't actually need it, I was just a dum dum and copied it from somewhere haha

quixstreams/sources/community/file/origins/base.py
tim-quix: Changed the approach, feel free to take a look =)

tim-quix:
TODO: 
- maybe add a simple file download queue?
- look over Tomas's updated replay timestamp logic and see if changes are needed
- double-check non-code stuff (adding doc paths, package stuff, etc)

docs/connectors/sources/azure-file-source.md
```diff
@@ -0,0 +1,205 @@
...
+
+`AzureOrigin`:
+
+- `connection_string`: Azure client authentication string.
+- `container`: Azure container name.
+
+
+`FileSource`:
+
+- `filepath`: a filepath to recursively read through (exclude bucket name).
```
gwaramadze: "bucket name" or "container"?

quixstreams/sources/community/file/origins/azure.py
```diff
@@ -0,0 +1,59 @@
...
+except ImportError as exc:
+    raise ImportError(
+        f"Package {exc.name} is missing: "
+        'run "pip install quixstreams[azure]" to use AzureOrigin'
+    ) from exc
+
+__all__ = ("AzureOrigin",)
+
+
+class AzureOrigin(ExternalOrigin):
```
gwaramadze: The official name of this service is "Files". You can't go more generic than that 🤣, I think `AzureFilesOrigin` might be better in this case.
```suggestion
class AzureFilesOrigin(ExternalOrigin):
```

quixstreams/sources/community/file/origins/base.py
```diff
@@ -0,0 +1,48 @@
...
+    Result should be ready for deserialization (and/or decompression).
+    """
+
+
+@dataclass
+class ExternalOrigin(Origin, ABC):
+    """An interface for interacting with an external file-based client"""
+
+    _client: Any
+    root_location: Union[str, Path]
```
gwaramadze: I think it is rather confusing for developers, because subclasses are overriding dataclass default `__init__` method. WDYT about this?

```suggestion
class ExternalOrigin(Origin, ABC):
    """An interface for interacting with an external file-based client"""

    _client: Any = NotImplemented
    root_location: Union[str, Path] = NotImplemented
```

quixstreams/sources/community/file/origins/local.py
```diff
@@ -0,0 +1,33 @@
...
+from .base import Origin
+
+__all__ = ("LocalOrigin",)
+
+
+class LocalOrigin(Origin):
+    def __init__(
+        self,
+    ):
+        self._client = None
```
gwaramadze: No need for client here anymore, right?

quixstreams/sources/community/file/origins/s3.py
```diff
@@ -0,0 +1,84 @@
...
+            "aws_secret_access_key": aws_secret_access_key,
+            "endpoint_url": aws_endpoint_url,
+        }
+        # S3 client runs into pickling errors with multiprocessing. We can't set it
+        # until multiprocessing starts it.
+        # We can work around it by setting it during file collection
+        self._client: Optional[S3Client] = None
+
+    def _get_client(self) -> S3Client:
+        return boto_client("s3", **self._credentials)
```
gwaramadze, line 59: I think it should be `self._get_client` too.

BTW, maybe you could replace the `_get_client` with a `_client` property and a good explanation as to why it reinstantiates a client every time. Then `self._client: Optional[S3Client] = None` may be removed from init.

```suggestion
    @property
    def _client(self) -> S3Client:
        return boto_client("s3", **self._credentials)
```

quixstreams/sources/community/file/origins/base.py
gwaramadze: Dropped some more comments in this area so lets close here

quixstreams/sources/community/file/origins/s3.py
gwaramadze: Converging the `client` discussion on a new comment. Closing this thread.

quixstreams/sources/community/file/file.py
```diff
@@ -63,7 +79,8 @@ class FileSource(Source):
     def __init__(
         self,
         filepath: Union[str, Path],
-        file_format: Union[Format, FormatName],
+        file_format: Union[Format, FormatName] = "json",
+        file_origin: Origin = LocalOrigin(),
         file_compression: Optional[CompressionName] = None,
```
ghost: Hey, a suggestion here: since we're changing the API, let's keep the names similar to `FileSink`.

- `filepath` -> `directory`
- `file_format` -> `format`
- `file_origin` -> `origin`
- `file_compression` -> `compression`

quixstreams/sources/community/file/origins/base.py
ghost: TBH, I don't think we even need this `ExternalOrigin` at all. 
All of this can be a part of individual implementations

quixstreams/sources/community/file/origins/s3.py
tim-quix: Well, it doesn't need to instantiate one every time...it basically must be instantiated _after_ it's kicked off as a `multiprocess` Process via `Source.run()` (which is where the pickling error is occurring).

The only reason this is a problem at all is I need to count the partition folders to see what the expected partition count is, which occurs before actually performing the run operation, so that's the thing I have to work around (by not storing that instance of the client when I do that call). 

Ideally I'd like to not have to instantiate a new client every time I call it...currently I'm setting the client on the first call I know that is made after `.run()` and then I can just re-use it as originally desired.

quixstreams/sources/community/file/origins/base.py
tim-quix: Went ahead and removed `ExternalOrigin` per Daniil's suggestion.

quixstreams/sources/community/file/origins/local.py
tim-quix: Correct, removed all the extra stuff, forgot to remove it =)

tim-quix:
This is ready to merge, depending on final thoughts on the client stuff for S3.

gwaramadze (approved):

PR #657, opened 2024-11-27
mypy: make quixstreams.state.* pass type checks

ghost:
mypy: make quixstreams.state.* pass type checks
    
    * Rework message headers type
    * state.partition: remove default from get
    * various type fixes and rework

### DIFF
--- a/quixstreams/models/rows.py
+++ b/quixstreams/models/rows.py
@@ -1,7 +1,7 @@
 from typing import Any, Optional
 
 from .messagecontext import MessageContext
-from .types import MessageHeadersTuples
+from .types import KafkaHeaders
 
 
 class Row:
@@ -19,7 +19,7 @@ class Row:
         key: Optional[Any],
         timestamp: int,
         context: MessageContext,
-        headers: Optional[MessageHeadersTuples] = None,
+        headers: KafkaHeaders = None,
     ):
         self.value = value
         self.key = key
--- a/quixstreams/models/types.py
+++ b/quixstreams/models/types.py
@@ -1,13 +1,16 @@
-from typing import Dict, List, Optional, Tuple, Union
+from typing import List, Mapping, Optional, Sequence, Tuple, Union
 
 from typing_extensions import Protocol
 
-MessageKey = Union[str, bytes]
+MessageKey = Optional[Union[str, bytes]]
 MessageValue = Union[str, bytes]
-HeaderValue = Optional[Union[str, bytes]]
-MessageHeadersTuples = List[Tuple[str, HeaderValue]]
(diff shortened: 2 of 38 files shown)


quixstreams/state/manager.py
```diff
@@ -230,7 +244,8 @@ def clear_stores(self):
                 "Cannot clear stores with active partitions assigned"
             )
 
-        shutil.rmtree(self._state_dir)
+        if self._state_dir:
```
gwaramadze:
```suggestion
        if self._state_dir is not None:
```

quixstreams/models/types.py
```diff
@@ -1,13 +1,16 @@
-from typing import Dict, List, Optional, Tuple, Union
+from typing import List, Mapping, Optional, Sequence, Tuple, Union
 
 from typing_extensions import Protocol
 
-MessageKey = Union[str, bytes]
+MessageKey = Optional[Union[str, bytes]]
```
gwaramadze, line 5: I think version without `Optional` is better. Using `Optional[MessageKey]` is more explicit and accounts for the possibility that some places will require non-optional `MessageKey`.

quixstreams/models/rows.py
```diff
@@ -19,7 +19,7 @@ def __init__(
         key: Optional[Any],
         timestamp: int,
         context: MessageContext,
-        headers: Optional[MessageHeadersTuples] = None,
+        headers: KafkaHeaders = None,
```
gwaramadze, line 22: Same here. It's slightly counterintuitive why do I see not `Optional` (or `| None` for later pythons) but default is set to `None`.

quixstreams/state/metadata.py
```diff
@@ -1,13 +1,14 @@
...
 METADATA_CF_NAME = "__metadata__"
 
-Undefined = NewType("Undefined", object)
-UNDEFINED = Undefined(object())
-DELETED = Undefined(object())
-
 DEFAULT_PREFIX = b""
+
+
+class markers(enum.Enum):
```
gwaramadze: That must not be lowercase plural but uppercase singular. Type hint like `Union[bytes, markers]` must be `Union[bytes, Marker]`.

```suggestion
class Marker(enum.Enum):
```

quixstreams/models/types.py
ghost: I'm a bit torn on this. I keep changing it back and forth. The main argument for including optional in the type is that confluent kafka can have messages keys/values/headers be `None`.

Having a non-optional `MessageKey` is not possible, it's not a message key anymore but a subset.

gwaramadze (approved):

PR #666, opened 2024-12-03
FileSource background file fetcher

tim-quix:
This adds a simple queue to the `FileSource` connector to download the next file while the current one is being processed.


tim-quix:
made a new rebased branch

PR #670, opened 2024-12-03
Source: background file downloads for FileSource

tim-quix:
This adds a simple queue to the FileSource connector to download the next file while the current one is being processed.

### DIFF
--- a/quixstreams/sources/community/file/file.py
+++ b/quixstreams/sources/community/file/file.py
@@ -1,6 +1,5 @@
 import logging
-import queue
-import threading
+from concurrent.futures import Future, ThreadPoolExecutor
 from pathlib import Path
 from time import sleep
 from typing import BinaryIO, Optional, Union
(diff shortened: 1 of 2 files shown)


quixstreams/sources/community/file/file.py
```diff
@@ -16,6 +19,54 @@
...
+            return file_name, file_content
+        except Exception as e:
+            logger.error("FileFetcher encountered an error", exc_info=e)
+            self.stop()
+            raise e
+
+    def stop(self):
+        logger.info("Stopping file download thread...")
+        self._stopped = True
+        self._executor.shutdown(wait=False)
```
ghost: we should cancel the in-progress future before shutting down.

We also should `wait=True` the shutdown to make it clean

tim-quix: So the original reason I did this was to hopefully kill the file download immediately rather than waiting to complete it, as it seems you cant in actuality cancel already running futures (you can still call cancel on it, but from what I can tell, it won't do anything). 

However, I think for now I'll do a wait, and later on if we really want to we can add a mid-download stop (where you read chunks of data of the file at a time and look for stopping). I can't imagine people will be stopping in the middle very often anyway.

ghost: In any case we have the source shutdown timeout

ghost (approved):

PR #673, opened 2024-12-04
mypy: make quixstreams.models.* pass type checks

ghost:


quixstreams/models/topics/manager.py
```diff
@@ -211,6 +211,9 @@ def _get_source_topic_config(
             topic_name
         ] or deepcopy(self._non_changelog_topics[topic_name].config)
 
+        if topic_config is None:
+            raise RuntimeError(f"No configuration can be found for topic {topic_name}")
+
```
ghost: When exactly can this happen?

quixstreams/models/topics/topic.py
```diff
@@ -137,6 +139,23 @@ def __clone__(
...
+        if self.config is None:
+            return NewTopic(topic=self.name)
+
+        return NewTopic(
+            topic=self.name,
+            num_partitions=self.config.num_partitions,
+            replication_factor=self.config.replication_factor,
+            config=self.config.extra_config,
+        )
+
```
ghost, line 158: Nice change 👍

ghost (dismissed):

quixstreams/models/topics/manager.py
ghost: `admin.inspect_topics` returns `{"topic-name": None}` if the topic doesn't already exist. 

While I don't think we can get to the exception using the app if you directly use the topic manager it's possible by calling `repartition_topic()` or `changelog_topic()` before registering the "main" topic.

ghost (approved):

PR #685, opened 2024-12-10
mypy: make quixstreams.core.* pass type checks

ghost:


quixstreams/core/stream/functions/base.py
```diff
@@ -38,6 +38,9 @@ def _resolve_branching(self, *child_executors: VoidExecutor) -> VoidExecutor:
         If there's only one executor - copying is not neccessary, and the executor
         is returned as is.
         """
+        if not child_executors:
+            raise TypeError("At least one executor is required")
```
ghost: I think it should be `ValueError` in this case since it's related to the input params

ghost (approved):

PR #688, opened 2024-12-13
Feature: Collection Aggregations

gwaramadze:
This PR introduces a new `.collect()` aggregation method for windowed operations that efficiently gathers all values within a window into a list.

## Features
- New `.collect()` method for windowed aggregations
- Optimized performance compared to using `reduce()` for list building
- Supports all window types (tumbling, hopping, sliding)

## Implementation Details
- Only supports `.final()` mode to ensure complete window results
- Groups values by message key like other window operations
- Returns standard window result format with collection in the `value` field:
  ```python
  {
    'start': window_start_ms,
    'end': window_end_ms,
    'value': [value1, value2, ...]
  }
  ```

### DIFF
--- a/quixstreams/dataframe/windows/definitions.py
+++ b/quixstreams/dataframe/windows/definitions.py
@@ -189,6 +189,17 @@ class FixedTimeWindowDefinition(abc.ABC):
             func_name="min", aggregate_func=func, aggregate_default=None
         )
 
+    def collect(self) -> "FixedTimeWindow":
+        def func(old: Any, new: Any) -> None:
+            return None
+
+        return self._create_window(
+            func_name="collect",
+            aggregate_func=func,
+            aggregate_default=None,
+            aggregate_collection=True,
+        )
+
 
 class HoppingWindowDefinition(FixedTimeWindowDefinition):
     def __init__(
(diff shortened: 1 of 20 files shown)


quixstreams/dataframe/windows/definitions.py
```diff
@@ -189,6 +190,36 @@ def func(old: Any, new: Any) -> Any:
             func_name="min", aggregate_func=func, aggregate_default=None
         )
 
+    def collect(self) -> "FixedTimeWindow":
```
ghost: Since you're using direct `FixedTimeWindow` references instead of `"FixedTimeWindow"`, can you update them here and it other methods too?

quixstreams/dataframe/windows/time_based.py
```diff
@@ -97,16 +100,24 @@ def process_window(
                 )
                 continue
 
+            if collect:
+                state.update_window(start, end, value=None, timestamp_ms=timestamp_ms)
+                continue
```
ghost: Let's add a comment here explaining why it needs to store an empty window with "collect" is True

ghost (approved):

PR #695, opened 2024-12-19
mypy: make quixstreams.dataframe.* pass type checks

ghost:


quixstreams/dataframe/series.py
```diff
@@ -275,31 +269,46 @@ def test(
         context.run(composed, value, key, timestamp, headers)
         return result
 
-    @_validate_operation
+    def _validate_other_serie(self, other: Any) -> None:
```
gwaramadze: I think is should be `_validate_other_series`

quixstreams/dataframe/windows/base.py
```diff
@@ -27,7 +27,7 @@ def get_window_ranges(
     if not step_ms:
         step_ms = duration_ms
 
-    window_ranges = deque()
+    window_ranges: Deque[Tuple[int, int]] = deque()
```
gwaramadze: We can use `Deque[tuple[int, int]]` since we dropped Python 3.8.

I'm not sure if this would work, you can try `deque[tuple[int, int]]`

ghost: Ah nice, forgot this was part of 3.9 :+1:

gwaramadze (approved):

PR #700, opened 2025-01-06
Fix lateness warnings in Windows

ghost:
Currently, late messages in Windows can be logged with negative `late_by_ms` values because it's calculated based on the start of the maximum expired window. 
The timestamp can be larger than the start (if it's within this expired window) but still late.

This PR calculates the `late_by_ms` based on the end of the maximum expired window.

### DIFF
--- a/quixstreams/dataframe/windows/time_based.py
+++ b/quixstreams/dataframe/windows/time_based.py
@@ -85,14 +85,15 @@ class FixedTimeWindow:
 
         state_ts = state.get_latest_timestamp() or 0
         latest_timestamp = max(timestamp_ms, state_ts)
-        max_expired_window_start = latest_timestamp - duration_ms - grace_ms
+        max_expired_window_end = latest_timestamp - grace_ms
+        max_expired_window_start = max_expired_window_end - duration_ms
         updated_windows: list[WindowResult] = []
         for start, end in ranges:
             if start <= max_expired_window_start:
                 self._log_expired_window(
                     window=[start, end],
                     timestamp_ms=timestamp_ms,
-                    late_by_ms=max_expired_window_start + 1 - timestamp_ms,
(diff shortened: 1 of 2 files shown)


quixstreams/dataframe/windows/time_based.py
```diff
@@ -85,14 +85,15 @@ def process_window(
...
+        max_expired_window_end = latest_timestamp - grace_ms
+        max_expired_window_start = max_expired_window_end - duration_ms
         updated_windows: list[WindowResult] = []
         for start, end in ranges:
             if start <= max_expired_window_start:
                 self._log_expired_window(
                     window=[start, end],
                     timestamp_ms=timestamp_ms,
-                    late_by_ms=max_expired_window_start + 1 - timestamp_ms,
+                    late_by_ms=max_expired_window_end - timestamp_ms,
```
gwaramadze, line 96: I'm not sure if + 1 really should be removed. Are you sure?

gwaramadze: If it should, then it should also be removed for sliding windows, shouldn't it?

gwaramadze (approved):
Not sure about the +1 thing, but overall this change is undoubtfully an improvement.

quixstreams/dataframe/windows/time_based.py
ghost: > If it should, then it should also be removed for sliding windows, shouldn't it?
The window end is exclusive here, so I think no +1 is needed.

For example, if the max expired window is `[100, 200)` and the timestamp is `199`, the `late_by_ms` will be `1`, which is correct - if the timestamp is 200, it would be outside of this expired window

ghost: The window end is inclusive for sliding windows, so the `+1` should stay. 
To be outside of the expired window `[100, 200]`, the timestamp must be `201`.

PR #701, opened 2025-01-07
Add a callback to react on late messages in Windows

ghost:
## What's changed

- Added an optional `on_late` callback for time windows to react on late-arriving records and configure their logging.  
If the callback returns `True`, the app will log late events as usual.  
If it returns `False`,  nothing will be logged
- `.process_window()` methods now accept `key` parameter to pass it to the callback, therefore so many tests are changed.

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -903,6 +903,12 @@ class StreamingDataFrame:
         :param name: The unique identifier for the window. If not provided, it will be
             automatically generated based on the window's properties.
 
+        :param on_late: an optional callback to react on late records in windows and
+            to configure the logging of such events.
+            If the callback returns `True`, the message about a late record will be logged
+            (default behavior).
+            Otherwise, no message will be logged.
+
         :return: `TumblingWindowDefinition` instance representing the tumbling window
             configuration.
             This object can be further configured with aggregation functions
@@ -995,6 +1001,12 @@ class StreamingDataFrame:
         :param name: The unique identifier for the window. If not provided, it will be
             automatically generated based on the window's properties.
 
+        :param on_late: an optional callback to react on late records in windows and
+            to configure the logging of such events.
+            If the callback returns `True`, the message about a late record will be logged
+            (default behavior).
+            Otherwise, no message will be logged.
+
         :return: `HoppingWindowDefinition` instance representing the hopping
             window configuration.
             This object can be further configured with aggregation functions
@@ -1019,6 +1031,7 @@ class StreamingDataFrame:
         duration_ms: Union[int, timedelta],
         grace_ms: Union[int, timedelta] = 0,
         name: Optional[str] = None,
+        on_late: Optional[WindowOnLateCallback] = None,
     ) -> SlidingWindowDefinition:
         """
         Create a sliding window transformation on this StreamingDataFrame.
@@ -1082,6 +1095,12 @@ class StreamingDataFrame:
         :param name: The unique identifier for the window. If not provided, it will be
             automatically generated based on the window's properties.
 
+        :param on_late: an optional callback to react on late records in windows and
+            to configure the logging of such events.
+            If the callback returns `True`, the message about a late record will be logged
+            (default behavior).
+            Otherwise, no message will be logged.
+
         :return: `SlidingWindowDefinition` instance representing the sliding window
             configuration.
             This object can be further configured with aggregation functions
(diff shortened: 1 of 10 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -921,6 +927,7 @@ def hopping_window(
         step_ms: Union[int, timedelta],
         grace_ms: Union[int, timedelta] = 0,
         name: Optional[str] = None,
+        on_late: Optional[WindowOnLateCallback] = None,
```
gwaramadze, line 935: Should be added also for sliding window

quixstreams/dataframe/windows/time_based.py
```diff
@@ -88,13 +92,29 @@ def process_window(
         max_expired_window_end = latest_timestamp - grace_ms
         max_expired_window_start = max_expired_window_end - duration_ms
         updated_windows: list[WindowResult] = []
+        on_late = self._on_late
```
gwaramadze: We cache class attributes at the beginning of the method, so you might want to move this line to line 83 right after `default = self._aggregate_default`.

quixstreams/dataframe/windows/time_based.py
```diff
@@ -88,13 +92,29 @@ def process_window(
...
+                    late_by_ms,
+                    start,
+                    end,
+                    self._name,
+                    ctx.topic,
+                    ctx.partition,
+                    ctx.offset,
                 )
+                if to_log:
+                    self._log_expired_window(
```
gwaramadze: WDYT if we convert `_log_expired_window` into a default callback and make the callback itself return `None`?

quixstreams/dataframe/windows/base.py
```diff
@@ -14,6 +14,35 @@ class WindowResult(TypedDict):
...
+        self,
+        value: Any,
+        timestamp_ms: int,
+        late_by_ms: int,
+        start: int,
+        end: int,
+        store_name: str,
+        topic: str,
+        partition: int,
+        offset: int,
```
gwaramadze, line 29: Maybe instead of topic/partition/offset accept `MessageContext`?

quixstreams/dataframe/windows/definitions.py
```diff
@@ -23,6 +28,7 @@ def __init__(
         dataframe: "StreamingDataFrame",
         name: Optional[str] = None,
         step_ms: Optional[int] = None,
+        on_late: Optional[WindowOnLateCallback] = None,
```
gwaramadze, line 30: Maybe lets set the default here instead of running `or` in the `__init__`?
```suggestion
        on_late: WindowOnLateCallback = default_on_late_callback,
```

quixstreams/dataframe/windows/base.py
```diff
@@ -14,6 +14,35 @@ class WindowResult(TypedDict):
...
+    ) -> bool: ...
```
gwaramadze, line 30: The user needs to understand why the boolean is expected here and that it controls logging. I think it's rather counterintuitive. That's why in another comment I propose to change it to `-> None` and let users log however they want, and if they want because maybe they don't.

quixstreams/dataframe/windows/time_based.py
```diff
@@ -88,13 +92,29 @@ def process_window(
...
         max_expired_window_start = max_expired_window_end - duration_ms
         updated_windows: list[WindowResult] = []
+        on_late = self._on_late
         for start, end in ranges:
             if start <= max_expired_window_start:
-                self._log_expired_window(
-                    window=[start, end],
-                    timestamp_ms=timestamp_ms,
-                    late_by_ms=max_expired_window_end - timestamp_ms,
+                window = [start, end]
```
gwaramadze: `window` var is only needed after `if to_log`

quixstreams/dataframe/windows/time_based.py
```diff
@@ -127,16 +135,42 @@ def process_window(
...
+                value,
+                key,
+                timestamp_ms,
+                late_by_ms,
+                start,
+                end,
+                self._name,
+                ctx.topic,
+                ctx.partition,
+                ctx.offset,
```
gwaramadze, line 162: Deja vu comment. Why not pass full ctx to the callback?

quixstreams/dataframe/windows/time_based.py
```diff
@@ -127,16 +135,42 @@ def process_window(
...
-            f"window={window} "
-            f"late_by_ms={late_by_ms} "
-            f"partition={ctx.topic}[{ctx.partition}] "
-            f"offset={ctx.offset}"
-        )
+        to_log = True
+        # Trigger the "on_late" callback if provided.
+        # Log the lateness warning if the callback returns True
+        if on_late := self._on_late:
+            to_log = on_late(
```
gwaramadze: Double lookup is faster than one lookup and assignment.
```suggestion
        if self._on_late:
            to_log = self._on_late(
```

quixstreams/dataframe/windows/time_based.py
ghost: Mainly to avoid complicating the callback interface with the `MessageContext` class.

Having three params of plain types seems easier than dealing with another class that is not really used anywhere else in the public API.
Also, wanna give us some room if we want to change `MessageContext` in the future.

gwaramadze (approved):

Issue #484, opened 2024-09-05
Feature: Joins


ghost:
Implement API to join multiple StreamingDataFrames based on condition and timeframe.

This feature is being researched, and more details will be added later.

PR #705, opened 2025-01-14
Switch to "range" assignor strategy from "cooperative-sticky"

ghost:
Switched the partition assignment strategy to `"range"` from the previously used `"cooperative-sticky"`.
With `"cooperative-sticky"`, commits during rebalancing may fail, and we always commit the current checkpoint on assign and revoke.

According to [this comment](https://github.com/confluentinc/librdkafka/pull/3812#issuecomment-1567361344), `range` assignor now provides co-partitioning and ensures that the same partitions are assigned to the same members.

This property will help us to implement  #484  too.

I briefly verified that it works this way, but I want to do more tests to be 100% sure.

## Main changes
- Use `"range"` assignment strategy
- For recovery, subscribe to changelog topics too (previously subscribed only to input & repartition topics), and rely on assignment to be co-partitioned (same numbers to same consumers).
  -  Added a check in case it's not co-partitioned
- `RecoveryManager` now doesn't assign/unassign partitions and instead only pauses, seeks and resumes them


quixstreams/state/recovery.py
```diff
@@ -348,21 +368,24 @@ def assign_partition(
...
                 logger.debug(f"Adding a recovery check for {rp}")
                 self._recovery_partitions.setdefault(partition, {})[changelog_name] = rp
-                # note: technically it should be rp.offset + 1, but to remain backwards
-                # compatible with <v2.7 +1 ALOS offsetting, it remains rp.offset.
-                # This means we will always re-write the "first" recovery message.
-                # More specifically, this is only covering for a very edge case:
-                # when first upgrading from <v2.7 AND a recovery was actually needed.
-                # Once on >=v2.7, this is no longer an issue...so we could eventually
-                # remove this, potentially.
-                self._consumer.incremental_assign(
```
tim-quix, line 363: Minor thought: since we no longer actually do any actual assignment in this function anymore, it may be worth considering renaming the method to better reflect it's new behavior (`handle_assign` or simply `on_assign` or `on_partition_assign` if we aren't concerned with reusing the name in multiple spots.)

Same applies to `revoke_partition`.

quixstreams/state/recovery.py
```diff
@@ -387,13 +410,15 @@ def assign_partition(
...
+        Pauses all provided RecoveryPartitions and cleans up any remaining
+        empty dictionary references.
+
+        The actual unassignment is done by Consumer.
 
         :param recovery_partitions: a list of `RecoveryPartition`
         """
         partition_nums = {rp.partition_num for rp in recovery_partitions}
-        self._consumer.incremental_unassign(
+        self._consumer.pause(
```
tim-quix, line 421: This probably isn't doing any harm by any means (unless maybe some weird race condition could occur?), but I'd imagine we prob don't need to pause the to-be-revoked partitions here?

tim-quix (dismissed):
Overall surprisingly simple/minimal changes, nice!

Besides my comments here, I had a very small PR #710 to remove some other redundant stuff that was prob just missed while you cleaned stuff up.

quixstreams/state/recovery.py
ghost: The whole module could be refactored a bit.
But I want to keep the changes minimal so it's easier to review at this moment.

quixstreams/state/recovery.py
ghost: Well, the changelog partitions won't be revoked until the next rebalancing. 
The `RecoveryManager` doesn't assign/revoke partitions anymore. 
Also, there's no way to revoke one partition like with `incremental_unassign` - you can only revoke all or replace the assignment.

After recovery, they will remain assigned, so I pause them to ensure the consumer doesn't receive messages it will produce to these changelogs.

gwaramadze (approved):

PR #707, opened 2025-01-14
Optimize window deletion

gwaramadze:
Optimizations around deleting windows that eliminate the need to deserialize the window key before deletion. 

1. `WindowedRocksDBPartitionTransaction.delete_window` has been removed because it depended on deserialized input parameters. Deleting windows will now be handled by a lower-level `PartitionTransaction._delete_bytes` method.
2. `WindowedRocksDBPartitionTransaction.delete_windows` internally will not use `get_windows` but a lower-level `_get_items` method which returns serialized keys.
3. `WindowedRocksDBPartitionTransaction.expire_windows` no longer handles deleting windows. It has been ultimately delegated to the outer scope, specifically the `process_window` method.

### DIFF
--- a/quixstreams/state/base/state.py
+++ b/quixstreams/state/base/state.py
@@ -36,7 +36,7 @@ class State(ABC):
         ...
 
     @abstractmethod
-    def delete(self, key: bytes):
+    def delete(self, key: Any):
         """
         Delete value for the key.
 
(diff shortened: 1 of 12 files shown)


quixstreams/state/base/state.py
```diff
@@ -36,7 +36,7 @@ def set(self, key: Any, value: Any):
         ...
 
     @abstractmethod
-    def delete(self, key: Any):
+    def delete(self, key: bytes):
```
ghost: Let's change it for the other operations as well, get/set/exists

ghost (dismissed):

quixstreams/state/base/state.py
gwaramadze: Actually changing it was a mistake because it breaks non-windowed state usage. I changed it back to `key: Any`.

PR #708, opened 2025-01-14
Add optional connector connect success and failure callbacks

tim-quix:
## What

Provides the ability to add an optional callback for a connection attempt with a connector.

> NOTE: There are defaults that use our desired Quix Cloud log output.

A  `setup` method is now an additional requirement when constructing a new connector template. 

## How to Use

This feature is fully backwards compatible: usage does not change for users who do not wish to use the callback.

To use it, the pattern is the same across all sources and sinks:

```
from quixstreams.sinks.core.influxdb3 import InfluxDB3Sink

def client_connect_success_cb():
    print(f"CONNECTED!")

def client_connect_failure_cb(error):
    print(f"ERROR!")
    raise error

influxdb_sink = InfluxDB3Sink(
    token="BAD TOKEN",
    host="my host",
    ...,
    on_client_connect_success=client_connect_success_cb,
    on_client_connect_failure=client_connect_failure_cb,
)
```

## Other things:
- standardized some things like attributes across sources and sinks
- added minor things in order to test connections as expected (some client setups do not connect it immediately)
- fixed a few bugs here n' there

### DIFF
--- a/quixstreams/sinks/base/sink.py
+++ b/quixstreams/sinks/base/sink.py
@@ -81,7 +81,7 @@ class BaseSink(abc.ABC):
         """
 
     @abc.abstractmethod
-    def setup_client(self):
+    def setup(self):
         """
         When applicable, set up the client here along with any validation to affirm a
         valid/successful authentication/connection.
@@ -93,7 +93,7 @@ class BaseSink(abc.ABC):
         Allows using a callback pattern around the connection attempt.
         """
         try:
-            self.setup_client()
+            self.setup()
             self._client_connect_success_cb()
         except Exception as e:
             self._client_connect_failure_cb(e)
--- a/quixstreams/sinks/base/sink.py
+++ b/quixstreams/sinks/base/sink.py
@@ -72,6 +72,10 @@ class BaseSink(abc.ABC):
         """
 
     def start(self):
+        """
+        Called as part of `Application.run()` to initialize the sink's client.
+        Allows using a callback pattern around the connection attempt.
+        """
         error = None
         try:
             self.setup_client()
@@ -80,6 +84,7 @@ class BaseSink(abc.ABC):
         finally:
             if cb := self._client_connect_cb:
                 cb(error)
+            # Only raise if no callback; callback could intentionally supress errors
             elif error:
                 raise error
 
@@ -92,7 +97,7 @@ class BaseSink(abc.ABC):
         """
 
 
-class BatchingSink(BaseSink, abc.ABC):
+class BatchingSink(BaseSink):
     """
     A base class for batching sinks, that need to accumulate the data first before
     sending it to the external destinations.
--- a/quixstreams/sinks/base/sink.py
+++ b/quixstreams/sinks/base/sink.py
@@ -12,11 +12,11 @@ ClientConnectSuccessCallback = Callable[[], None]
 ClientConnectFailureCallback = Callable[[Optional[Exception]], None]
 
 
-def _default_client_connect_success_cb():
+def _default_on_client_connect_success():
     logger.info("CONNECTED!")
 
 
-def _default_client_connect_failure_cb(exception: Exception):
+def _default_on_client_connect_failure(exception: Exception):
     logger.error(f"ERROR: Failed while connecting to client: {exception}")
 
 
@@ -31,22 +31,22 @@ class BaseSink(abc.ABC):
 
     def __init__(
         self,
-        client_connect_success_cb: Optional[ClientConnectSuccessCallback] = None,
-        client_connect_failure_cb: Optional[ClientConnectFailureCallback] = None,
+        on_client_connect_success: Optional[ClientConnectSuccessCallback] = None,
+        on_client_connect_failure: Optional[ClientConnectFailureCallback] = None,
     ):
         """
-        :param client_connect_success_cb: An optional callback made after successful
+        :param on_client_connect_success: An optional callback made after successful
             client authentication, primarily for additional logging.
-        :param client_connect_failure_cb: An optional callback made after failed
+        :param on_client_connect_failure: An optional callback made after failed
             client authentication (which should raise an Exception).
             Callback should accept the raised Exception as an argument.
             Callback must resolve (or propagate/re-raise) the Exception.
         """
-        self._client_connect_success_cb = (
-            client_connect_success_cb or _default_client_connect_success_cb
+        self._on_client_connect_success = (
+            on_client_connect_success or _default_on_client_connect_success
         )
-        self._client_connect_failure_cb = (
-            client_connect_failure_cb or _default_client_connect_failure_cb
+        self._on_client_connect_failure = (
+            on_client_connect_failure or _default_on_client_connect_failure
         )
 
     @abc.abstractmethod
@@ -94,9 +94,9 @@ class BaseSink(abc.ABC):
         """
         try:
             self.setup()
-            self._client_connect_success_cb()
+            self._on_client_connect_success()
         except Exception as e:
-            self._client_connect_failure_cb(e)
+            self._on_client_connect_failure(e)
 
     def on_paused(self, topic: str, partition: int):
         """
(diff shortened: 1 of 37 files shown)


quixstreams/sinks/base/sink.py
```diff
@@ -17,6 +17,28 @@ class BaseSink(abc.ABC):
     Note that Sinks are currently in beta, and their design may change over time.
     """
 
+    def __init__(
+        self,
+        client_connect_cb: Optional[Callable[[Optional[Exception]], None]] = None,
+    ):
+        """
+        :param client_connect_cb: Am optional callback made once a client connection
+            is established. Callback expects an Exception or None as an argument.
```
gwaramadze: We could clarify this a bit, because it suggests that the cb will run _only_ `once a client connection is established`. We can get exceptions, maybe something along these lines:

```suggestion
        :param client_connect_cb: An optional callback invoked after attempting 
            to establish a client connection. The callback receives either a None for 
            successful connections or an Exception if the connection failed.
```

quixstreams/sinks/base/sink.py
```diff
@@ -17,6 +17,28 @@ class BaseSink(abc.ABC):
...
+        """
+        self._init_client(client_connect_cb)
+
+    def _init_client(self, client_connect_cb):
```
gwaramadze, line 30: add typings 😉

gwaramadze: missing typings 😉

quixstreams/sinks/base/sink.py
```diff
@@ -17,6 +17,28 @@ class BaseSink(abc.ABC):
     Note that Sinks are currently in beta, and their design may change over time.
     """
 
+    def __init__(
+        self,
+        client_connect_cb: Optional[Callable[[Optional[Exception]], None]] = None,
```
gwaramadze: Given that we use it in several places, you might want to pack this verbose type into a var like `ClientConnectCallback` or `ClientConnectCallbackType`.

quixstreams/sinks/base/sink.py
```diff
@@ -73,8 +102,16 @@ class BatchingSink(BaseSink):
...
+    ):
+        """
+        :param client_connect_cb: An optional callback made once a client connection
+            is established. Callback expects an Exception or None as an argument.
+        """
         self._batches = {}
+        super().__init__(client_connect_cb=client_connect_cb)
```
gwaramadze: Maybe `BatchingSink` doesn't need to know what parent class `__init__` is up to, just make sure it is called with whatever we throw at it.

I think it may safely run before initializing `_baches`.

```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._batches = {}
```

Removes some verbosity. But up to you 😉

quixstreams/sinks/community/bigquery.py
```diff
@@ -123,12 +123,17 @@ def __init__(
...
             kwargs["credentials"] = credentials
-
-        self._client = bigquery.Client(**kwargs)
-        logger.info("Successfully authenticated to BigQuery.")
-        if self.schema_auto_update:
-            # Initialize a table in BigQuery if it doesn't exist already
-            self._init_table()
+        self._client: Optional[bigquery.Client] = None
+        self._client_settings = kwargs
+        super().__init__(client_connect_cb=client_connect_cb)
```
gwaramadze: That's nasty that the parent class `__init__` _must_ run after the subclass `__init__` logic.

I'm not sure there is a way around it. `setup_client` depends on a ton of input shit 🤔

gwaramadze (commented):
A fistful of comments.

quixstreams/sinks/base/sink.py
tim-quix: Agreed, added some more details in the docstring :+1:

quixstreams/sinks/base/sink.py
tim-quix: Good thinking, added a `ClientConnectCallback` type =)

quixstreams/sinks/community/bigquery.py
tim-quix: Glad we chatted about this cuz I also had the icks about it :rofl:  

I think I have something much more elegant in place now...there is no longer any dependency on init order (I readjusted things back where necc).

Only downside is the implementation is slightly different for Source vs Sink and I needed to add a `start_sinks` method to the `SinkManager`, but they are inherently run differently so that's not really a big deal, and it was a very insignificant (and I believe easy to follow) addition.

This also opens the ability to have a sanctioned teardowns of the Sinks in the `SinkManager` too like the sources currently have, but I'm not gonna add that for now.

quixstreams/sinks/base/sink.py
```diff
@@ -48,6 +64,30 @@ def add(
...
+        try:
+            self.setup_client()
+        except Exception as e:
+            error = e
+        finally:
+            if cb := self._client_connect_cb:
+                cb(error)
+            # Only raise if no callback; callback could intentionally supress errors
+            elif error:
+                raise error
```
gwaramadze: Two alternatives that you might consider to shave off one assignment and `elif` check.

```python
        if self._client_connect_cb:
            try:
                self.setup_client()
            except Exception as e:
                self._client_connect_cb(e)
            else:
                self._client_connect_cb(None)
        else:
            self.setup_client()
```
or if you don't like nesting try/except blocks:

```python
        if not self._client_connect_cb:
            self.setup_client()
            return

        try:
            self.setup_client()
        except Exception as e:
            self._client_connect_cb(e)
        else:
            self._client_connect_cb(None)
```

quixstreams/sinks/community/file/sink.py
```diff
@@ -36,12 +42,21 @@ def __init__(
...
+            It should accept a single argument, which will be populated with an
+            Exception if connecting failed (else None).
+            If used, errors must be resolved (or propagated) with the callback.
         """
-        super().__init__()
         self._format = resolve_format(format)
         self._destination = destination or LocalDestination()
         self._destination.set_directory(directory)
         self._destination.set_extension(self._format)
+        super().__init__(client_connect_cb=client_connect_cb)
```
gwaramadze: This should go to the beginning of the function right?

quixstreams/sinks/community/iceberg.py
```diff
@@ -139,37 +147,46 @@ def __init__(
...
+            location=self._iceberg_config.location,
+            partition_spec=self._partition_spec,
             properties={"write.distribution-mode": "fanout"},
         )
         logger.info(
-            f"Loaded Iceberg table '{table_name}' at '{self.iceberg_config.location}'."
+            f"Loaded Iceberg table '{self._table_name}' at "
+            f"'{self._iceberg_config.location}'."
         )
+        return
```
gwaramadze: redundant 😉
```suggestion
```

quixstreams/sinks/community/file/sink.py
```diff
@@ -36,12 +42,21 @@ def __init__(
...
+
+    def setup_client(self) -> Destination:
+        self._destination.connect()
+        return self._destination
```
gwaramadze: Lets remove this line for consistency

quixstreams/sinks/community/redis.py
```diff
@@ -44,21 +45,36 @@ def __init__(
...
+            "socket_timeout": socket_timeout,
+            **kwargs,
+        }
+
+    def setup_client(self):
+        self._redis_uri = (
+            f"{self._client_settings['host']}:"
+            f"{self._client_settings['port']}/"
+            f"{self._client_settings['db']}"
+        )
```
gwaramadze, line 82: Maybe move it to `__init__` as it is not used for actual connection. I see that it utilized only in a log.

quixstreams/sources/base/source.py
```diff
@@ -11,7 +11,7 @@
 
 logger = logging.getLogger(__name__)
 
-__all__ = ("BaseSource", "Source", "StatefulSource")
+ClientConnectCallback = Optional[Callable[[Optional[Exception]], None]]
```
gwaramadze: This is defined twice. I just realized we don't have a place for a common source/sink code 🤔
Not sure where to put it. Did you think about it?

gwaramadze (commented):
I like the change😉

I'm still rather against locking on the "client" word, but it's just a name, not a big blocker.

A separate problem I see here is the fact that we don't really test community connectors and I see a ton of changes. A tad too much to be confident that everything will click 🙈

quixstreams/sources/base/source.py
tim-quix: Yeah, I was thinking we could maybe put something in our `quixstreams/utils`, or maybe a new file or folder in root like `quixstreams/connector-utils`,

Maybe right now we aren't at the tipping point for common code quite yet, but it did cross my mind for sure.

We also run into similar things with the various connection patterns across all the sources/sinks. Lots of duplicated code here n' there.

quixstreams/sinks/community/file/destinations/azure.py
```diff
@@ -85,6 +86,11 @@ def _validate_container(self) -> None:
             raise
         raise AzureContainerNotFoundError(f"Container not found: {self._container}")
 
+    def connect(self):
```
ghost:
```suggestion
    def setup_client(self):
```

ghost (dismissed):

quixstreams/sinks/core/influxdb3.py
```diff
@@ -177,6 +194,16 @@ def _tags_callable(self, setter: TagsSetter) -> TagsCallable:
...
         return lambda value: setter
 
+    def setup(self):
+        self._client = InfluxDBClient3(**self._client_args)
+        try:
+            # We cannot safely parameterize the table (measurement) selection, so
+            # the best we can do is confirm authentication was successful
+            self._client.query("")
+        except Exception as e:
+            if "No SQL statements were provided in the query string" not in str(e):
```
tomas-quix, line 204: We can list measurements here, for example.

ghost (approved):

PR #709, opened 2025-01-14
Implement count windows

ghost:
Implement `sdf.tumbling_count_window`. It creates a count-based tumbling window transformation on the StreamingDataFrame.

Tumbling windows divide messages into fixed-batch, non-overlapping windows. They allow performing stateful aggregations like `sum`, `reduce`, etc. on top of the data and emit results downstream.

This type of window close after a set count of messages.

  ```python
  app = Application()
  sdf = app.dataframe(...)

  sdf = (
      # Define a tumbling window of 10 messages
      sdf.tumbling_count_window(count=10)

      # Specify the aggregation function
      .sum()

      # Specify how the results should be emitted downstream.
      # "current()" will emit results as they come for each updated window,
      # possibly producing multiple messages per key-window pair
      # "final()" will emit windows only when they are closed and cannot
      # receive any updates anymore.
      .current()
  )
  ```


quixstreams/dataframe/dataframe.py
```diff
@@ -844,9 +845,9 @@ def tumbling_window(
         duration_ms: Union[int, timedelta],
         grace_ms: Union[int, timedelta] = 0,
         name: Optional[str] = None,
-    ) -> TumblingWindowDefinition:
+    ) -> FixedTimeTumblingWindowDefinition:
         """
-        Create a tumbling window transformation on this StreamingDataFrame.
+        Create a time-based tumbling window transformation on this StreamingDataFrame.
```
gwaramadze, line 852: You can add "time-based" for hopping and sliding windows too.

quixstreams/dataframe/windows/count_based.py
```diff
@@ -0,0 +1,122 @@
...
+        if data is None:
+            data = CountWindowData(
+                count=0, start=timestamp_ms, end=timestamp_ms, value=[]
+            )
+
+        data["count"] += 1
+        if timestamp_ms > data["end"]:
+            data["end"] = timestamp_ms
+
+        data["value"].append(value)
```
gwaramadze: This will perform similarly to `reduce` because the whole list must be deserialized and re-serialized on every incoming message. You must use `state.add_to_collection(value)` instead.

Then to collect the value for windows closure you will use `state._get_values` method so it will have to be made "public" for this purpose.

gwaramadze (changes requested):

quixstreams/dataframe/windows/count_based.py
```diff
@@ -0,0 +1,133 @@
...
+class CountWindowData(TypedDict):
+    count: int
+    start: int
+    end: int
+    value: Any
+
+
+class FixedCountWindow(Window):
+    STATE_KEY = "metadata"
+    VALUE_KEY = "value"
```
gwaramadze: this seems unused

quixstreams/dataframe/windows/count_based.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if data["count"] < self._max_count:
+            state.add_to_collection(data["count"], value)
+            state.set(key=self.STATE_KEY, value=data)
+            return [], []
+
+        # window is full, closing ...
+        state.delete(key=self.STATE_KEY)
+
+        values = state.get_from_collection(start=1, end=10)
```
gwaramadze: can you explain why is this constant 1 and 10?

quixstreams/dataframe/windows/count_based.py
```diff
@@ -0,0 +1,133 @@
...
+        elif timestamp_ms > data["end"]:
+            data["end"] = timestamp_ms
+
+        if data["count"] < self._max_count:
+            state.add_to_collection(data["count"], value)
+            state.set(key=self.STATE_KEY, value=data)
+            return [], []
+
+        # window is full, closing ...
+        state.delete(key=self.STATE_KEY)
```
gwaramadze: maybe this should be moved by the end of the method, together with `state.delete_from_collection`

quixstreams/dataframe/windows/count_based.py
gwaramadze: Also maybe lets name the method `get_collection` and variable also `collection` because `value` in this context has slightly different meaning.
```suggestion
        collection = state.get_collection(start=1, end=10)
```

Given that, renaming `_get_values` to `get_collection` is a good move.

Issue #487, opened 2024-09-05
Feature: API to merge Streaming DataFrames


ghost:
In #419, we introduce branching to Streaming DataFrames to implement non-linear processing. 

In this feature, we want to implement the merging of these branches back into a single one as well as merge StreamingDataFrames created from different topics and sources

ghost:
Closed by #802

PR #725, opened 2025-01-23
 StreamingDataFrame.merge(): prep work

ghost:
This PR adds some prep work for the `merge()` feature.
No new features have been added yet, and SDFs are expected to work like before.

I want to merge it before implementing the actual `merge()` to split the review process into steps.

### Main changes:
- Added support for multiple parents in `Stream` class to implement the merge() feature.
- The `Stream.compose()` method now uses `graphlib.TopologicalSorter` to resolve the sequence of operations in the tree of `Stream` objects.
- With multiple parents, each tree may have multiple root Streams (e.g. when merging 2 SDFs will result in a single tree with two starting points). This PR handles that.
 
Related to #487


quixstreams/core/stream/stream.py
```diff
@@ -302,91 +299,101 @@ def diff(self, other: Self) -> Self:
...
+                )
+
+            diff.append(stream)
+
+        if not diff:
+            # The case of "sdf['x'] = sdf" or "sdf = sdf[sdf]"
+            raise InvalidOperation("Cannot assign or filter using the same SDF")
+
+        # Reverse the diff to be in "parent->children" order
+        diff.reverse()
```
gwaramadze: Maybe use deque and append to the left

quixstreams/core/stream/stream.py
```diff
@@ -302,91 +299,101 @@ def diff(self, other: Self) -> Self:
...
-        while (parent := node.parent) and (allow_splits or len(parent.children) < 2):
-            tree_.append(parent)
-            node = node.parent
-
-        # Reverse to get expected ordering.
-        tree_.reverse()
-
-        return tree_
+        # Cut off the diffed nodes from the rest of the tree
+        # by removing the diff head node the current node's children
```
gwaramadze: missing "from"?
```suggestion
        # by removing the diff head node from the current node's children
```

quixstreams/core/stream/stream.py
```diff
@@ -302,91 +299,101 @@ def diff(self, other: Self) -> Self:
...
+
+            diff.append(stream)
+
+        if not diff:
+            # The case of "sdf['x'] = sdf" or "sdf = sdf[sdf]"
+            raise InvalidOperation("Cannot assign or filter using the same SDF")
+
+        # Reverse the diff to be in "parent->children" order
+        diff.reverse()
+        diff_head = diff[0]
```
gwaramadze: As far as I can see `diff` is never used fully so to slightly improve readability and later avoid `diff[1:]` you could do this here.
```suggestion
        diff_head, *diff_rest = diff
```
or some other name like `diff_tail`, `diff_remainder`, idk.

quixstreams/core/stream/stream.py
```diff
@@ -302,91 +299,101 @@ def diff(self, other: Self) -> Self:
...
+
+        if diff_head.pruned:
             raise InvalidOperation(
                 "Cannot use a filtering or column-setter SDF more than once"
             )
-
-        elif diff[0] not in self.children:
+        elif not self_found:
```
gwaramadze: Are you sure it shouldn't be `if` instead of `elif`?

quixstreams/core/stream/stream.py
```diff
@@ -302,91 +299,101 @@ def diff(self, other: Self) -> Self:
...
+        while to_traverse:
+            node = to_traverse.popleft()
+            if node in visited:
+                continue
+            sorter.add(node, *node.parents)
+            visited.add(node)
+            for parent in node.parents:
+                to_traverse.append(parent)
+            for child in node.children:
+                to_traverse.append(child)
```
gwaramadze: Consider something more concise
```suggestion
            to_traverse.extend(node.parents)
            to_traverse.extend(node.children)
```
or
```
to_traverse += node.parents + node.children
```

quixstreams/core/stream/stream.py
```diff
@@ -451,8 +460,9 @@ def compose_returning(self) -> ReturningExecutor:
         # after executing the Stream.
         # The composed stream must have only the "apply" functions,
         # which always return a single.
```
gwaramadze: I know it's not a part of the PR, but could you try and improve this comment? It's like "a single" is used as a noun in this sentence and it sounds weird.

quixstreams/core/stream/stream.py
```diff
@@ -465,96 +475,74 @@ def compose_returning(self) -> ReturningExecutor:
...
-        if not diff:
-            raise ValueError("The diff is empty")
-        return diff
+    def root_path(self) -> List[Self]:
+        """
+        Start from self and collect all parents until reaching the root nodes
+        """
+        nodes = [self]
+        to_traverse: Deque = collections.deque()
+        to_traverse.append(self)
```
gwaramadze:
```suggestion
        to_traverse: Deque = collections.deque(nodes)
```

tests/test_quixstreams/test_core/test_stream.py
```diff
@@ -50,11 +50,11 @@ def test_root_path(self):
...
         assert len(tree) == 5
-        assert isinstance(tree[0].func, ApplyFunction)
-        assert isinstance(tree[1].func, ApplyFunction)
+        assert isinstance(tree[0].func, TransformFunction)
+        assert isinstance(tree[1].func, UpdateFunction)
         assert isinstance(tree[2].func, FilterFunction)
-        assert isinstance(tree[3].func, UpdateFunction)
-        assert isinstance(tree[4].func, TransformFunction)
+        assert isinstance(tree[3].func, ApplyFunction)
+        assert isinstance(tree[4].func, ApplyFunction)
```
gwaramadze, line 57: Can you please explain why the order is reversed?

ghost: Because `root_path()` now returns nodes in the child-to-parent order (last node in the tree comes first, "root" is the last)

quixstreams/core/stream/stream.py
ghost: It has slightly different meaning here:
`nodes` are used to accumulate the traversed result, and `to_traverse` is used to keep track of the nodes to be traversed

ghost: I think passing `nodes` to `deque` here confuses things.
But I agree that it can be shortened to `collections.deque([self])`.

quixstreams/core/stream/stream.py
ghost: It's a typo, yep

quixstreams/core/stream/stream.py
ghost: I don't think it matters here: if the first condition is True, it will raise an exception; otherwise, it will check the second condition. Both `if` and `elif` would work the same here.

gwaramadze (approved):

PR #733, opened 2025-01-31
Sink: MongoDB

tim-quix:
There's definitely a few different ways to approach this one, but I think I managed to strike a decent balance between "will do what most people want by default, or only requires simple flags or Literal change" along with a few "here's a couple of useful bells and whistles if you need em".

### DIFF
--- a/docs/connectors/sinks/mongodb-sink.md
+++ b/docs/connectors/sinks/mongodb-sink.md
@@ -39,6 +39,8 @@ Regardless of approach, the following is always true:
 How data is dumped with `MongoDBSink` primarily depends on two parameters:
 `update_method` and `document_matcher`.
 
+
+
 ### Default: kafka `key` == MongoDB `_id`
 
 The default `document_matcher` assumes the message `key` corresponds to an equivalently 
@@ -54,6 +56,8 @@ However, you can do full document replacement/set by
 setting `update_method="ReplaceOne"`, which will drop any fields that are not present in
 the message.
 
+
+
 ### Using A Custom `_id`
 
 A custom `_id` can be used by simply providing your own 
@@ -64,6 +68,7 @@ with that `_id`.
 
 If no `document_matcher` or `_id` specification is specified (and `upsert=True`), `MongoDB` will 
 create a new document where `_id` will be assigned an `ObjectID` (default MongoDB behavior).
+
 
 
 ### Alternate behavior: pattern-based updates
@@ -78,6 +83,37 @@ only the first encountered match will be updated.
 
 If no match is made, it will instead create a new document with a random `_id` 
 (assuming `upsert=True`) with the provided updates.
+
+
+
+### Include Message Metadata
+
+You can include `topic` (topic, partition, offset) and `message` 
+(key, headers, timestamp) metadata using the flags 
+`add_topic_metadata=True` and `add_message_metadata=True` for `MongoDBSink`. 
+
+They will be included as
+`__{field}` in the document.
+
+#### Example document with `add_message_metadata=True`:
+```python
+{
+    "field_x": "value_a",
+    "field_y": "value_b",
+    "__key": b"my_key",
+    "__headers": {},
+    "__timestamp": 1234567890,
+}
+
+```
+
+### Final Outgoing Value Editing
+
+In case other callables need access to fields you would otherwise exclude in the
+document, you can optionally provide a callable to `value_selector` that receives the
+current document as an argument, and returns the desired finalized outgoing document.
+
+> **Note**: any of the `add_*_metadata` flags will have already added their data.
 
 ## How To Use
 
@@ -160,6 +196,10 @@ mongodb_sink = MongoDBSink(
     **Default**: False
 - `include_topic_metadata`: include topic, partition, and offset as `__{field}`    
     **Default**: False
+- `value_selector`: An optional callable that allows final editing of the
+     outgoing document (right before submitting it).    
+    Largely used when a field is necessary for `document_matcher`, but not otherwise.    
+    **NOTE**: metadata is added before this step, so don't accidentally exclude it here!
 - Additional keyword arguments are passed to the `MongoClient`.
 
 ## Error Handling and Delivery Guarantees
@@ -168,7 +208,7 @@ The sink provides **at-least-once** delivery guarantees, which means:
 
 - Messages are published in batches for better performance
 - During checkpointing, the sink waits for all pending publishes to complete
-- If any messages fail to publish, a `SinkBackpressureError` is raised
+- If any messages fail to publish after several retries, a `SinkBackpressureError` is raised
 - When `SinkBackpressureError` occurs:
     - The application will retry the entire batch from the last successful offset
     - Some messages that were successfully published in the failed batch may be published again
--- a/docs/connectors/sinks/mongodb-sink.md
+++ b/docs/connectors/sinks/mongodb-sink.md
@@ -29,7 +29,7 @@ Regardless of approach, the following is always true:
 
 - The Kafka message `value` fields are dumped as whatever objects they are.
 - The sink ensures that the order of messages is preserved within each partition. 
-    - This means that messages are sent to MongoDB in the same order they are received from Kafka for each specific partition.
+- This means that messages are sent to MongoDB in the same order they are received from Kafka for each specific partition.
 
 
 ## Export Behavior
--- a/docs/connectors/sinks/mongodb-sink.md
+++ b/docs/connectors/sinks/mongodb-sink.md
@@ -74,7 +74,7 @@ create a new document where `_id` will be assigned an `ObjectID` (default MongoD
 ### Alternate behavior: pattern-based updates
 
 The `document_matcher` can alternatively be used to match more than one document at a time; it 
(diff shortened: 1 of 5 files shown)


docs/connectors/sinks/mongodb-sink.md
```diff
@@ -0,0 +1,262 @@
+# MongoDB Sink
+
+!!! info
+
+    This is a **Community** connector. Test it before using in production.
+
+    To learn more about differences between Core and Community connectors, see the [Community and Core Connectors](../community-and-core.md) page.
+
+This sink writes data to an MongoDB Database, with a few ways to dump data.
```
gwaramadze:
```suggestion
This sink writes data to a MongoDB Database, with a few ways to dump data.
```

docs/connectors/sinks/mongodb-sink.md
```diff
@@ -0,0 +1,262 @@
...
+
+By default, it assumes 1:1 relationship with MongoDB's `_id` field and Kafka keys, but 
+other update patterns are possible.
```
gwaramadze: This is mentioned a moment later in How It Works

```suggestion
```

docs/connectors/sinks/mongodb-sink.md
```diff
@@ -0,0 +1,262 @@
...
+create a new document where `_id` will be assigned an `ObjectID` (default MongoDB behavior).
+
+
+#### Example
+```python
+from quixstreams.sinks.community.mongodb import MongoDBSink
+from quixstreams.sinks.base.item import SinkItem
+
+def match_on_last_name(batch_item: SinkItem):
+    return {"_id": SinkItem.value["name"]["last"]}
```
gwaramadze:
```suggestion
    return {"_id": batch_item.value["name"]["last"]}
```

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+
+try:
+    from pymongo import MongoClient, ReplaceOne, UpdateMany, UpdateOne
+    from pymongo.collection import Collection
+    from pymongo.errors import (
+        ExecutionTimeout,
+        NetworkTimeout,
+        WriteConcernError,
+    )
+
```
gwaramadze:
```suggestion
```

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+            except (NetworkTimeout, ExecutionTimeout, WriteConcernError) as e:
+                logger.error(f"MongoDB: Encountered a network or server error: {e}")
+            attempts_remaining -= 1
+            if attempts_remaining:
+                backoff = 5
+                logger.error(
+                    f"Sleeping for {backoff} seconds and re-attempting. "
+                    f"Attempts remaining: {attempts_remaining}"
+                )
+                time.sleep(backoff)
```
gwaramadze: 1. I'd move it to except block, otherwise it is slightly confusing.
2. You can decrement `attempts_remaining` at the end to avoid `if`.
3. There are two backoff constants, maybe you can set those in the module scope as `ATTEMPT_BACKOFF` and `RETRY_BACKOFF`.
```suggestion
                logger.error(
                    f"Sleeping for {ATTEMPT_BACKOFF} seconds and re-attempting. "
                    f"Attempts remaining: {attempts_remaining}"
                )
                time.sleep(ATTEMPT_BACKOFF)
                attempts_remaining -= 1
```

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+        backoff = 30
```
gwaramadze: Use module level constant, mentioned in another commit.

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+                    "Sent update requests to MongoDB; "
+                    f"messages_processed={batch.size} "
+                    f"host_address={self._client.address[0]} "
+                    f"database={self._db_name} "
+                    f"collection={self._collection_name} "
+                    f"min_timestamp={min_timestamp} "
+                    f"max_timestamp={max_timestamp} "
+                    f"time_elapsed={round(time.monotonic() - start, 2)}s"
+                )
+                return
```
gwaramadze, line 200: Could you move the log and return to the `else` block?

docs/connectors/sinks/mongodb-sink.md
```diff
@@ -0,0 +1,262 @@
...
+
+There are a few different ways to handle/export them with `MongoDBSink`, but the most common
+(and default) approach is having a 1:1 correspondence between Kafka message `key` and 
+document `_id`.
+
+Regardless of approach, the following is always true:
+
+- The Kafka message `value` fields are dumped as whatever objects they are.
+- The sink ensures that the order of messages is preserved within each partition. 
+    - This means that messages are sent to MongoDB in the same order they are received from Kafka for each specific partition.
```
ghost: indentation issue

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+        doc_match = self._document_matcher
+        update_method = self._update_method
+        upsert = self._upsert
+        value_selector = self._value_selector
+        is_update = self._update_method != ReplaceOne
+
+        if self._add_message_metadata or self._add_topic_metadata:
+            added_metadata = partial(self._add_metadata, batch.topic, batch.partition)
+        else:
+            added_metadata = None
```
ghost, line 156: Is this done only for optimisation ? Is there really a performance hit vs always calling `seld._add_metadata` ?

quixstreams/sinks/community/mongodb.py
```diff
@@ -0,0 +1,210 @@
...
+            "ReplaceOne": Replaces the first matching document (usually based on `_id`).
+            Default: "UpdateOne".
+        :param add_message_metadata: add key, timestamp, and headers as `__{field}`
+        :param add_topic_metadata: add topic, partition, and offset as `__{field}`
+        :param value_selector: An optional callable that allows final editing of the
+            outgoing document (right before submitting it).
+            Largely used when a field is necessary for `document_matcher`,
+            but not otherwise.
+            NOTE: metadata is added before this step, so don't accidentally
+            exclude it here!
```
ghost, line 106: I'm unclear why this is needed, couldn't one deal with that directly in the document_matcher ?

tim-quix: Sorry that it's unclear, maybe needs better names or descriptions?

Essentially, `document_matcher` is only the "find the documents I should update, based on this field matching criteria" which uses the original message (the state of it when you call `sdf.sink()`) for generating said criteria.

The `value_selector` part is for altering how the documents gathered from `document_matcher` get updated; it is applied separately...by default it assumes the full value from `sdf.sink()`), but you can use this to trim down superfluous field updates (like ones you dont want, or ones that didn't change and you just needed them for the `document_matcher`), or just change the values outright, etc. 

It is mostly necessary because you might need certain fields just for `document_matcher`, so you can't filter them out during SDF manipulation.

In short: `document_matcher` is the "what" documents to update, `value_selector` is the "how" those documents get updated.

`value_selector` is going to be a little niche, which is also why it's optional; it's likely not needed for most people, or only needed for extra speed/write efficiency. LMKYT, i can adjust as needed =)

quixstreams/sinks/community/mongodb.py
tim-quix: Yeah, I was torn on a little on the layout here. Indent suggestion seems reasonable and I originally had it that way before changing it to this :joy: 

The `if` is still useful here since it means not sleeping an extra `ATTEMPT_BACKOFF` seconds when it's not supposed to attempt anymore...in the grand scheme of things it probably wont matter, but that was my reasoning for doing it this way.

quixstreams/sinks/community/mongodb.py
tim-quix: Didn't know you could do an `else` in a `try` block, TIL. Makes a lot of sense :+1:

quixstreams/sinks/community/mongodb.py
tim-quix: Yeah, theoretically for optimization and a little bit for readability below...though really, this does assume the `partial` is faster than constantly looking up all 3 vars each time...I admittedly did not speed test it, maybe I should to sanity check?

ghost: yeah I'm not sure it's a significant speed boost. Always doing `value = added_metadata(record)` is much more readable even if it ends up doing nothing.

quixstreams/sinks/community/mongodb.py
ghost: ah I see, I was a bit thrown of by the description. I'm partial to pushing user to do this kind of manipulation in the sdf but as you say there can be cases when it's not enough.

gwaramadze (dismissed):

tomas-quix:
Sink is working for me. Free to get merge from my side (have not review code, just from user perspective).

gwaramadze:
@tim-quix you need to add an entry in `docs/build/build.py`

gwaramadze (approved):

PR #734, opened 2025-01-31
state: refactor collection store to not rely on timestamp

ghost:
n-based windows don't rely on timestamp, the state needs to reflect that.

### DIFF
--- a/quixstreams/state/rocksdb/windowed/state.py
+++ b/quixstreams/state/rocksdb/windowed/state.py
@@ -105,16 +105,16 @@ class WindowedTransactionState(TransactionState, WindowedState):
         """
         return self._transaction.delete_from_collection(end=end, prefix=self._prefix)
 
-    def get_highest_id(self) -> Optional[int]:
+    def get_latest_timestamp(self) -> Optional[int]:
         """
-        Get the highest observed message ID for the current message key.
+        Get the latest observed timestamp for the current message key.
 
-        Use this ID to determine if the arriving event is late and should be
+        Use this timestamp to determine if the arriving event is late and should be
         discarded from the processing.
 
-        :return: latest observed event ID
+        :return: latest observed timestamp
         """
-        return self._transaction.get_highest_id(prefix=self._prefix)
+        return self._transaction.get_latest_timestamp(prefix=self._prefix)
 
     def expire_windows(
         self,
--- a/quixstreams/state/rocksdb/windowed/state.py
+++ b/quixstreams/state/rocksdb/windowed/state.py
@@ -112,8 +112,9 @@ class WindowedTransactionState(TransactionState, WindowedState):
         Use this timestamp to determine if the arriving event is late and should be
         discarded from the processing.
 
-        :return: latest observed timestamp
+        :return: latest observed event timestamp in milliseconds
         """
+
         return self._transaction.get_latest_timestamp(prefix=self._prefix)
 
     def expire_windows(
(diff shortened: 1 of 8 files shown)


quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -246,8 +287,8 @@ def expire_windows(
         if delete:
             for (start, end), _ in expired_windows:
                 self.delete_window(start, end, prefix=prefix)
-            if collect:
-                self._delete_values(max_timestamp=start, prefix=prefix)
+            if collect and start is not None:
```
gwaramadze: We shouldn't need `start is not None` because earlier in this method there is a check
```
if not expired_windows:
    return []
```
So if expired_windows is not empty then `start` will be assigned at least once.

quixstreams/state/rocksdb/windowed/state.py
```diff
@@ -1,12 +1,13 @@
...
 
+from quixstreams.state.base import TransactionState
 from quixstreams.state.types import WindowedState
 
 if TYPE_CHECKING:
     from .transaction import WindowedRocksDBPartitionTransaction
 
 
-class WindowedTransactionState(WindowedState):
+class WindowedTransactionState(TransactionState, WindowedState):
```
gwaramadze, line 10: Can you explain why you decided to do multiple inheritance here?

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -124,27 +128,63 @@ def update_window(
...
-        key = encode_integer_pair(timestamp_ms, self._get_next_count())
+    ) -> int:
+        counter = self._get_next_count()
+        if id:
+            key = encode_integer_pair(id, counter)
+        else:
+            key = encode_integer_pair(counter, counter)
+
         self.set(key=key, value=value, prefix=prefix, cf_name=VALUES_CF_NAME)
+        return counter
```
gwaramadze, line 157: This function doesn't need to return anything. Where is this integer used?

quixstreams/state/rocksdb/windowed/transaction.py
ghost: The linter was complaining that's why I changed it but you are right it can't happen.

quixstreams/state/rocksdb/windowed/transaction.py
ghost: When adding an item to the collection the `id` (used to be timestamp) is now optional. If you don't pass an ID we fallback on the global counter, in that case it's useful to return the generated id.

n-based windows will leverage this to fetch the appropriate IDs from the collection

quixstreams/state/rocksdb/windowed/state.py
ghost: `TransactionState` provides an actual implementation for get/set/delete/exists. `WindowedState` only provide the protocol state used for windows should follow.

tbh I'd like to refactor the windowed state at one point, similar to what I did for the "normal" state when implementing the memory store.

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -213,17 +253,18 @@ def expire_windows(
...
 
-        self._set_timestamp(
+        self._set_highest_id(
             cache=self._last_expired_timestamps,
             prefix=prefix,
-            timestamp_ms=last_expired__gt,
+            id=last_expired__gt,
         )
 
+        start = None
```
gwaramadze: I tried removing this and I don't get linter errors. Can you try again?

ghost: hmm, vscode complains but not our linter. I'll remove it

gwaramadze (dismissed):

ghost (approved):

PR #736, opened 2025-02-03
implement count windows

ghost:
Implements:
* `sdf.tumbling_count_window`
* `sdf.hopping_count_window`
* `sdf.sliding_count_window`

Those windows divide messages into fixed-size batch of messages. They allow performing stateful aggregations like `sum`, `reduce`, etc. on top of a fixed number of messages and emit results downstream.

  ```python
  app = Application()
  sdf = app.dataframe(...)

  sdf = (
      # Define a tumbling window of 10 messages
      sdf.tumbling_count_window(count=10)

      # Specify the aggregation function
      .sum()

      # Specify how the results should be emitted downstream.
      # "current()" will emit results as they come for each updated window,
      # possibly producing multiple messages per key-window pair
      # "final()" will emit windows only when they are closed and cannot
      # receive any updates anymore.
      .current()
  )
  ```

### DIFF
--- a/docs/windowing.md
+++ b/docs/windowing.md
@@ -236,7 +236,7 @@ In a tumbing window each message is only assigned to a **single** interval.
 
 **Example**
 
-Imagine you have to interact with an external HTTP API. To optimise your data analysis pipeline you want to batch data before sending in to the external API. In this example we batch data from 3 messages before sending the request.
+Imagine you have to interact with an external HTTP API. To optimize your data analysis pipeline, you want to batch data before sending it to the external API. In this example, we batch data from 3 messages before sending the request.
 
 Input:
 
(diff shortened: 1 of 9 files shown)


docs/windowing.md
```diff
@@ -217,8 +218,88 @@ sdf = (
...
+[4, 7]  :     ....
+[8, 11] :         ....
+[12, 15] :            ....
+```
+
+In a tumbing window each message is only assigned to a **single** interval.
+
+**Example**
+
+Imagine you have to interact with an external HTTP API. To optimise your data analysis pipeline you want to batch data before sending in to the external API. In this example we batch data from 3 messages before sending the request.
```
ghost:
```suggestion
Imagine you have to interact with an external HTTP API. To optimize your data analysis pipeline, you want to batch data before sending it to the external API. In this example, we batch data from 3 messages before sending the request.
```

ghost (approved):

PR #747, opened 2025-02-12
windows: Implement per-partition window expiration for tumbling and h…

ghost:
…opping time windows


quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -282,18 +288,67 @@ def expire_windows(
...
+            for (start, end), _, _ in expired_windows:
                 self.delete_window(start, end, prefix=prefix)
             if collect:
                 self.delete_from_collection(end=start, prefix=prefix)
 
         return expired_windows
 
+    def expire_all_windows(
+        self,
+        max_end_time: int,
```
gwaramadze, line 332: The param is `max_end_time` but the docstring says `max_start_time`. I'm not sure which one is it now but something needs to be corrected.

quixstreams/dataframe/windows/time_based.py
```diff
@@ -50,13 +64,48 @@ def __init__(
...
+        its end timestamp + grace period.
+        The closed windows cannot receive updates anymore and are considered final.
+
+        >***NOTE:*** By default windows are closed within the same message key.
+        If some message keys appear irregularly in the stream, the latest windows
+        can remain unprocessed until a message with the same key is received.
+
+        :param expiration_strategy: choose the windows expiration strategy. Either per-key,
+            only message key windows are expired on new messages, or per-partition,
+            the windows of all the keys in the message partition are expired on new messages.
```
```suggestion

        :param expiration_strategy: the strategy to use when expiring windows.  
            Possible values: 
              - `"key"` - messages advance time and expire windows with the same key.    
              If some message keys appear irregularly in the stream, the latest windows can remain unprocessed until a message with the same key is received.
              - `"partition"` - messages advance time and expire windows for the whole partition to which this message key belongs.   
              If timestamps between keys are not ordered, it may increase the number of discarded late messages.
              Default - `"key"`.
```

quixstreams/dataframe/windows/time_based.py
```diff
@@ -97,21 +146,60 @@ def process_window(
...
+                WindowResult(start=start, end=end, value=self._merge_func(aggregated)),
+            )
+
+    def expire_by_key(
+        self,
+        key: bytes,
+        state: WindowedState,
+        max_expired_start: int,
+        collect: bool,
+    ) -> Iterable[WindowKeyResult]:
```
ghost: Can these two functions have similar API?
In one case it's `timestamp_ms`, in another `max_expired_start`.

quixstreams/state/base/partition.py
```diff
@@ -107,6 +107,17 @@ def exists(self, key: bytes, cf_name: str = "default") -> bool:
         :return: `True` if the key is present, `False` otherwise.
         """
 
+    @abstractmethod
+    def keys(self, cf_name: str = "default") -> Iterable[bytes]:
```
ghost: 1. How about naming it `iter_keys()` similar to `iter_items()`?
2. Should `iter_times` be an abstractmethod too?

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -282,18 +288,70 @@ def expire_windows(
...
+        step_ms: int,
+        delete: bool = True,
+        collect: bool = False,
+    ) -> Iterable[WindowDetail]:
+        """
+        Get all expired windows for all prefix from RocksDB up to the specified `max_start_time` timestamp.
```
ghost: It says "max_start_time", but the param is "max_end_time"

quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -282,18 +288,70 @@ def expire_windows(
...
+        """
+        last_expired = self._get_timestamp(
+            prefix=b"", cache=self._last_expired_timestamps, default=0
+        )
+
+        windows = windows_to_expire(last_expired, max_end_time, step_ms)
+        if not windows:
+            return
+
+        suffixes: tuple[bytes, ...] = tuple(
```
ghost: Could a set be better here?

Suffixes have a fixed length, so you can do this:
```
suffixes: set[bytes]

for key in keys:
    suffix = key[:8] # eight bytes
    if suffix in suffixes: ...
```

quixstreams/dataframe/windows/time_based.py
ghost: I changed `timestamp_ms` to `max_expired_end` but we can't have the same parameters since it's not the same timestamp we are using

quixstreams/state/base/partition.py
ghost: 1. I think `keys` is better, it's what's normally used to iterate over keys
2. We can do it, we don't need it right now

quixstreams/state/rocksdb/windowed/transaction.py
ghost: Good point. It's slightly faster from my tests

quixstreams/state/base/partition.py
ghost: >I think keys is better, it's what's normally used to iterate over keys

Then pls make it named similarly. 
Either `keys()` and `items()`, or `iter_keys()` and `iter_items()`

benchmark.py
ghost, line 1: Should it be included to the PR?

docs/windowing.md
```diff
@@ -990,10 +990,68 @@ Also, specifying a grace period using `grace_ms` will increase the latency, beca
 
 You can use `final()` mode when some latency is allowed, but the emitted results must be complete and unique.
 
->***NOTE:*** Windows can be closed only by the records with the **same** message key.
-> If some message keys appear irregularly in the stream, the latest windows 
-> can remain unprocessed until the message with the same key is received.
+##### Closing strategy
```
ghost:
```suggestion
## Closing strategies
```

tests/test_quixstreams/test_app.py
```diff
@@ -2429,128 +2429,128 @@ def on_message_processed(topic_, partition, offset):
...
+    #     self,
+    #     app_factory,
+    #     row_consumer_factory,
+    #     executor,
+    # ):
+    #     """
+    #     Test that StreamingDataFrame processes 6 messages from Kafka and groups them
+    #     by each record's specified column value.
+    #     """
+
```
ghost: You can use `pytest.mark.skip()` here instead of commenting the test.

quixstreams/dataframe/windows/time_based.py
```diff
@@ -97,21 +183,76 @@ def process_window(
...
+                    start=window_start,
+                    end=window_end,
+                    value=self._merge_func(aggregated),
+                ),
+            )
+
+        if count:
+            logger.debug(
+                "Expired %s windows in %ss", count, round(time.monotonic() - start, 2)
+            )
```
ghost, line 239: I like this addition.
What about doing the same for `expire_per_key`?

docs/windowing.md
```diff
@@ -990,10 +990,68 @@ Also, specifying a grace period using `grace_ms` will increase the latency, beca
...
+# -> Timestamp=8000, Key="A", value=1 -> emit nothing (the window is not closed yet)
+# -> Timestamp=10001, Key="A", value=1 -> emit one message with key "A" and value {"start": 0, "end": 10000, "value": 2}, the time has progressed beyond the window end for the "A" key.
+
+# Results:
+# (key="B", value={"start": 0, "end": 10000, "value": 2})
+# (key="A", value={"start": 0, "end": 10000, "value": 2})
+# No message for key "C" as the window is never closed since no messages with key "C" and a timestamp later than 10000 was received 
+```
+
+An alternative is to use the **partition** closing strategy. In this strategy messages advance time and close windows for the whole partition to which this message key belongs.
```
ghost:
```suggestion
An alternative is to use the **partition** closing strategy.  
In this strategy, messages advance time and close windows for the whole partition to which this key belongs.
```

docs/windowing.md
```diff
@@ -990,10 +990,69 @@ Also, specifying a grace period using `grace_ms` will increase the latency, beca
 
 You can use `final()` mode when some latency is allowed, but the emitted results must be complete and unique.
 
->***NOTE:*** Windows can be closed only by the records with the **same** message key.
-> If some message keys appear irregularly in the stream, the latest windows 
-> can remain unprocessed until the message with the same key is received.
+## Closing strategies
 
+By default windows use the **key** closing strategy. In this strategy messages advance time and close windows with the **same** message key.
```
ghost:
```suggestion
By default, windows use the **key** closing strategy.  
In this strategy, messages advance time and close only windows with the **same** message key.
```

ghost (approved):

PR #740, opened 2025-02-05
Feature: Table Print

gwaramadze:
This PR introduces a new experimental feature that allows real-time visualization of StreamingDataFrame records in a formatted table view.

### Key Features
- ✨ Live-updating table display of StreamingDataFrame records
- 📊 Support for both data and metadata columns
- ⚡ Automatic column discovery and ordering
- 📐 Customizable column widths
- 🕒 Timeout and slowdown parameters
- 🔄 Different behaviors for interactive and non-interactive environments
- 📑 Support for multiple tables printing simultaneously

### Printing Behavior
- **Interactive mode** (terminal/console):
  - Table refreshes in-place
  - New rows appear at the bottom
  - Old rows are removed from the top when full
  - Updates happen in real-time

- **Non-interactive mode** (redirected output):
  - Collects records until either:
    - Table is full
    - Timeout is reached
  - Prints complete table and starts new collection

### Example Usage

```python
sdf = app.dataframe(topic)
# Show the last 5 records, update at most every 1 second
sdf.print_table(size=5, title="Live Records", slowdown=1)
```

This will produce a live-updating table like this:

        Live Records
        ┏━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━┳━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┓
        ┃ _key       ┃ _timestamp ┃ active ┃ id  ┃ name    ┃ score ┃ status   ┃
        ┡━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━╇━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━┩
        │ b'53fe8e4' │ 1738685136 │ True   │ 876 │ Charlie │ 27.74 │ pending  │
        │ b'91bde51' │ 1738685137 │ True   │ 11  │         │       │ approved │
        │ b'6617dfe' │ 1738685138 │        │     │ David   │       │          │
        │ b'f47ac93' │ 1738685139 │        │ 133 │         │       │          │
        │ b'038e524' │ 1738685140 │ False  │     │         │       │          │
        └────────────┴────────────┴────────┴─────┴─────────┴───────┴──────────┘

PR #752, opened 2025-02-13
`Application.run(timeout={N})` and sink `ListSink`

tim-quix:
## Addition: `RunTracker`

Added a new `RunTracker` class which manages the previous `app._running` variable and tracks new user stopping conditions handed to `Application.run()`. 

Made sure that `RunTracker` maximizes performance for the original/traditional infinite runtime approach (`timeout=0.0`) since `timeout` should really only be used for debugging/inspection and thus can sacrifice some performance. We should see very minimal performance degradation.

`RunTracker` is also a little "verbose" atm with expectations of being added to for count-based operations.

## Feature: `Application.run(timeout={float}, count={int})`

`Application.run()` now accepts new stop condition arguments; it will run the Application for either:

- `timeout`: the given amount of time in seconds after initial rebalance or recovery (default: `0.0` == infinite)
- `count`: number of processed messages by main SDF input topics (default `0` == infinite)

...or both (whichever happens first).

You can also now re-run an Application by just calling `.run()` again (with new `.run()` arguments each time if you prefer).

This will enable opportunities for interactive debugging and inspection with things like `ListSink` (below).

## Feature: `ListSink`

A new sink intended to paired with new Application.run() `timeout`.

Allows you to collect data and interact with it like a normal list. Example use:

```
    from quixstreams.sinks.core.list import ListSink

    app = Application(broker_address="localhost:9092")
    topic = app.topic("some-topic")
    list_sink = ListSink()  # sink will be a list-like object
    sdf = app.dataframe(topic=topic).sink(list_sink)
    app.run(timeout=10)  # collect data for 10 seconds
```

After 10 seconds:
```
    >>> print(list_sink) 
    [1, 2, 3]

    >>> list_sink[0]
    1
```

Note that records will continue to accumulate during subsequent runs; to delete old records, simply do `sink.clear()` or `sink[:] = []`

## Current limitations

Currently, source functionality is limited due to `multiprocessing` and `if __name__ == '__main__'` issues.

- You cannot run any source-based Application "interactively" (i.e. using ListSink to inspect values)
- You can only run a plain source (no SDF) with the `timeout` argument

## Miscellaneous 

- Made some other minor performance improvements to the runtime.

- Thanks to Remy for initial PoC :smiley:


quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+
+    def __init__(self):
+        """
+        Tracks the runtime status of an Application, along with managing variables
+        associated with stopping the app based on a timeout
+        """
+        self.running: bool = False
+        self._start_time: float = time.monotonic()
+        self._timeout: float = self._timeout_default
+        self._run_count: int = 0
```
gwaramadze: What if we use a real boolean, because this is how it is used in reality (`if not self._run_tracker.run_count`). Maybe something like `RunTracker.has_run` or very explicit `RunTracker.is_before_first_run`.

quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+        if stopper := self._stop_checker:
+            if stopper():
+                self.stop()
+
+    def start(self):
+        """
+        Called as part of app.run() to initialize/reset stop conditions.
+        """
+        self.running = True
+        self._run_count += 1
```
gwaramadze: As mentioned in another comment, we could rename this var and set `True` here which will stay forever `True`.

quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+        :return:
+        """
+        return self._run_count
+
+    def update_status(self):
+        """
+        Trigger stop if its condition is met.
+        """
+        # most optimal way to keep performance with typical operation (no timeout)
+        if stopper := self._stop_checker:
```
gwaramadze: I have the same problem to solve with the table print. I think the most optimal way is to maintain a little bool on the side like self._stop_checker_active. For reference [link to my PR](https://github.com/quixio/quix-streams/pull/740/files#diff-9aa9590ea8236677eb5c80be8cb6cfb3cbe05f1b5d33722cb18547b56eb02147R81).

In this case self._stop_checker will be `None`. I run a little timeit and the results are
```
Testing with var = None:
if var:             0.056 seconds
if var is None:     0.034 seconds

Testing with var = False:
if var:             0.032 seconds
```

More than that, don't use `:=` operator because it performs the assignment regardless of whether the condition evaluates to True or False.

If you disagree, we should at least change it to `if self._stop_checker is not None`

quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+
+    def start(self):
+        """
+        Called as part of app.run() to initialize/reset stop conditions.
+        """
+        self.running = True
+        self._run_count += 1
+        self.reset_tracking()
+        if self._timeout:
+            self._stop_checker = self._at_timeout
```
gwaramadze: I suggest to maintain an additional boolean to indicate the activity of a stop checker. Elaborated in a different comment.

```suggestion
            self._stop_checker = self._at_timeout
            self._stop_checker_active = True
```

quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+
+    def stop(self):
+        """
+        Called when Application is stopped, or self._stop_checker condition is met.
+        :return:
+        """
+        self.running = False
+        self._timeout = self._timeout_default
+
+    def set_stop_condition(self, timeout: Optional[float] = None):
```
gwaramadze: why not `timeout: float = 0.0`?

quixstreams/app.py
```diff
@@ -75,6 +75,78 @@ def __call__(
...
+        """
+        Called as part of app.run(), where user can pass their time limit.
+        """
+        if timeout := (timeout or self._timeout_default):
```
gwaramadze: I think you can take a shortcut here (if you decide to change the default to 0.0 in the method signature) or is there something I'm not seeing?
```suggestion
        if self._timeout := timeout:
```

quixstreams/app.py
```diff
@@ -712,7 +785,11 @@ def add_source(self, source: BaseSource, topic: Optional[Topic] = None) -> Topic
         )
         return topic
 
-    def run(self, dataframe: Optional[StreamingDataFrame] = None):
+    def run(
+        self,
+        dataframe: Optional[StreamingDataFrame] = None,
+        timeout: Optional[float] = None,
```
gwaramadze: Same here, I think we can keep the default float here and ditch the `None`
```suggestion
        timeout: float = 0.0,
```

quixstreams/app.py
```diff
@@ -794,36 +874,41 @@ def _run_dataframe(self):
...
             on_lost=self._on_lost,
         )
-        logger.info("Waiting for incoming messages")
-        # Start polling Kafka for messages and callbacks
-        self._running = True
-
-        # Initialize the checkpoint
-        self._processing_context.init_checkpoint()
 
+        # set refs for performance improvements
```
gwaramadze, line 861: nice!

quixstreams/app.py
tim-quix: Unfortunately, you can't set class attributes with a walrus operator, only variables.

In either case I changed the way this was setup so it doesn't need it now anyway =)

quixstreams/state/recovery.py
```diff
@@ -70,7 +70,7 @@ def offset(self) -> int:
 
         :return: changelog offset (int)
         """
-        offset = self._store_partition.get_changelog_offset() or 0
+        offset = self._store_partition.get_changelog_offset() or -1001
```
tim-quix: NOTE: This is a tiny bugfix. 

Recovery was not occurring if exactly 1 message was processed by a **new** consumer group -> state immediately deleted -> app is run again. 

This bug only occurs in this exact circumstance due to the way we handle our changelog offset numbers. Good ol' "off by one" errors!

ghost: Hey, good find!
Can you please cherry-pick it, put it to another PR, and add a test to verify the fix?

quixstreams/app.py
```diff
@@ -76,6 +86,288 @@ def __call__(
     ) -> TopicManager: ...
 
 
+class RunTracker:
```
ghost: Please move it to a separate module, `app.py` is already quite big

quixstreams/app.py
```diff
@@ -715,13 +1011,49 @@ def add_source(self, source: BaseSource, topic: Optional[Topic] = None) -> Topic
         )
         return topic
 
-    def run(self, dataframe: Optional[StreamingDataFrame] = None):
+    def run(
+        self,
+        dataframe: Optional[StreamingDataFrame] = None,
+        timeout: float = 0.0,
+        count: int = 0,
+        timeout_starts_on_first_message: bool = False,
```
gwaramadze: I favored starting the timer after the first message is consumed. I wasn't aware that the app may never stop if there are literally zero messages to consume.

We must not make this param part of our public API. If we can't detect and tap into the moment right **before** the first message is consumed, then we need to roll back to the initial version when the timer starts on assignment and accept the fact that there is an initial "gap with no data". This may be documented and can be fixed by a user with a bigger timeout.

quixstreams/app.py
```diff
@@ -746,6 +1087,17 @@ def run(self, dataframe: Optional[StreamingDataFrame] = None):
                 "the argument should be removed.",
                 FutureWarning,
             )
+        if not self._dataframe_registry.consumer_topics:
+            if count or timeout_starts_on_first_message:
```
gwaramadze: Did you mean `timeout` here?
```suggestion
            if timeout or timeout_starts_on_first_message:
```

tests/test_quixstreams/test_app.py
```diff
@@ -2778,3 +2780,410 @@ def revoke_partition(store, partition):
...
+        Count-based stopping does not track/count repartition (groupby) messages
+        It does however flush all downstream repartitions when it's the stopping trigger
+        """
+        app = app_factory(
+            auto_offset_reset="earliest",
+        )
+        input_topic = app.topic(
+            str(uuid.uuid4()),
+            value_deserializer="json",
+            value_serializer="json",
```
gwaramadze: "json" is the default, we can save a couple of lines here and there
```suggestion
```

quixstreams/app.py
```diff
@@ -76,6 +86,288 @@ def __call__(
...
+
+        self._start_time = 0.0
+        self._timeout = 0.0
+        self._start_timeout_after_first_msg = False
+        self._needs_assigning = True
+        self._is_recovering = False
+
+        self._max_count = 0
+        self._count = 0
+        self._repartition_stop_points = {}
```
gwaramadze: Looks like `stop` is a subset of `__init__`. I think it's ok to call `stop` from `__init__`. Perhaps with a comment to explain the weird semantics (as in "why would we call `stop` on class initialization?"). If it is a no-go for you, consider a private method like `_reset` that would be called from both `__init__` and `stop`.

quixstreams/app.py
```diff
@@ -76,6 +86,288 @@ def __call__(
...
+            if self._start_timeout_after_first_msg:
+                log_snippet = "(after consuming a non-recovery message)"
+            else:
+                log_snippet = "(after initial rebalance or recovery)"
+            logger.info(
+                f"APP RUN: timeout detected; "
+                f"Application will run for up to {timeout}s {log_snippet}"
+            )
+            self._stop_checker = self._at_timeout_func()
+        self._timeout = timeout
```
gwaramadze: I think it doesn't need to be re-set to 0.0
```suggestion
            self._timeout = timeout
```

quixstreams/app.py
```diff
@@ -76,6 +86,288 @@ def __call__(
...
+
+        if count:
+            if count < 0:
+                raise ValueError("run count must be >= 0")
+            logger.info(
+                f"APP RUN: count limit detected; "
+                f"Application will process up to {count} records"
+            )
+            self._stop_checker = self._at_count_func()
+        self._max_count = count
```
gwaramadze: Same here
```suggestion
            self._max_count = count
```

tests/test_quixstreams/test_app.py
```diff
@@ -2778,3 +2780,410 @@ def revoke_partition(store, partition):
...
+        app.run(count=2)
+
+        assert list_sink == values_in[:3]
+
+        with app.get_consumer() as consumer:
+            assert consumer.get_watermark_offsets(TopicPartition(output_topic.name, 0))[
+                1
+            ] == len(list_sink)
+
+    def test_timeout_after_recovery(
```
gwaramadze: I think there is some sneaky leakage here. When I try this locally `pytest -k test_timeout_after_recovery` passes but when I run `pytest -k TestRunTracker` this particular test always fails.

quixstreams/app.py
tim-quix: I agree that in reality it would be nice/more intuitive to start the timer like we originally planned (right when the first message is actually recognized)...here I was mostly trying to piggyback off the current implementation and avoid an added check in the non-stopper-based operations (keep everything as self-contained as possible).

It should be possible to add another bool check in the `process_message` (ex. `runner.needs_timeout_set`) if we wanted to move it to that. With `timeit` in python 3.9 on my laptop, I clocked a ref like that to take a total of ~.7s across 10mil iterations. 

Main downside is of course, this is added even to non-stopper applications unless we wanted to get really ugly, in which i'd definitely say we just ditch it.

Of course, this does not alleviate the running indefinite issue, but that behavior will be clearly documented...and hopefully a fairly obvious side-effect?

tim-quix: It might also be argued that it's not really needed at all. 

Tomas originally had issues because timeout was originally starting right before the poll loop began; I have since changed it to start after assignment or recovery, so it **should** be more accurate/reasonable now, but there are probably various circumstances where this is still not sufficient...hard to know when everything happens so smoothly and consistently locally.

@daniil-quix  also initially mentioned he felt the setting was maybe a little weird and if possible, should reduce the amount of possible arguments, which I also totally get. Not sure what his sentiments are now as that was only from a passive standup conversation about it.

Anywho, lemme know your thoughts!

quixstreams/app.py
tim-quix: It should be correct: `count` is not a valid stop option when it's just a plain source (no sdf). I have added a comment to clarify what check is going on here =)

tests/test_quixstreams/test_app.py
tim-quix: Maybe I'm just paranoid, but I set them in case we ever change defaults :see_no_evil: 

Seems extremely unlikely though :smiley:

quixstreams/app.py
tim-quix: I considered the same, but I swear IDE's get mad when you do that, and I figured it was probably because it was against PEP standards. I'll look into it and see, because I agree it's very tedious doing it this way and just invites mistakes and bugs.

quixstreams/app.py
tim-quix: good catch; should have changed this after I tweaked them to no longer use None.

tests/test_quixstreams/test_app.py
tim-quix: New band name: Sneaky Leakage

quixstreams/app.py
gwaramadze: > Tomas originally had issues because timeout was originally starting right before the poll loop began; I have since changed it to start after assignment or recovery, so it should be more accurate/reasonable now

Yes, I can see that now. Nice work! I'd vote to remove the argument and rely on automatic start after assignment/recovery.

quixstreams/app.py
tim-quix: I found a sanctioned way to do it without things like mypy or pycharm getting mad at it:

```
class MyClass:
    __slots__ = ("my_val",)
    
    my_val: int
    
    def __init__(self):
        self.set_my_val()
    
    def set_my_val(self):
        self.my_val = 1
```

docs/debugging.md
```diff
@@ -0,0 +1,223 @@
...
+new arguments!); it will simply pick up where it left off. 
+
+There is no need to do any manual cleanup when finished; each run cleans up after itself.
+
+> **NOTE:** You do not need to re-execute the entire Application setup, just `.run()`.
+
+If the Application seems to take longer than desired with getting initial messages or 
+you know there will be a delay, you can also specify it to wait to start counting the 
+timeout until after a first message is received by specifying 
+`timeout_starts_on_first_message=True`.
```
gwaramadze: This needs to go 😉

docs/debugging.md
```diff
@@ -0,0 +1,223 @@
...
+
+It collects data where it's used, which can then be interacted with like
+a list once the `Application` stops.
+
+#### Using ListSink
+
+To use a `ListSink`, simply store one in a separate variable and pass it
+like any other `Sink`:
+
+```
```
gwaramadze:
```suggestion
```python
```

docs/debugging.md
```diff
@@ -0,0 +1,223 @@
...
+  - Be sure to use it with `Application.run()` stopping conditions.
+
+- `ListSink` does not "refresh" itself per `.run()`; it collects data indefinitely. 
+  - You can remove the current data stored by doing either:
+
+      - `ListSink.clear()` 
+    
+        or...
+    
+      - `ListSink[:] = []`
```
gwaramadze: Let's simplify this bit, and I also propose to reuse the variable name from previous examples rather than the class name.
```suggestion
  - You can remove the current data stored by doing `list_sink.clear()`.
```

docs/debugging.md
tim-quix: DEATH IS ONLY THE BEGINNING!

gwaramadze (dismissed):

quixstreams/app.py
```diff
@@ -761,19 +822,21 @@ def _exception_handler(self, exc_type, exc_val, exc_tb):
...
+                f'auto_offset_reset="{self._config.auto_offset_reset}" '
+                f"commit_interval={self._config.commit_interval}s "
+                f"commit_every={self._config.commit_every} "
+                f'processing_guarantee="{self._config.processing_guarantee}"'
+            )
+            if self.is_quix_app:
+                self._quix_runtime_init()
 
-        self.setup_topics()
+            self.setup_topics()
```
ghost: What if you change your app between re-runs?
Like change topics, add stateful funcs, group-bys, etc

ghost: Same for the logs

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+        "_count",
+        "_repartition_stop_points",
+    )
+
+    # Here we establish all attrs that will be reset every `Application.run()` call
+    # This avoids various IDE and mypy complaints around not setting them in init.
+
+    running: bool
+    _stop_checker: Optional[Callable[[], bool]]
+    _has_stop_checker: bool
```
ghost: Is this really needed?
Doing `if self._has_stop_checker` and `if self._stop_checker is None` seems like the very same thing

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+        "_is_first_run",
+        "_processing_context",
+        "running",
+        "_stop_checker",
+        "_has_stop_checker",
+        "_needs_assigning",
+        "_is_recovering",
+        "_timeout",
+        "_start_time",
+        "_topic_manager",
```
ghost: Suggestion: consider passing source and repartion topics to the RunTracker directly instead of TopicManager

ghost: It should reduce the number of things the Tracker depends on

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+    def set_stop_condition(
+        self,
+        timeout: float = 0.0,
+        count: int = 0,
+    ):
+        """
+        Called as part of app.run(); this handles the users optional stop conditions.
+        """
+        if not (timeout or count):
+            logger.info("APP RUN: preparing to run until errors or manually stopped...")
```
ghost: I don't think we need to log that - it's the usual behavior

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+            return
+
+        logger.info("APP RUN: preparing to run with stop conditions...")
+        if timeout:
+            if timeout < 0.0:
+                raise ValueError("run timeout must be >= 0.0")
```
ghost: I suggest doing `timeout = max(timeout, 0)` instead
We can safely assume that negative timeout means "no timeout" and reduce the nesting

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+                f"APP RUN: timeout detected; "
+                f"Application will run for up to {timeout}s "
+                f"(after initial rebalance or recovery)"
+            )
+            self._stop_checker = self._at_timeout_func()
+            self._timeout = timeout
+
+        if count:
+            if count < 0:
+                raise ValueError("run count must be >= 0")
```
ghost: Same here: `count = max(count, 0)` to reduce the nesting

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+            logger.info(
+                f"APP RUN: count limit detected; "
+                f"Application will process up to {count} records"
+            )
```
ghost: Can we have only one log for the whole thing?
Feels weird to print three lines for that kind of info

docs/debugging.md
ghost, line 1: About `handle_recovery` and `handle_rebalance`: do we need to track these events?

What if we simplify it, start the timer, and count only when the first message is received somewhere inside `self._process_message`?
I understand that rebalance may happen somewhere in between, but that's rare when you run it locally.

quixstreams/runtracker.py
```diff
@@ -0,0 +1,290 @@
...
+    _has_stop_checker: bool
+
+    # timeout-specific attrs
+    _start_time: float
+    _timeout: float
+    _needs_assigning: bool
+    _is_recovering: bool
+
+    # count-specific attrs
+    last_consumed_tp: Optional[tuple]
```
ghost: Can you pls explain why it needs to track the last consumed TP?
I don't fully understand it.

quixstreams/sinks/core/list.py
```diff
@@ -0,0 +1,83 @@
+from collections import UserList
+from typing import Any
+
+from quixstreams.models import HeadersTuples
+
+from ..base import BaseSink
```
ghost:
```suggestion
from quixstreams.sinks.base import BaseSink
```

quixstreams/app.py
tim-quix: That seems out of scope with the intended use of this. I would assume you make no changes to your app, and you are just calling `app.run` for chunks of data under the same app/configuration. I'd argue making app-level changes is a different app and should thus be fully re-executed.

quixstreams/runtracker.py
tim-quix: This is one of those "for performance" tweaks. It is faster to check a bool like this.

One thing we could do is move the bool check outside the class?

quixstreams/runtracker.py
tim-quix: I like it :+1:

docs/debugging.md
tim-quix: That is definitely an option, and it was actually discussed in earlier review comments.

The main reason it is not done that way is to minimize the impact on typical runtime performance, which in this case we'll have to check a flag on the run tracker each pass. Small cost, but it is a cost none-the-less.

If we value reducing the size of the RunTracker more, then I'll do it :+1:

quixstreams/runtracker.py
tim-quix: This is so we do not count repartition topic reads as part of the count, only "main" topics.

quixstreams/runtracker.py
tim-quix: As extra clarification, I mean it's faster for the case where no stopper is set (typical running)

quixstreams/runtracker.py
tim-quix: Could do that, though they'll have to be set as another call (or direct attr assigns) since we need to wait until the topics have been fully set/populated first from the user (as `RunTracker` is init in `App` `__init__` like most of the other objects).

Could also instead init `RunTracker` with the first `App.run()` call.

docs/debugging.md
tim-quix: Another issue with that approach is if you never consume a message, the timer never starts.

quixstreams/app.py
ghost: I think that avoiding this check will make the code easier to follow.
I don't see why we need it really

quixstreams/runtracker.py
ghost: I believe the difference between `if self._something` and `if self._something is None` is negligible :)

quixstreams/app.py
```diff
@@ -889,6 +962,7 @@ def _process_message(self, dataframe_composed):
         self._processing_context.store_offset(
             topic=topic_name, partition=partition, offset=offset
         )
+        self._run_tracker.last_consumed_tp = (topic_name, partition)
```
ghost: Let's replace it with some method call instead of just setting the attribute to better incapsulate the logic

quixstreams/runtracker.py
tim-quix: It is def pretty small. I'll change it :+1:

quixstreams/app.py
tim-quix: Mostly for avoiding superflous checks and logging, but I can remove it.

tim-quix:
Closing this, merged branch #780 instead

PR #759, opened 2025-02-21
Fix default state dir for Quix apps

gwaramadze:
### Problem
After the latest update, Quix Cloud changed the default app placement from `/app` to `/app/<something>`.

Because of that, the default state dir in Quix now points to `/app/<something>/state` instead of `/app/state` (mounted PVC).

### Solution
Set the default path for Quix apps to `/app/state` instead of the current `state/` (relative).

Users should be able to override it, so we need to change the defaults for the whole Application.

### DIFF
--- a/quixstreams/app.py
+++ b/quixstreams/app.py
@@ -63,7 +63,7 @@ _default_producer_extra_config = {"enable.idempotence": True}
 consumer_extra_config_overrides = {"partition.assignment.strategy": "range"}
 
 _default_max_poll_interval_ms = 300000
-_default_state_dir = Path("state")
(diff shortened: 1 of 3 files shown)


quixstreams/app.py
```diff
@@ -62,6 +63,7 @@
 consumer_extra_config_overrides = {"partition.assignment.strategy": "range"}
 
 _default_max_poll_interval_ms = 300000
+_default_state_dir = Path("state")
```
ghost: Actually, you can even do this:
```
_default_state_dir = Path("/app/state") if is_quix_deployment() else Path("state")
```

ghost: and keep the other code intact

ghost (approved):

PR #763, opened 2025-02-24
Create topics eagerly the moment they are defined

ghost:
We must know the number of topic partitions in advance to join and merge topics correctly.  
Previously, the topics were created and validated only when the Application started.

This PR makes them eagerly created as soon as they are defined, so the app has information about partitions before it starts and can use it to validate the `StreamingDataFrames`.

In the future, it will also enable optimizations like bypassing repartition topics when the `StreamingDataFrame.group_by()` is called on a one-partition topic and more.

## Changes
- Refactored code related to topics, especially `QuixTopicManager` to accommodate the changes
- Separated the `Topic.config` into `create_config` and `real_config`:
  -  `create_config` is set by the user or derived from the existing topics for internal topics. It represents the desired configuration but can differ from what's in the broker.
  - `broker_config` is retrieved from the broker and represents the actual topic config at the moment.
- Topics are now created and/or validated when they are defined. 
   - E.g. `app.topic()` will try to create a topic if `Application(auto_create_topics=True)`. It will validate and retrieve its config regardless of the `auto_create_topics` setting, too.
  
- Fixed some cases when topics were created even when `auto_create_topics=False`

## Notes
- Previously, topics were created in a batch, so creating them one by one may slightly increase the app startup time depending on the distance to the broker and the number of topics.
- Topics retrieve their actual configs only before the app starts.  If the config changes afterward, the app won't learn about it until restart.


quixstreams/models/topics/topic.py
```diff
@@ -122,39 +123,43 @@ def __init__(
...
-        `AdminClient.create_topic()`.
+        return self._create_config
 
-        :return: confluent_kafka `NewTopic`s
+    @create_config.setter
+    def create_config(self, config: Optional[TopicConfig]):
+        self._create_config = config
+
+    @property
+    def real_config(self) -> TopicConfig:
```
ghost: I'm not really a fan of the name. Should we keep `config` or maybe `broker_config` ?

quixstreams/platforms/quix/topic_manager.py
```diff
@@ -56,31 +61,77 @@ def __init__(
...
+                    f'Topic "{topic.name}" not found on the broker'
+                )
+            raise
+        quix_topic = self._quix_config_builder.convert_topic_response(quix_topic_info)
+        return quix_topic
+
     def _finalize_topic(self, topic: Topic) -> Topic:
         """
         Returns a Topic object with the true topic name by attempting by finding or
         creating it in Quix Cloud and using the returned ID (the true topic name).
```
ghost, line 81: I believe this is no longer true since we are using the `__quix_topic_name__` and not the `id`.

quixstreams/platforms/quix/topic_manager.py
```diff
@@ -56,31 +61,77 @@ def __init__(
...
 
     def _create_topics(
         self, topics: List[Topic], timeout: float, create_timeout: float
     ):
-        """
-        Because we create topics immediately upon Topic generation for Quix Cloud,
-        this method simply confirms topics all have a "Ready" status.
-        """
-        self._quix_config_builder.wait_for_topic_ready_statuses(topics)
+        topic = topics[0]
```
ghost: We should either accept one topic or loop over all of them here.

quixstreams/platforms/quix/config.py
```diff
@@ -198,14 +197,14 @@ def convert_topic_response(
...
-                num_partitions=topic_config["partitions"],
-                replication_factor=topic_config["replicationFactor"],
-                extra_config=extra_config,
-            ),
+        config = TopicConfig(
+            num_partitions=topic_config["partitions"],
+            replication_factor=topic_config["replicationFactor"],
+            extra_config=extra_config,
         )
+        topic = Topic(name=api_response["id"], create_config=config)
```
ghost, line 205: Does it make sense to set the `create_config` here ? If we do it shouldn't be the returned config but the config passed by the user.

quixstreams/platforms/quix/config.py
```diff
@@ -372,6 +370,10 @@ def create_topic(self, topic: Topic, timeout: Optional[float] = None):
         ret_bytes = cfg.extra_config.get("retention.bytes")
 
         # an exception is raised (status code) if topic is not created successfully
+        logger.info(
+            f'Creating a topic "{topic.name}" '
```
ghost:
```suggestion
            f'Creating topic "{topic.name}" '
```

quixstreams/platforms/quix/config.py
```diff
@@ -459,18 +461,18 @@ def wait_for_topic_ready_statuses(
                 f"topics: {[t['name'] for t in topic_resps if t['id'] in topic_ids]}"
             )
 
-    def get_topic(self, topic_name: str, timeout: Optional[float] = None) -> dict:
+    def get_topic(self, topic: Topic, timeout: Optional[float] = None) -> dict:
         """
         return the topic ID (the actual cluster topic name) if it exists, else raise
 
-        :param topic_name: name of the topic
+        :param topic: a Topic to get
```
ghost:
```suggestion
        :param topic: The Topic to get
```

quixstreams/platforms/quix/config.py
ghost: Previously, it was just setting a config, so I'm trying to keep the old behavior here without changing the API too much since it's out of scope.

quixstreams/platforms/quix/topic_manager.py
ghost: It's still true, because "id" is still used as a topic name (see the `convert_topic_response` method)
The `__quix_topic_name__` is just saved to the mapping inside QuixTopicManager to be used later

quixstreams/platforms/quix/topic_manager.py
ghost: Agree 👍

ghost (approved):

PR #769, opened 2025-03-03
Bugfix for recovering from exactly 1 changelog message

tim-quix:
Fixes a bug around recovering after a single changelog was successfully produced, and then state deleted immediately after.


quixstreams/state/recovery.py
```diff
@@ -70,7 +70,7 @@ def offset(self) -> int:
 
         :return: changelog offset (int)
         """
-        offset = self._store_partition.get_changelog_offset() or 0
+        offset = self._store_partition.get_changelog_offset() or -1001
         if not self._initial_offset:
```
ghost: Should it do `is None` instead of `not` check here then?

tim-quix: Yeah prob a good idea to change that too :+1:

ghost (approved):

PR #771, opened 2025-03-04
Increase producer poll timeout and retries

ghost:


quixstreams/kafka/producer.py
```diff
@@ -30,6 +30,9 @@
     KafkaError._DESTROY,  # noqa: SLF001
 )
 
+PRODUCER_POLL_TIMEOUT = 30.0
+PRODUCER_ON_ERROR_RETRIES = 10
```
daniil-quix, line 34: Suggestion here: let's change it to be "timeout=5.0" and "retries=60"
I realized that with the timeout=30, the app can get stuck for these 30 seconds even if you're trying to stop it.
It works similarly with other timeouts in confluent_kafka

ghost (approved):

PR #772, opened 2025-03-05
windows: Split Aggregations and collectors is own classes

ghost:
in preparation for multiple aggregations and collectors support

### DIFF
--- a/quixstreams/dataframe/windows/aggregations.py
+++ b/quixstreams/dataframe/windows/aggregations.py
@@ -18,6 +18,11 @@ class Aggregator(ABC):
     Base class for window aggregation.
 
     Subclass it to implement custom aggregations.
+
+    An Aggregator reduce incoming items into a single value or group of values. When the window
+    is closed the aggregator produce a result based on the reduced value.
+
+    To store all incoming items without reducing them use a `Collector`.
     """
 
     @abstractmethod
@@ -52,11 +57,11 @@ class ROOT:
     pass
 
 
-COLUMN: TypeAlias = Union[Hashable, type[ROOT]]
+Column: TypeAlias = Union[Hashable, type[ROOT]]
 
 
 class Sum(Aggregator):
-    def __init__(self, column: COLUMN = ROOT) -> None:
+    def __init__(self, column: Column = ROOT) -> None:
         self.column = column
 
     def initialize(self) -> int:
@@ -83,7 +88,7 @@ class Count(Aggregator):
 
 
 class Mean(Aggregator):
-    def __init__(self, column: COLUMN = ROOT) -> None:
+    def __init__(self, column: Column = ROOT) -> None:
         self.column = column
 
     def initialize(self) -> tuple[float, int]:
@@ -123,7 +128,7 @@ class Reduce(Aggregator, Generic[R]):
 
 
 class Max(Aggregator):
-    def __init__(self, column: COLUMN = ROOT) -> None:
+    def __init__(self, column: Column = ROOT) -> None:
         self.column = column
 
     def initialize(self) -> None:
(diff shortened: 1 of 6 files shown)


quixstreams/dataframe/windows/aggregations.py
```diff
@@ -0,0 +1,179 @@
...
+V = TypeVar("V", int, float)
+
+ROOT = object()
+
+
+class Sum(Aggregation):
+    def __init__(self, column: Any = ROOT) -> None:
+        self.column = column
+
+    def start(self) -> int:
```
ghost: What about `initialize`?
Or `init`?

"Start" makes it sound like there's some background work going on :)

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        This method is triggered when a window is closed.
+        It should return the final aggregation result.
+        """
+        ...
+
+
+V = TypeVar("V", int, float)
+
+ROOT = object()
```
ghost: Why does it need to be `ROOT` vs `None`?

ghost: `None` is a valid dict key. Weird but someone might be doing that

quixstreams/dataframe/windows/aggregations.py
ghost: Not a fan of `init` due to `__init__` existing. I'll go with the full form

quixstreams/dataframe/windows/aggregations.py
ghost: Ok, fair enough

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -0,0 +1,179 @@
+import sys
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Generic, Iterable, Optional, TypeVar, Union
+
+
+class Aggregation(ABC):
```
ghost: 1. Maybe do `Aggregator` vs `Aggregation`? There's a `Collector` nearby.
2. Can you please document the difference between `Aggregator` and `Collector` here?  Aggregators clearly assume that the aggregation is reduced to a single key

ghost: It's important since we now provide base classes allowing users to implement custom aggregations.

ghost: I was planning to do the docs with the complete multi-aggregation MR. Since those classes can't be used right now

ghost: I'll rename to Aggregator, I hesitated for a while on it

quixstreams/dataframe/windows/base.py
```diff
@@ -51,13 +52,26 @@ def __init__(
...
 
+        self._aggregations = aggregations
+        self._aggregate = len(aggregations) > 0
+
+        self._collectors = collectors
+        self._collect = len(collectors) > 0
+
+        if not self._collect and not self._aggregate:
+            raise ValueError("At least one aggregation or collector must be defined")
+        elif sum((len(collectors), len(aggregations))) > 1:
```
ghost:
```suggestion
        elif len(collectors) + len(aggregations) > 1:
```

quixstreams/dataframe/windows/base.py
```diff
@@ -51,13 +52,26 @@ def __init__(
         self,
         name: str,
         dataframe: "StreamingDataFrame",
+        aggregations: dict[str, Aggregator],
```
ghost: "aggregators" maybe?

quixstreams/dataframe/windows/count_based.py
```diff
@@ -35,22 +33,21 @@ class CountWindow(Window):
...
         self,
-        name: str,
         count: int,
-        aggregate_func: WindowAggregateFunc,
-        aggregate_default: Any,
-        aggregate_collection: bool,
+        name: str,
         dataframe: "StreamingDataFrame",
-        merge_func: Optional[WindowMergeFunc] = None,
+        aggregations: dict[str, Aggregator],
```
ghost: Same here, "aggregators"

ghost (approved):

PR #774, opened 2025-03-05
Refactor recovery to support Stores belonging to multiple topics

ghost:
# Main changes
## Updated the format of the `"__processed_offsets__"` headers in the changelog messages
- The header name is now `__processed_offsets__` (previously `__processed_offset__`)
- The header contains JSON encoded dict with topic offsets for this partition (e.g. `{"topic": 123}`)
- During recovery, we check if all offsets in the dictionary are below the committed offsets for the same TPs.  Currently, there's only one offset possible in the dict.

**Upgrade considerations:** 
- **ALOS mode:** the last checkpoint must be committed successfully before upgrading to the new version.  
Otherwise, the app won't be able to filter out over-produced changelog messages because it won't recognize the old header format used previously.
- **EOS mode:** works fine as before.

## Removed the warning about processed offset being behind the committed offset.
Long before implementing changelog recovery, we added a mechanism to check if the processed offset in the state is aligned with the committed offset as a best-effort to signal state inconsistency.
Since changelog topics are widely used now, we can remove the check without any harm, and also stop storing the processed offsets directly in the State.

## Moved Recovery checks and validations to `RecoveryPartition`
Previously, `StorePartition` subclasses verified the changelog message format and checked the processed and committed offsets.
It was a repeatable code, and it now lives in `RecoveryPartition`.

Also, I added the `StorePartition.write_changelog_offsets()` method to roll the changelog offsets for the skipped overproduced changelogs.


quixstreams/state/rocksdb/partition.py
```diff
@@ -63,75 +61,40 @@ def __init__(
...
+    ):
+        cf_handle = self.get_column_family_handle(cf_name)
+        batch = WriteBatch(raw_mode=True)
+        if value is None:
+            batch.delete(key, cf_handle)
+        else:
+            batch.put(key, value, cf_handle)
+
+        # Update the changelog offset and flush outstanding writes.
         batch.put(
```
tim-quix: Looks like you might have intended to convert this to `self._update_changelog_offset(batch=batch, offset=offset)`?

quixstreams/state/base/partition.py
```diff
@@ -45,42 +40,34 @@ def __init__(
...
-        processed_offset: Optional[int],
-        committed_offset: int,
-    ): ...
-
-    @abstractmethod
-    def get_processed_offset(self) -> Optional[int]:
+    def get_changelog_offset(self) -> Optional[int]:
         """
-        Get last processed offset for the given partition
+        Get offset that the changelog is up-to-date with.
```
tim-quix: maybe a little nitpicky or subjective, but I think this verbiage might match the behavior a little closer (feel free to tweak it of course :smiley: ): 

"get the changelog offset that the state is up-to-date with"

quixstreams/app.py
```diff
@@ -923,40 +924,26 @@ def _on_assign(self, _, topic_partitions: List[TopicPartition]):
             non_changelog_tps = [
                 tp for tp in topic_partitions if tp.topic in non_changelog_topics
             ]
+            committed_tps = self._consumer.committed(
+                partitions=non_changelog_tps, timeout=30
+            )
+            committed_offsets: dict[int, dict[str, int]] = defaultdict(dict)
```
tim-quix, line 930: I can see why you're generating `committed_offsets` here (we need to do this call at some point, and here we do it before we're deep in any sort of setup), but also we are doing what's mostly a passthrough with it across several calls:

_state_manager.on_partition_assign() -> SM._recovery_manager.assign_partition() -> RM._generate_recovery_partition() -> RecoveryPartition()

Maybe we could cut down on all the passthrough/ simplify it by refactoring this portion a bit?

tim-quix (commented):
Did an initial pass, might look again tomorrow. Only one thing that feels like it could maybe use a second glance so far.

Either way, it was easy to follow and I like that it's actually a bit more simplified now! Really nice :+1:

quixstreams/state/rocksdb/partition.py
ghost: Yes, I missed that part 👍

quixstreams/state/base/partition.py
ghost:
```suggestion
        Get the changelog offset that the state is up-to-date with.
```

quixstreams/app.py
ghost: I don't like the long chain of calls either, but here we can do it once for the whole assignment.

We could change the `RecoveryManager` part a bit and merge a couple of func calls, but that would be out of scope of this PR.

quixstreams/state/base/partition.py
ghost: Updated 👍

quixstreams/state/base/transaction.py
```diff
@@ -397,33 +397,32 @@ def prepare(self, processed_offset: Optional[int]):
         If changelog is disabled for this application, no updates will be produced
         to the changelog topic.
 
-        :param processed_offset: the offset of the latest processed message
+        :param processed_offsets: the dict with <topic: offset> of the latest processed message
         """
 
         try:
-            self._prepare(processed_offset=processed_offset)
+            self._prepare(processed_offsets=processed_offsets)
```
tim-quix, line 404: Minor consideration:

Unless I'm mistaken, there's nothing to really prepare if you have no processed offsets (or changelog_producer for that matter), so maybe you can forego calling `_prepare` unless those conditions are met. 

Can maybe even remove the "Optional" type of processed_offsets too for tighter typing (at least for the `_prepare` level, but also could do it for `prepare` as well since only the `StatefulSource` uses None, which could be replaced with a throwaway stored dict if you wanted to).

quixstreams/state/base/transaction.py
```diff
@@ -463,18 +461,17 @@ def flush(
             not flush ANY data to the database including the offset to optimize
             I/O.
 
-        :param processed_offset: offset of the last processed message, optional.
         :param changelog_offset: offset of the last produced changelog message,
             optional.
         """
         try:
-            self._flush(processed_offset, changelog_offset)
+            self._flush(changelog_offset)
```
tim-quix, line 468: similar considerations here as my comment around `_prepare` input/behavior.

tim-quix (approved):
couple minor comments...but more so just considerations. Feel free to take a look and ping me to approve after if you decide to change anything :+1:

quixstreams/state/base/transaction.py
ghost: I see the point, but I'd rather touch it in the next iteration since the logic around `processed_offsets` being `None` doesn't change

quixstreams/state/base/transaction.py
ghost: Same, good point, but out of scope of this PR

PR #785, opened 2025-03-11
windows: Support multi-aggregation windows

ghost:


docs/aggregations.md
```diff
@@ -0,0 +1,412 @@
...
+
+Currently, [windows](windowing.md) support the following aggregation and collection functions:
+
+- [`Count()`](api-reference/quixstreams.md#count) - to count the number of values within a window 
+- [`Min()`](api-reference/quixstreams.md#min) - to get a minimum value within a window
+- [`Max()`](api-reference/quixstreams.md#max) - to get a maximum value within a window
+- [`Mean()`](api-reference/quixstreams.md#mean) - to get a mean value within a window 
+- [`Sum()`](api-reference/quixstreams.md#sum) - to sum values within a window 
+- [`Collect()`](api-reference/quixstreams.md#collect) - to collect all values within a window into a list
+
```
gwaramadze, line 10: Is `Reduce()` not in the list because it is deprecated?

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -196,12 +287,31 @@ def result(self, items: Iterable[I]) -> Any:
...
+    :param column: The column to collect. Use `None` to collect the whole message.
+        Default - `None`
+    """
+
+    def __init__(self, column: Optional[str] = None) -> None:
         self._column = column
 
     @property
-    def column(self) -> Column:
+    def column(self) -> Optional[str]:
```
gwaramadze: Do we need it as a `property`? Aggregators don't have it.

docs/aggregations.md
```diff
@@ -0,0 +1,412 @@
...
+```
+
+By default a collector will collect the full message. Sometime that's unecessary if you are only interested in a specific column. A custom collector can override the [`column`](api-reference/quixstreams.md#basecollectorcolumn) property to tell quixstreams which column it needs.
+
+An implementation of `ReversedCollect` taking an optinal `column` parameter
+
+```python
+class ReversedCollect(Collector):
+    def __init__(self, column = None)
+        self._column = column
```
gwaramadze: Why can't this be a part of a built-in Collector? Passing the column seems like a feature generic enough and will be on par with what Aggregator can do

docs/aggregations.md
```diff
@@ -0,0 +1,412 @@
...
+#   'start': <window start>, 
+#   'end': <window end>, 
+#   'min_temp': 1
+#   'max_temp': 999
+#   'avg_temp': 34.32
+#   'total_events': 999
+# }
+```
+
+Here is how you can do that with `reduce()`:
```
gwaramadze: Do we need this section if reduce is deprecated?

docs/aggregations.md
```diff
@@ -0,0 +1,412 @@
...
+**Example**:
+
+Assume you receive the temperature data from the sensor, and you need to calculate these aggregates for each 10-minute tumbling window:
+
+- min temperature
+- max temperature
+- total count of events
+- average temperature
+
+Here is how you can do that with multiple aggregations
```
gwaramadze: I think you can get rid of this line.
```suggestion
```

gwaramadze (commented):
first round

docs/windowing.md
```diff
@@ -272,26 +251,14 @@ sdf = app.dataframe(...)
...
     .tumbling_count_window(count=3)
 
     # Specify the "collect" aggregate function
-    .collect()
+    .agg(data=Collect())
 
     # Emit updates once the window is closed
     .final()
-
-    # Unwrap the aggregated result to match the expected output format
```
gwaramadze, line 287: You need to update the expected output on line 231

docs/windowing.md
```diff
@@ -1055,6 +727,55 @@ sdf = sdf.tumbling_window(timedelta(seconds=10)).sum().final(closing_strategy="p
 # (key="C", value={"start": 0, "end": 10000, "value": 3})
 ```
 
+## Transforming the result of a windowed aggregation
+Windowed aggregations return aggregated results in the following format/schema:
+
+```python
+{"start": <window start ms>, "end": <window end ms>, <aggregated_result_colum>: <aggregated value>}
```
gwaramadze:
```suggestion
{"start": <window start ms>, "end": <window end ms>, <aggregated result colum>: <aggregated value>}
```

docs/windowing.md
```diff
@@ -1089,9 +809,11 @@ Quix Streams handles some of the situations, like:
 
 - Updating window type (e.g. from tumbling to hopping)
 - Updating window period or step 
-- Updating an aggregation function (except the `reduce()`)
+- Adding/Removing/Updating an aggregation function (except the `reduce()`)
```
gwaramadze: should it be `reduce()` or `Reduce()` ?

docs/aggregations.md
```diff
@@ -0,0 +1,412 @@
...
+Currently, [windows](windowing.md) support the following aggregation and collection functions:
+
+- [`Count()`](api-reference/quixstreams.md#count) - to count the number of values within a window 
+- [`Min()`](api-reference/quixstreams.md#min) - to get a minimum value within a window
+- [`Max()`](api-reference/quixstreams.md#max) - to get a maximum value within a window
+- [`Mean()`](api-reference/quixstreams.md#mean) - to get a mean value within a window 
+- [`Sum()`](api-reference/quixstreams.md#sum) - to sum values within a window 
+- [`Collect()`](api-reference/quixstreams.md#collect) - to collect all values within a window into a list
+
+You can also create your own custom aggregator and collector. We will go over each ot them in more detail below.
```
gwaramadze:
```suggestion
You can also create your own custom aggregator and collector. We will go over each of them in more detail below.
```

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -3,17 +3,27 @@
...
-from typing_extensions import TypeAlias
+__all__ = [
+    "Collect",
+    "Count",
+    "Max",
+    "Mean",
+    "Min",
+    "Reduce",
+    "Sum",
+]
```
gwaramadze, line 24: `Aggregator` and `Collector` need to be in `__all__`

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -25,117 +35,159 @@ class Aggregator(ABC):
...
     @abstractmethod
-    def initialize(self) -> Any:
+    def state_suffix(self) -> str:
+        """
+        The state suffix used to store the aggregation state in the window.
+
+        The complete state key is built using the result column name and this suffix. If any of these
+        values change the state key will change and the aggregation state restart from zero.
+
+        Aggregations should change the state suffix when there parameters change to avoid
```
gwaramadze:
```suggestion
        Aggregations should change the state suffix when their parameters change to avoid
```

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -25,117 +35,159 @@ class Aggregator(ABC):
...
 
-class ROOT:
-    pass
+    @property
+    def state_suffix(self) -> str:
+        return self.__class__.__name__
 
 
-Column: TypeAlias = Union[Hashable, type[ROOT]]
+class Count(Aggregator):
```
gwaramadze, line 98: We were discussing this with @daniil-quix yesterday. This should also accept a column name and there is a need start checking and counting only non-missing values.

Check out what `pandas` would do:
```
In [37]: df = pd.DataFrame({"a": [1,2,3], "b": [10, None, pd.NA]})

In [38]: df
Out[38]:
   a     b
0  1    10
1  2  None
2  3  <NA>

In [39]: df.a.count()
Out[39]: 3

In [40]: df.b.count()
Out[40]: 1
```

For us, in this case missing data is both missing column and missing value (value is None).

I could do it in a subsequent PR, but it feels like this belongs to this initiative: since we enable to choose columns for all the other aggregations, this one should not lag behind.

Additionally, maybe thanks to this we can use a single base class since the `state_suffix` will be the same for all 5?

gwaramadze (changes requested):
Posting a second round with one bigger blocker.

quixstreams/dataframe/windows/aggregations.py
gwaramadze: Same goes for root count, if someone sends a plain null over the wire, we should not count it. We were doing it wrong all along 😅

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -25,117 +35,159 @@ class Aggregator(ABC):
...
-        if self.column is ROOT:
+        if self.column is None:
             return old_sum + new, old_count + 1
         return old_sum + new[self.column], old_count + 1
 
-    def result(self, value: tuple[Union[int, float], int]) -> float:
+    def result(self, value: tuple[Union[int, float], int]) -> Optional[float]:
         sum_, count_ = value
+        if count_ == 0:
+            return None
```
gwaramadze: Is this supposed to cover the case when the `Mean()` aggregator is added to an existing aggregation?

I planned to cover the case of missing data with this
```python
        if sum_ == 0.0:
            return 0.0
```
Not sure what we can do about it, we will need to settle on one, probably `None` 🤔

quixstreams/dataframe/windows/aggregations.py
ovv: Yes, it's needed to tell the window which column to collect so we don't always collect the whole message. Reworked it to make it part of `Collector` directly (and same with `Aggregator`)

docs/windowing.md
ovv: ah yeah I need to update all those example

quixstreams/dataframe/windows/base.py
```diff
@@ -196,6 +183,201 @@ def window_callback(
...
+
+        super().__init__(**kwargs)
+
+    @property
+    def aggregate(self) -> bool:
+        return self._collector is None
+
+    @property
+    def collect(self) -> bool:
+        return self._aggregator is None
```
gwaramadze: I think this will be more accurate
```suggestion
    @property
    def aggregate(self) -> bool:
        return self._aggregator is not None

    @property
    def collect(self) -> bool:
        return self_collector is not None
```

quixstreams/dataframe/windows/count_based.py
```diff
@@ -16,12 +16,19 @@
...
     count: int
     start: int
     end: int
-    value: Any
+
+    # Can be None for single aggregation windows not migrated
+    aggregations: Union[Any, dict[str, Any]]
+    collection_start_id: int
+
+    # value: Any deprecated. Only used in single aggregation windows for both collection id tracking and aggregation
```
gwaramadze, line 31: Probably should be removed.
```suggestion
```

quixstreams/dataframe/windows/aggregations.py
ovv: Yes it's used if no data was aggregated. We decided on `None` since a mean of 0 is fine and doesn't show that no data was received.

quixstreams/dataframe/windows/count_based.py
ovv: No it's there on purpose. Pre-migration it's there.

I don't set it in the type as you'd need to specify it when creating a `CountWindowData` but old `CountWindowData` that aren't migrated to the new structure have it.

docs/aggregations.md
ovv: I think it helps in showing how custom aggregators and multi aggregation simplify the code vs reduce

docs/aggregations.md
```diff
@@ -0,0 +1,378 @@
...
+)
+
+# Output:
+# {
+#   'start': <window start>, 
+#   'end': <window end>, 
+#   'min_temp': 1
+#   'max_temp': 999
+#   'avg_temp': 34.32
+#   'total_events': 999
```
gwaramadze:
```suggestion
#   'min_temp': 1,
#   'max_temp': 999,
#   'avg_temp': 34.32,
#   'total_events': 999,
```

docs/aggregations.md
```diff
@@ -0,0 +1,378 @@
...
+# }
+```
+
+Here is how you can do that with `Reduce()`:
```
gwaramadze: I think we should promote a custom aggregation instead of `Reduce`

quixstreams/dataframe/windows/definitions.py
```diff
@@ -179,7 +199,32 @@ def collect(self) -> "Window":
...
             func_name="collect",
-            collectors={"value": Collect(column=ROOT)},
+            collectors={"value": Collect(column=None)},
+        )
+
+    def agg(
+        self, **operations: dict[str, Union[BaseAggregator, BaseCollector]]
+    ) -> "Window":
+        if "start" in operations or "end" in operations:
+            raise ValueError("`start` and `end` are reserved keywords")
```
gwaramadze: let's explain that those are reserved for the window start/end timestamps

tests/test_quixstreams/test_dataframe/test_windows/test_aggregations.py
```diff
@@ -10,62 +11,92 @@
...
-    assert aggregator.result(old) == expected
+    @pytest.mark.parametrize(
+        "aggregation, result",
+        [
+            (Count(), "Count"),
+            (Sum(), "Sum"),
+            (Mean(), "Mean"),
+            (Max(), "Max"),
+            (Min(), "Min"),
+            (Sum("value"), "Sum/value"),
```
gwaramadze, line 82:
```suggestion
            (Count("value"), "Count/value"),
            (Sum("value"), "Sum/value"),
```

docs/aggregations.md
ovv: I put 3 examples. I want to keep one about Reduce so people still using it have something in the documentation.

gwaramadze (dismissed):

quixstreams/dataframe/windows/definitions.py
```diff
@@ -179,7 +199,34 @@ def collect(self) -> "Window":
...
+        collectors: dict[str, BaseCollector] = {}
+
+        for column, op in operations.items():
+            if isinstance(op, BaseAggregator):
+                aggregators[column] = op
+            elif isinstance(op, BaseCollector):
+                collectors[column] = op
+            else:
+                raise TypeError(
+                    f"operation `{column}:{op}` must be a BaseAggregator or a BaseCollector"
```
ghost:
```suggestion
                    f"operation `{column}:{op}` must be either BaseAggregator or BaseCollector"
```

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -25,76 +39,125 @@ class Aggregator(ABC):
...
-    def initialize(self) -> Any:
+    def state_suffix(self) -> str:
+        """
+        The state suffix used to store the aggregation state in the window.
+
+        The complete state key is built using the result column name and this suffix. If any of these
+        values change the state key will change and the aggregation state restart from zero.
+
+        Aggregations should change the state suffix when their parameters change to avoid
+        conflicts with previous state values.
```
ghost:
```suggestion
        The state suffix is used to store the aggregation state in the window.

        The complete state key is built using the result column name and this suffix.  
        If these values change, the state key will also change, and the aggregation state will restart from zero.

        Aggregations should change the state suffix when their parameters change to avoid
        conflicts with previous state values.
```

quixstreams/dataframe/windows/base.py
```diff
@@ -196,6 +183,201 @@ def window_callback(
...
+            return self._aggregator.initialize()
+        return None
+
+    def _aggregate_value(self, state_value: Any, value: Any, timestamp: int) -> Any:
+        if self._aggregator:
+            return self._aggregator.agg(state_value, value, timestamp)
+        return None
+
+    def _collect_value(self, value: Any):
+        # Single aggregation collect() always store the full message
```
ghost:
```suggestion
        # Single aggregation collect() always stores the full message
```

quixstreams/dataframe/windows/sliding.py
```diff
@@ -257,46 +265,40 @@ def process_window(
...
+    def _expired_windows(self, state, max_expired_window_start, collect):
+        for window in state.expire_windows(
+            max_start_time=max_expired_window_start,
+            delete=False,
+            collect=collect,
+            end_inclusive=True,
+        ):
+            (start, end), (max_timestamp, aggregated), collected, key = window
+            if end == max_timestamp:
+                yield (key, self._results(aggregated, collected, start, end))
```
ghost:
```suggestion
                yield key, self._results(aggregated, collected, start, end)
```

quixstreams/dataframe/windows/time_based.py
```diff
@@ -218,26 +213,14 @@ def expire_by_partition(
...
             count += 1
-            yield (
-                key,
-                WindowResult(
-                    start=window_start,
-                    end=window_end,
-                    value=value,
-                ),
-            )
+            yield (key, self._results(aggregated, collected, window_start, window_end))
```
```suggestion
            yield key, self._results(aggregated, collected, window_start, window_end)
```

ghost (dismissed):

ovv (approved):

PR #786, opened 2025-03-12
Detach sinks and backpressure from topic-partitions

ghost:
## Motivation
Currently, Sinks are tightly coupled with topic partitions, which doesn't work well with the upcoming joins and merges. 

This PR removes the dependency on TPs from certain Sinks methods.
`BatchingSink` classes keep batching data per TP to reduce the amount of the changed code, but `BatchingSink.flush()` will flush all the batches.

## Main changes

1. Sinks are now flushed first on each checkpoint before producing changelogs. 
This is done to minimize potential over-production in case of Sink's failure.
2. Updated `Sink.flush()` - it is now expected to flush all the accumulated data for all TPs.
3. Updated Sink backpressure handling - `SinkBackpressureError` now pauses the whole assignment instead of certain partitions only. 
4. Updated existing Sinks to follow the new API

### DIFF
--- a/quixstreams/checkpointing/checkpoint.py
+++ b/quixstreams/checkpointing/checkpoint.py
@@ -191,37 +191,35 @@ class Checkpoint(BaseCheckpoint):
         logger.debug("Checkpoint: flushing sinks")
         sinks = self._sink_manager.sinks
         # Step 1. Flush sinks
-        for (topic, partition), offset in self._tp_offsets.items():
-            for sink in sinks:
-                if self._pausing_manager.is_paused(topic=topic, partition=partition):
-                    # The topic-partition is paused, skip flushing other sinks for
-                    # this TP.
-                    # Note: when flushing multiple sinks for the same TP, some
-                    # of them can be flushed before one of the sinks is backpressured.
-                    sink.on_paused(topic=topic, partition=partition)
-                    continue
+        backpressured = False
+        for sink in sinks:
+            if backpressured:
+                # Drop the accumulated data for the other sinks
+                # if one of them is backpressured to limit the number of duplicates
+                # when the data is reprocessed again
+                sink.on_paused()
+                continue
 
-                try:
-                    sink.flush(topic=topic, partition=partition)
-                except SinkBackpressureError as exc:
-                    logger.warning(
-                        f'Backpressure for sink "{sink}" is detected, '
-                        f"the partition will be paused and resumed again "
-                        f"in {exc.retry_after}s; "
-                        f'partition="{topic}[{partition}]" '
-                        f"processed_offset={offset}"
-                    )
-                    # The backpressure is detected from the sink
-                    # Pause the partition to let it cool down and seek it back to
-                    # the first processed offset of this Checkpoint (it must be equal
-                    # to the last committed offset).
-                    offset_to_seek = self._starting_tp_offsets[(topic, partition)]
-                    self._pausing_manager.pause(
-                        topic=topic,
-                        partition=partition,
-                        resume_after=exc.retry_after,
-                        offset_to_seek=offset_to_seek,
-                    )
+            try:
+                sink.flush()
+            except SinkBackpressureError as exc:
+                logger.warning(
+                    f'Backpressure for sink "{sink}" is detected, '
+                    f"all partitions will be paused and resumed again "
+                    f"in {exc.retry_after}s"
+                )
+                # The backpressure is detected from the sink
+                # Pause the assignment to let it cool down and seek it back to
+                # the first processed offsets of this Checkpoint (it must be equal
+                # to the last committed offset).
+                self._pausing_manager.pause(
+                    resume_after=exc.retry_after,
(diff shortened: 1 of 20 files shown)


quixstreams/checkpointing/checkpoint.py
```diff
@@ -181,13 +181,47 @@ def commit(self):
...
+         4. Commit topic offsets.
+         5. Flush each state store partition to the disk.
         """
 
-        # Step 1. Produce the changelogs
+        logger.debug("Checkpoint: flushing sinks")
+        sinks = self._sink_manager.sinks
+        # Step 1. Flush sinks
+        backpressured = False
+        for sink in sinks:
```
gwaramadze: 1. `sinks` is accessed only once, no need to assign to a variable
2. log belongs to step 1

```suggestion
        # Step 1. Flush sinks
        logger.debug("Checkpoint: flushing sinks")
        backpressured = False
        for sink in self._sink_manager.sinks:
```

quixstreams/processing/pausing.py
```diff
@@ -18,91 +18,75 @@ class PausingManager:
...
 
     def resume_if_ready(self):
         """
-        Resume consuming from topic-partitions after the wait period has elapsed.
+        Resume consuming from assigned data partitions after the wait period has elapsed.
         """
-        now = time.monotonic()
-        if self._next_resume_at > now:
-            # Nothing to resume yet, exit early
+        if self._resume_at > time.monotonic():
```
gwaramadze, line 71: Minor inconsistency: The name of this variable should be `self._resume_after` or else the condition should be `>=`.

quixstreams/processing/pausing.py
```diff
@@ -18,91 +18,75 @@ class PausingManager:
...
             return
 
-        tps_to_resume = [
-            tp for tp, resume_at in self._paused_tps.items() if resume_at <= now
+        # Resume only data TPs excluding changelog TPs
+        non_changelog_tps = [
+            tp
+            for tp in self._consumer.assignment()
+            if tp.topic in self._topic_manager.non_changelog_topics
```
gwaramadze: This code is repeated twice; drop it into a helper method.

quixstreams/processing/pausing.py
```diff
@@ -18,91 +18,75 @@ class PausingManager:
...
 
-    _paused_tps: Dict[Tuple[str, int], float]
-
-    def __init__(self, consumer: BaseConsumer):
+    def __init__(self, consumer: BaseConsumer, topic_manager: TopicManager):
         self._consumer = consumer
-        self._paused_tps = {}
-        self._next_resume_at = _MAX_FLOAT
+        self._topic_manager = topic_manager
+        self._resume_at = _MAX_FLOAT
```
gwaramadze:
```suggestion
        self.reset()
```
Plus, I'd move the reset method to the top right under the `__init__`.

quixstreams/sinks/base/sink.py
```diff
@@ -177,28 +177,26 @@ def add(
...
-            # Always drop the batch after flushing it
-            self._batches.pop((topic, partition), None)
+            # Always drop batches after flushing
+            self._batches.clear()
 
-    def on_paused(self, topic: str, partition: int):
+    def on_paused(self):
         """
-        When the destination is already backpressure, drop the accumulated batch.
+        When the destination is already backpressure, drop the accumulated batches.
```
gwaramadze:
```suggestion
        When the destination is already backpressured, drop the accumulated batches.
```

quixstreams/sinks/community/kinesis.py
```diff
@@ -130,30 +130,25 @@ def add(
...
         complete, ensuring that all records are successfully sent to the Kinesis
         stream.
         """
-        topic_partition = (topic, partition)
-
         # Submit any remaining records
-        if records := self._records.pop(topic_partition, None):
-            self._submit(topic_partition, records)
+        for tp, records in self._records.items():
+            # if records := self._records.pop(topic_partition, None):
```
gwaramadze:
```suggestion
```

tests/test_quixstreams/test_processing/test_pausing.py
```diff
@@ -1,114 +1,95 @@
...
+        topic = topic_manager.topic(
+            name=str(uuid.uuid4()),
+            create_config=TopicConfig(num_partitions=2, replication_factor=1),
+        )
+        # Create a changelog topic
+        topic_manager.changelog_topic(
+            topic_name=topic.name, store_name="default", config=topic.broker_config
+        )
+
+        # partition1, partition2 = 0, 1
```
gwaramadze:
```suggestion
```

gwaramadze (commented):
I like how this simplifies the code overall.

quixstreams/processing/pausing.py
ghost: There's already `resume_after` variable in the same method with a different meaning.
Tried to rename it, but looks more confusing

gwaramadze (approved):

PR #787, opened 2025-03-13
Updated file source

tim-quix:
First attempt at readjusting file source.

It is generally "flattened" now. There are things like the async file fetching, file deserialization, and file decompressing that are done via imported functions or classes, but they are not objects that users are expected to manage or alter; they just help encapsulate/hide some default functionality that is unlikely to be altered by users. 

If any custom functionality is required, you can always just override FileSource methods/attributes directly...there are no additional objects to manage. I think the current approach is flexible enough to handle many use cases directly though.

I also attempted to maintain feature parity with the original by adding a new flag called `has_partition_folders` to give the option of having a topic with more than 1 partition, so those who use it alongside FileSink can still get the original functionality, but now the default behavior assumes nothing about file structure.

Example new S3FileSource use:

```
from quixstreams.sources.community.file.s3 import S3FileSource


def key_setter(record: dict) -> str:
    return record["host_id"]


def value_setter(record: dict) -> dict:
    return {k: record[k] for k in ["field_x", "field_y"]}


def timestamp_setter(record: dict) -> int:
    return record['timestamp']


app = Application(broker_address="localhost:9092", consumer_group='s3')
source = S3FileSource(
    filepath='folder_a/folder_b',
    bucket="my_bucket",
    region_name="us-east2",
    aws_access_key_id="my_key_id",
    aws_secret_access_key="my_secret",
    key_setter=key_setter,
    value_setter=value_setter,
    timestamp_setter=timestamp_setter,
    file_format="json",
    compression="gzip",
    has_partition_folders=False,
    replay_speed=0.5,
)
app.add_source(source)


if __name__ == "__main__":
    app.run()

```


tomas-quix (dismissed):

tim-quix:
Gonna do some last double checking to ensure everything works with the client-based ones as expected (only tested the local file one), but otherwise good for reviewing.

quixstreams/sources/community/file/azure.py
```diff
@@ -0,0 +1,157 @@
...
+
+try:
+    from azure.storage.blob import ContainerClient
+except ImportError as exc:
+    raise ImportError(
+        f"Package {exc.name} is missing: "
+        'run "pip install quixstreams[azure-file]" to use AzureFileOrigin'
+    ) from exc
+
+__all__ = ("AzureFileSource",)
```
gwaramadze: I think `MissingAzureContainer` should be here too.

quixstreams/sources/community/file/s3.py
```diff
@@ -0,0 +1,171 @@
...
+        }
+        self._client: Optional[S3Client] = None
+
+    def _close_client(self):
+        logger.debug("Closing S3 client session...")
+        if self._client:
+            try:
+                self._client.close()
+            except Exception as e:
+                logger.error(f"Azure client session exited non-gracefully: {e}")
```
gwaramadze: 😉
```suggestion
                logger.error(f"S3 client session exited non-gracefully: {e}")
```

quixstreams/sources/community/file/azure.py
```diff
@@ -0,0 +1,157 @@
...
+            value_setter=value_setter,
+            timestamp_setter=timestamp_setter,
+            file_format=file_format,
+            compression=compression,
+            has_partition_folders=has_partition_folders,
+            replay_speed=replay_speed,
+            name=name or f"azurefilestore_{container}_{Path(filepath).name}",
+            shutdown_timeout=shutdown_timeout,
+            on_client_connect_success=on_client_connect_success,
+            on_client_connect_failure=on_client_connect_failure,
```
gwaramadze, line 104: WDYT if we focus only on what is needed to make this subclass distinct? This should do, but I have no way to test it, so you'd have to do it. I understand if you prefer not to 😅

```suggestion
        name: Optional[str] = None,
        **kwargs
    ):
        """
        :param connection_string: Azure client authentication string.
        :param container: Azure container name.
        :param filepath: folder to recursively iterate from (a file will be used directly).
        :param name: The name of the Source application (Default: last folder name).
        :param kwargs: Additional keyword arguments passed to FileSource.
        """
        super().__init__(
            filepath=filepath,
            name=name or f"azurefilestore_{container}_{Path(filepath).name}",
            **kwargs
```

quixstreams/sources/community/file/azure.py
```diff
@@ -0,0 +1,157 @@
...
+        self._client: Optional[ContainerClient] = None
+
+    def _close_client(self):
+        logger.debug("Closing Azure client session...")
+        if self._client:
+            try:
+                self._client.close()
+            except Exception as e:
+                logger.error(f"Azure client session exited non-gracefully: {e}")
+        self._client = None
```
gwaramadze: nit
```suggestion
            self._client = None
```

quixstreams/sources/community/file/azure.py
```diff
@@ -0,0 +1,157 @@
...
+            name=name or f"azurefilestore_{container}_{Path(filepath).name}",
+            shutdown_timeout=shutdown_timeout,
+            on_client_connect_success=on_client_connect_success,
+            on_client_connect_failure=on_client_connect_failure,
+        )
+        self._container = container
+        self._auth = connection_string
+        self._client: Optional[ContainerClient] = None
+
+    def _close_client(self):
```
gwaramadze: Can you move this private method to the bottom of the class? Right after the `stop` method is a perfect place for it.

quixstreams/sources/community/file/base.py
```diff
@@ -0,0 +1,216 @@
...
+        value_setter: Optional[Callable[[object], object]] = None,
+        timestamp_setter: Optional[Callable[[object], int]] = None,
+        file_format: Union[Format, FormatName] = "json",
+        compression: Optional[CompressionName] = None,
+        has_partition_folders: bool = False,
+        replay_speed: float = 1.0,
+        name: Optional[str] = None,
+        shutdown_timeout: float = 30,
+        on_client_connect_success: Optional[ClientConnectSuccessCallback] = None,
+        on_client_connect_failure: Optional[ClientConnectFailureCallback] = None,
```
gwaramadze, line 56: If those arguments are not overloaded, we could just catch and pass kwargs, WDYT?
```suggestion
        **kwargs
```

PS. if you agree, remember to remove those from the docstring as well

quixstreams/sources/community/file/components/file_fetcher.py
```diff
@@ -0,0 +1,61 @@
...
+        return self
+
+    def __next__(self) -> tuple[Path, BinaryIO]:
+        if self._stopped:
+            raise StopIteration
+
+        try:
+            file_name = self._downloading_file_name
+            file_content = self._downloading_file_content.result()
+            self._download_next_file()
```
gwaramadze, line 39: What if `_download_next_file` raises `StopIteration`? I think we will not return last `file_name` and `file_content`. Or did I get it wrong?

quixstreams/sources/community/file/local.py
```diff
@@ -0,0 +1,97 @@
...
+            file_format=file_format,
+            compression=compression,
+            has_partition_folders=has_partition_folders,
+            replay_speed=replay_speed,
+            name=name,
+            shutdown_timeout=shutdown_timeout,
+            on_client_connect_success=on_client_connect_success,
+            on_client_connect_failure=on_client_connect_failure,
+        )
+
```
gwaramadze, line 85: This `__init__` is 100% redundant
```suggestion
```

quixstreams/sources/community/file/local.py
```diff
@@ -0,0 +1,97 @@
...
+    def get_file_list(self, filepath: Path) -> Iterable[Path]:
+        if filepath.is_dir():
+            for i in sorted(filepath.iterdir(), key=lambda x: x.name):
+                yield from self.get_file_list(i)
```
gwaramadze: I think key is redundant in this case, also `dir` variable name will be slightly more explicit
```suggestion
            for dir in sorted(filepath.iterdir()):
                yield from self.get_file_list(dir)
```

quixstreams/sources/community/file/local.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            yield filepath
+
+    def read_file(self, filepath: Path) -> BinaryIO:
+        return BytesIO(filepath.read_bytes())
+
+    def file_partition_counter(self) -> int:
+        return len([f for f in self._filepath.iterdir()])
```
gwaramadze: nit
```suggestion
        return len(list(self._filepath.iterdir()))
```

quixstreams/sources/community/file/s3.py
```diff
@@ -0,0 +1,171 @@
...
+    from mypy_boto3_s3 import S3Client
+except ImportError as exc:
+    raise ImportError(
+        f"Package {exc.name} is missing: "
+        'run "pip install quixstreams[s3]" to use S3Origin'
+    ) from exc
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("S3FileSource",)
```
gwaramadze: I think `MissingS3Bucket` should be here too.

quixstreams/sources/community/file/s3.py
```diff
@@ -0,0 +1,171 @@
...
+        timestamp_setter: Optional[Callable[[object], int]] = None,
+        has_partition_folders: bool = False,
+        file_format: Union[Format, FormatName] = "json",
+        compression: Optional[CompressionName] = None,
+        replay_speed: float = 1.0,
+        name: Optional[str] = None,
+        shutdown_timeout: float = 30,
+        on_client_connect_success: Optional[ClientConnectSuccessCallback] = None,
+        on_client_connect_failure: Optional[ClientConnectFailureCallback] = None,
+    ):
```
gwaramadze, line 66: Same story as with Azure's `__init__`; most of these may be tossed into `kwargs` so that we are left with the focus on what makes this subclass unique.

quixstreams/sources/community/file/s3.py
```diff
@@ -0,0 +1,171 @@
...
+                raise
+
+    def get_file_list(self, filepath: Union[str, Path]) -> Iterable[Path]:
+        resp = self._client.list_objects(
+            Bucket=self._bucket,
+            Prefix=str(filepath),
+            Delimiter="/",
+        )
+        for _folder in resp.get("CommonPrefixes", []):
+            yield from self.get_file_list(_folder["Prefix"])
```
gwaramadze: why do we have an underscore here?
```suggestion
        for folder in resp.get("CommonPrefixes", []):
            yield from self.get_file_list(folder["Prefix"])
```

quixstreams/sources/community/file/s3.py
```diff
@@ -0,0 +1,171 @@
...
+        self._bucket = bucket
+        self._credentials = {
+            "region_name": region_name,
+            "aws_access_key_id": aws_access_key_id,
+            "aws_secret_access_key": aws_secret_access_key,
+            "endpoint_url": endpoint_url,
+        }
+        self._client: Optional[S3Client] = None
+
+    def _close_client(self):
```
gwaramadze: same story, let's toss it to the bottom of the class

quixstreams/sources/community/file/azure.py
tim-quix: I don't anticipate anyone needing to import it which is why I didn't include it...what are your thoughts around it?

gwaramadze: * It is a "public" class, an exception raised by another public member of the module.
* It is our decision to raise it under this or that condition, so we sort of, by definition, should expect that the user may want to handle the condition gracefully
* Not being in the `__all__` does not make an object not importable; it's just not importable by `*` import. I believe we should expose public classes consistently.

tim-quix: Yeah I knew it didn't stop the import, I was just trying to keep the list of imports as clutter-free as possible when people are trying to import a given Source/Sink...though even if it's just us, I could see the argument that someone would maybe want to gracefully handle the error, good point :+1:

quixstreams/sources/community/file/s3.py
tim-quix: leftover from some previous name overlaps, whoops

quixstreams/sources/community/file/azure.py
tim-quix: In general I like the idea because I've always hated all the passthrough we've had to do, but we otherwise lose type hinting and auto-complete for the **kwargs, so I'm keeping this consistent with our usual approach as the unfortunate necessary evil haha.

That said, what I sent you in Slack outlines another way we could potentially achieve both, but I'm not sure if it's too much of a change. Take a look and lemme know what you think.

gwaramadze: You mean we lose type hint and auto-complete in the platform, right?

In this case, I think it's best to leave it as is. I'm not sure I like the `Unpack` workaround. Maintaining a separate class to describe `kwargs` seems even stranger 🤔

gwaramadze (dismissed):

quixstreams/sources/community/file/local.py
tim-quix: dir overlapped with a builtin name so i used fp instead and removed the key :+1:

gwaramadze (approved):

PR #788, opened 2025-03-14
InfluxDB3Source: new connector

tim-quix:
NOTE: This PR is primarily just me cleaning up code that was written by someone else; I largely tried to leave what they had in place while cleaning up what i could, or standardizing it to fit our framework.

Example use:

```
from quixstreams.sources.community.influxdb3 import InfluxDB3Source
import datetime
import uuid

source = InfluxDB3Source(
    host="http://localhost:8181",
    organization_id='abc123',
    database="connector5",
    token='123',
    start_date=datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=2),
    end_date=datetime.datetime.now(tz=datetime.timezone.utc),
    time_delta='24h',
    max_retries=2,
)

app = Application(
    consumer_group=str(uuid.uuid4()),
)
sdf = app.dataframe(source=source).print(metadata=True)


if __name__ == '__main__':
    app.run()
```

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -34,6 +34,7 @@ requirements:
     - pandas >=1.0.0,<3.0
     - elasticsearch >=8.17,<9
     - rich >=13,<15
+    - influxdb3-python[pandas] >=0.7,<1.0
 
 test:
   imports:
--- a/conda/post-link.sh
+++ b/conda/post-link.sh
@@ -3,7 +3,7 @@
 $PREFIX/bin/pip install \
 'rocksdict>=0.3,<0.4' \
 'protobuf>=5.27.2,<7.0' \
-'influxdb3-python>=0.7,<1.0' \
+'influxdb3-python[pandas]>=0.7,<1.0' \
 'pyiceberg[pyarrow,glue]>=0.7' \
 'redis[hiredis]>=5.2.0,<6' \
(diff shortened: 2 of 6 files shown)


tim-quix:
[Currently blocked](https://app.shortcut.com/quix/story/65114/create-influxdbsource), awaiting Tomas feedback.

tim-quix:
The cleanup is complete, only need to add docs now. I have tested it and it works.

A few other questions:

1. Do we want to add key, value, and timestamp setters? Seems like something we should have, especially since the message key is just the name of the measurement + randint(1, 1000)
2. Do we need the messages per second tracker that Tomas added in the newest version?
3. Do we want to allow users to specify more frequent queries when there's no `end_date` and its caught up?
4. Should we have the added measurement name field added to the data be called `_measurement_name` rather than `measurement_name`?

Besides these questions, just need to add connector docs and should be good.

tim-quix:
@tomas-quix 

> Do we need the messages per second tracker that Tomas added in the newest version?
> Do we want to allow users to specify more frequent queries when there's no end_date and its caught up? Right now it sleeps a long while when it could be more up-to-date

Otherwise I believe this is good to go, please have a look. I only added some minor things like a key setter and being able to name the outgoing measurement_name column (default `_measurement_name`)

ghost (approved):

PR #789, opened 2025-03-17
Feature: Skip None in Aggregations

gwaramadze:
* Application will not fail if None is encountered in the aggregation context.
* ~~Additionally, NaN (not a number) will be ignored as well as any calculation with it gives back a useless NaN~~

### DIFF
--- a/quixstreams/dataframe/windows/aggregations.py
+++ b/quixstreams/dataframe/windows/aggregations.py
@@ -100,13 +100,20 @@ class Mean(Aggregator):
 
     def agg(self, old: tuple[V, int], new: Any) -> tuple[V, int]:
         old_sum, old_count = old
-        if self.column is ROOT:
+        new = new if self.column is ROOT else new.get(self.column)
+        try:
             return old_sum + new, old_count + 1
-        return old_sum + new[self.column], old_count + 1
+        except TypeError:
+            if new is None:
+                return old
+            raise
 
     def result(self, value: tuple[Union[int, float], int]) -> float:
         sum_, count_ = value
-        return sum_ / count_
+        try:
+            return sum_ / count_
+        except ZeroDivisionError:
+            return 0.0
 
 
 R = TypeVar("R", int, float)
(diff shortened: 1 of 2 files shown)


tim-quix (dismissed):

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -153,10 +158,11 @@ def initialize(self) -> None:
         return None
 
     def agg(self, old: Optional[V], new: Any) -> V:
-        if self.column is not ROOT:
-            new = new[self.column]
+        new = new if self.column is ROOT else new.get(self.column)
         if old is None:
             return new
+        elif new is None:
+            return old
```
ghost, line 164: Can both `old` and `new` be Nones?

ovv: Yes. `old` can be `None` because we initialise the aggregation to `None`

gwaramadze: Yes, but it's ok. Let me know if you want to discuss it.

quixstreams/dataframe/windows/aggregations.py
```diff
@@ -95,13 +94,17 @@ def initialize(self) -> tuple[float, int]:
...
         old_sum, old_count = old
-        if self.column is ROOT:
-            return old_sum + new, old_count + 1
-        return old_sum + new[self.column], old_count + 1
+        return old_sum + new, old_count + 1
 
     def result(self, value: tuple[Union[int, float], int]) -> float:
         sum_, count_ = value
+        if sum_ == 0.0:
+            return 0.0
```
ovv: An empty mean should return `None` imo

ovv (changes requested):

quixstreams/dataframe/windows/aggregations.py
gwaramadze: fixed

ovv (approved):

ghost (approved):

PR #792, opened 2025-03-18
Feature: Fill missing data

gwaramadze:
## Summary

This pull request introduces a new fill method to the StreamingDataFrame class, enabling users to fill missing values in their data streams with specified constants. This enhancement provides more flexibility in data preprocessing and ensures that data streams can be standardized before further processing.

## Key Features

1. **Fill Missing Values**: The fill method allows users to specify columns and their corresponding fill values. If a column is missing or contains None, it will be replaced with the specified fill value.
2. **Flexible Usage**: Users can fill multiple columns at once using either positional arguments (for default None values) or keyword arguments (for specific fill values).

## Example Usage
Fill missing values for a single column with a None:
```python
# This would transform {"x": 1} to {"x": 1, "y": None}
sdf.fill("y")
```

Fill missing values for multiple columns with a None:
```python
# This would transform {"x": 1} to {"x": 1, "y": None, "z": None}
sdf.fill("y", "z")
```

Fill missing values in the value with a constant value using a dictionary:
```python
# This would transform {"x": None} to {"x": 1, "y": 2}
sdf.fill(x=1, y=2)
```

Use a combination of positional and keyword arguments:
```python
# This would transform {"y": None} to {"x": None, "y": 2}
sdf.fill("x", y=2)
```

### DIFF
--- a/docs/missing-data.md
+++ b/docs/missing-data.md
@@ -1,0 +1,53 @@
+# Missing Data
+
+Handling missing data is a crucial aspect of data processing, especially in streaming environments. This document outlines what is considered missing data within the `StreamingDataFrame` class and how it is managed.
+
+## Types of Missing Data in Streaming
+
+- **Missing Column**: This occurs when a field is not present in the message at all. In streaming data, schemas can be dynamic, meaning that not all fields are required to be present in every message. This type of missing data is handled by the system's ability to adapt to changes in the schema over time.
+
+- **Missing Value**: This occurs when a field is present in the message, but its value is `None`. This indicates that the data for that field is missing, even though the field itself is part of the message schema.
+
+## Handling Missing Data in Aggregations
+
+- **Rows with `None` Values**: These rows are ignored during aggregation operations. This means that if a row contains a `None` value, it will not contribute to the aggregation result. This applies to the following aggregations: Sum, Mean, Min, and Max.
+
+- **`NaN` Values**: Unlike `None`, `NaN` values are propagated to the aggregation result. This is because `NaN` is not considered missing data in the same way as `None`. Instead, it represents a numerical value that is undefined or unrepresentable, and it is treated as such in calculations.
+
+## `StreamingDataFrame.fill` Method
+
+The `fill` method in the `StreamingDataFrame` class is used to fill missing data in the message value with a constant value.
+
+### Example Usage
+
+```python
+from quixstreams import Application
+
+# Initialize the Application
+app = Application(...)
+sdf = app.dataframe(...)
+```
+
+Fill missing data for a single column with `None`:
+```python
+# This would transform {"x": 1} to {"x": 1, "y": None}
+sdf.fill("y")
+```
+
+Fill missing data for multiple columns with `None`:
+```python
+# This would transform {"x": 1} to {"x": 1, "y": None, "z": None}
+sdf.fill("y", "z")
+```
+
+Fill missing data with a constant value using a dictionary:
+```python
+# This would transform {"x": None} to {"x": 1, "y": 2}
+sdf.fill(x=1, y=2)
+```
+
+Use a combination of positional and keyword arguments:
+```python
+# This would transform {"y": None} to {"x": None, "y": 2}
+sdf.fill("x", y=2)
+```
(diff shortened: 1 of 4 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -1385,6 +1385,55 @@ def sliding_count_window(
...
+
+        :param columns: a list of column names as strings.
+        :param mapping: a dictionary where keys are column names and values are the fill values.
+        :return: the original `StreamingDataFrame` instance for chaining.
+        """
+        mapping = {**{column: None for column in columns}, **mapping}
+
+        def _fill(value: Any) -> Any:
+            if not isinstance(value, dict):
+                return value
```
ovv, line 1458: Do we need this ?

I think we could assume the value is a dict (or dict-like) and catch the exception (`AtributeError`/`TypeError`) if it's not ?

quixstreams/dataframe/dataframe.py
```diff
@@ -1385,6 +1385,55 @@ def sliding_count_window(
...
+            for column, fill_value in mapping.items():
+                incoming_value = value.get(column)
+                value[column] = fill_value if incoming_value is None else incoming_value
```
ovv:
```suggestion
                if incoming_value is None:
                    value[column] = fill_value
```
We don't need to set the value if it's non-none

gwaramadze: That's true, I've reduced even more code with
```python
                if value.get(column) is None:
                    value[column] = fill_value
```

quixstreams/dataframe/dataframe.py
gwaramadze: try/except rather adds more code in this case, I am inclined towards keeping the if. But I have added a test case for this.

docs/missing-data.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Types of Missing Data in Streaming
+
+- **Missing Column**: This occurs when a field is not present in the message at all. In streaming data, schemas can be dynamic, meaning that not all fields are required to be present in every message. This type of missing data is handled by the system's ability to adapt to changes in the schema over time.
+
+- **Missing Value**: This occurs when a field is present in the message, but its value is `None`. This indicates that the data for that field is missing, even though the field itself is part of the message schema.
+
+## Handling Missing Data in Aggregations
+
+- **Rows with `None` Values**: These rows are ignored during aggregation operations. This means that if a row contains a `None` value, it will not contribute to the aggregation result. This applies to the following aggregations: Sum, Mean, Min, and Max.
```
ovv: We need to remember to add `Count` here

docs/missing-data.md
```diff
@@ -0,0 +1,53 @@
...
+
+- **`NaN` Values**: Unlike `None`, `NaN` values are propagated to the aggregation result. This is because `NaN` is not considered missing data in the same way as `None`. Instead, it represents a numerical value that is undefined or unrepresentable, and it is treated as such in calculations.
+
+## `StreamingDataFrame.fill` Method
+
+The `fill` method in the `StreamingDataFrame` class is used to fill missing data in the message value with a constant value.
```
ovv:
```suggestion
The `fill` method in the `StreamingDataFrame` class is used to fill missing column and missing value in the message with a constant value.
```
We should use `missing column` and `missing value` to indicate `fill` supports both and we just defined what it means.

quixstreams/dataframe/dataframe.py
```diff
@@ -1385,6 +1385,62 @@ def sliding_count_window(
...
+        :return: the original `StreamingDataFrame` instance for chaining.
+        """
+        mapping = {**{column: None for column in columns}, **mapping}
+
+        if not mapping:
+            warnings.warn(
+                "No columns or mapping provided to fill(). "
+                "No changes will be made to the data."
+            )
+            return self
```
ovv: Should we raise instead ?

ovv (dismissed):

quixstreams/dataframe/dataframe.py
ghost: Agree, if it's an incorrect use it's better to raise

ghost (dismissed):

ghost (approved):

PR #793, opened 2025-03-19
add convert_ints_to_floats to InfluxDB3Sink

tim-quix:
Note the main challenge here is the set of fields is dynamic except when a static list of fields is passed, so for the most part we need to inspect each message value. 

We could add a separate check for when users pass a static list of fields (that is, `field_keys` is a non-empty, non-callable), but I figured if users are doing so, they probably are handling a small enough subset that it wouldn't make a huge difference to do something other than loop through them anyway.

I also considered making a dict with a hash of input fields and the list of fields that were ints or floats for that input set so we wouldn't have to loop through each field every time, but it was only beneficial with 50+ fields (of course better as field count went up) and when float fields were not the majority. Figured there were too many caveats with that approach to warrant overcomplicating the code.

### DIFF
--- a/quixstreams/sinks/core/influxdb3.py
+++ b/quixstreams/sinks/core/influxdb3.py
@@ -270,18 +270,20 @@ class InfluxDB3Sink(BatchingSink):
                     tags["__topic"] = batch.topic
                     tags["__partition"] = batch.partition
 
-                field_values = (
-                    (f, value[f])
-                    for f in _fields_keys
-                    if f in value or not self._allow_missing_fields
-                )
+                if _fields_keys:
+                    fields = {
+                        f: value[f]
+                        for f in _fields_keys
+                        if f in value or not self._allow_missing_fields
+                    }
+                else:
+                    fields = value
+
                 if self._convert_ints_to_floats:
                     fields = {
                         k: float(v) if isinstance(v, int) else v
-                        for k, v in (field_values if _fields_keys else value.items())
+                        for k, v in fields.items()
                     }
-                else:
-                    fields = {k: v for k, v in field_values} if _fields_keys else value
 
                 ts = value[time_key] if time_key is not None else item.timestamp
                 record = {
--- a/quixstreams/sinks/core/influxdb3.py
+++ b/quixstreams/sinks/core/influxdb3.py
@@ -61,6 +61,7 @@ class InfluxDB3Sink(BatchingSink):
         time_precision: TimePrecision = "ms",
         allow_missing_fields: bool = False,
         include_metadata_tags: bool = False,
+        convert_ints_to_floats: bool = False,
         batch_size: int = 1000,
         enable_gzip: bool = True,
         request_timeout_ms: int = 10_000,
@@ -119,6 +120,8 @@ class InfluxDB3Sink(BatchingSink):
         :param include_metadata_tags: if True, includes record's key, topic,
             and partition as tags.
             Default - `False`.
+        :param convert_ints_to_floats: if True, converts all integer values to floats.
+            Default - `False`.
         :param batch_size: how many records to write to InfluxDB in one request.
             Note that it only affects the size of one write request, and not the number
             of records flushed on each checkpoint.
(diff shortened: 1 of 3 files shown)


quixstreams/sinks/core/influxdb3.py
```diff
@@ -261,16 +269,19 @@ def write(self, batch: SinkBatch):
                     tags["__topic"] = batch.topic
                     tags["__partition"] = batch.partition
 
-                fields = (
-                    {
-                        field_key: value[field_key]
-                        for field_key in _fields_keys
-                        if (field_key in value or not self._allow_missing_fields)
-                        and field_key not in _tags_keys
```
gwaramadze, line 270: This check was evicted. I'm not sure what was it for, just want to make sure it is an intended change 😅

quixstreams/sinks/core/influxdb3.py
```diff
@@ -261,16 +269,19 @@ def write(self, batch: SinkBatch):
...
-                    }
-                    if _fields_keys
-                    else value
+                field_values = (
+                    (f, value[f])
+                    for f in _fields_keys
+                    if f in value or not self._allow_missing_fields
```
gwaramadze: I don't get one thing, please help me grasp it.

if `f` key is not in `value` dict and `not self._allow_missing_fields` evaluates to True, won't it mean that the `value[f]` will run and raise `KeyError`? What did I miss here 🤔

quixstreams/sinks/core/influxdb3.py
```diff
@@ -261,16 +269,19 @@ def write(self, batch: SinkBatch):
...
                 )
+                if self._convert_ints_to_floats:
+                    fields = {
+                        k: float(v) if isinstance(v, int) else v
+                        for k, v in (field_values if _fields_keys else value.items())
+                    }
+                else:
+                    fields = {k: v for k, v in field_values} if _fields_keys else value
+
```
gwaramadze, line 287: I was looking for a way to improve readability here. Maybe this:
```python
                if _fields_keys:
                    fields = {
                        f: value[f]
                        for f in _fields_keys
                        if f in value or not self._allow_missing_fields
                    }
                else:
                    fields = value

                if self._convert_ints_to_floats:
                    fields = {
                        k: float(v) if isinstance(v, int) else v
                        for k, v in fields.items()
                    }
```

I can't post it as a suggestion, it should replace everything starting from line 272. I ran the tests and seems fine.

tests/test_quixstreams/test_sinks/test_core/test_influxdb_v3.py
```diff
@@ -363,3 +365,47 @@ def test_write_fails_propagates_exception(self, influxdb3_sink_factory):
...
+            "record": [
+                {
+                    "measurement": measurement,
+                    "tags": {},
+                    "fields": result,
+                    "time": timestamp,
+                }
+            ],
+            "write_precision": "ms",
+        }
```
gwaramadze: There is a shorthand for this
```suggestion
        client_mock.write.assert_called_once_with(
            record=[
                {
                    "measurement": measurement,
                    "tags": {},
                    "fields": result,
                    "time": timestamp,
                }
            ],
            write_precision="ms",
        )
```

quixstreams/sinks/core/influxdb3.py
tim-quix: Good eye! I removed it because it was a redundant check. 

Basically the goal is to make sure the same `key` is not in both `fields` and `tags` because it throws errors if you try to include it in both when pushing to InfluxDB.

There's a previous step that does:

```
tags = {}
for tag_key in _tags_keys:
    tag = value.pop(tag_key)
    tags[tag_key] = tag
```

Which means that for this:

```
...
if (field_key in value or not self._allow_missing_fields)
and field_key not in _tags_keys
...
```
 if a given key in `fields` was also in `tags`, it should have been popped from the `value` dict previously, which means the check `if (field_key in value...)` would return `False` as desired, and  `and field_key not in _tags_keys` would basically never resolve when it would theoretically matter and always be True otherwise.

On top of that, we prob shouldn't quietly ignore an issue like that anyway and instead let it raise, esp since we already raise for this problem when users pass static lists for `fields` and `keys` that overlap since it's an easy one time check.

quixstreams/sinks/core/influxdb3.py
tim-quix: No you're exactly right, it should raise a `KeyError` when you try to include field `f` that does not exist in `value` when you have not allowed there to be missing fields.

Basically, if you specify a field in your field_set that is NOT included in your outgoing data set, InfluxDB's SDK will auto-populate it will a null value.

If you expect that field to always be present in your data set and do not want this auto-populating behavior, you can say `self._allow_missing_fields = False`, which means it'll always attempt to extract it, which should then `KeyError`.

tests/test_quixstreams/test_sinks/test_core/test_influxdb_v3.py
tim-quix: Indeed, I was mostly just copying the previous test patterns for consistency, but this is obviously the better way to do it =)

I can also tweak the other ones to match this style if you think it's worthwhile, or just leave it all as-is.

quixstreams/sinks/core/influxdb3.py
tim-quix: I can def see the argument for readability, i was originally thinking something like this. 

The current approach ends up only looping through the field keys once by using a generator (which it does create each time regardless of being used), but maybe it's too cute for too little gain, esp if the field_keys list is small, which it's more likely to be than not since it's user generated.

Anywho, nice suggestion, I think this is overall probably better in most situations =)

tests/test_quixstreams/test_sinks/test_core/test_influxdb_v3.py
gwaramadze: up to you man 😎
it's not a big deal, I'd probably do it

gwaramadze (dismissed):

gwaramadze (approved):

PR #802, opened 2025-03-26
Implement `StreamingDataFrame.concat()`

ghost:
This PR adds a new feature - `StreamingDataFrame.concat()`.
With `.concat()`, you can combine two different SDFs, get a new one, and transform data from multiple streams like a single one.
These SDFs may belong to different topics and be branches of the same SDF.

To achieve that, a massive overhaul was done.

### Use case 1: Combining two topics.

**Requirements for using State stores when combining different topics**:

- Topics must have the same number of partitions.  
Otherwise, the code will raise an error.

- The message keys should be distributed using the same algorithm (same partitioner) to perform stateful operations and repartitioning.  

**Note:** The list applies only when the application uses State via `stateful=True` functions or windows.
If the app doesn't use state, it's ok to concat topics with different numbers of partitions.

```python
from quixstreams import Application, State

topic_a = app.topic("topic-a")
topic_b = app.topic("topic-b")

sdf_a = app.dataframe(topic_a)
sdf_b = app.dataframe(topic_b)

# Combine "sdf_a" and "sdf_b" into a new dataframe "sdf_c"
# "sdf_c" will process messages from both topics and can use a shared state 
sdf_c = sdf_a.concat(sdf_b)

# "sdf_c" will register a new State store, independent of "sdf_a" and "sdf_b"
sdf_c.update(..., stateful=True)

# The repartition topics used by "group_by" are also independent of original "sdf_a" and "sdf_b"
sdf_c.group_by(...)
```

### Use case 2: Combining branches of the same original StreamingDataFrame and topic.  

**Note:** If the branches are not exclusive (when each message goes to each branch), the `.concat()` operation will output X times more values, where X is the number of branches. 

```python
from quixstreams import Application, State

topic_a = app.topic("topic-a")


sdf_a = app.dataframe(topic_a)

# Create branches with some processing code
branch_1 = sdf_a.filter(...).apply(...)

branch_2 = sdf_a.filter(...).apply(...)

# Combine branches back into the same dataframe
sdf_c = branch_1.combine(branch_2)

# "sdf_c" will use the same State stores as the branches
sdf_c.update(..., stateful=True)

# The repartition topics are also registered in the same namespace as the original topic
sdf_c.group_by(...)
```

### Introduced `StreamingDataFrame.stream_id`
Stream ID is used to break a 1:1 linking between input topics and State stores + repartitioning.

Though `stream_id` is still generated from topic names (and SDF can now belong to multiple topics), this ID is agnostic of the topics themselves and only identifies the data streams.  
In future releases, we may allow customizing it. 

There's a mapping between stream_ids and actual topics, which is used to assign proper state stores on rebalancing and to record the processed offsets to the changelog topics on checkpoint commits.
The mapping is maintained by `DataFrameRegistry`.

### Added `Topic.quix_name`
Added `Topic.quix_name` to simplify mapping of topic IDs (prefixed with workspace_id) and actual names.
For other platforms, `quix_name == name`

### Added `StreamingDataFrame.concat()`

`StreamingDataFrame.concat()` combines two different SDFs into one.
The new SDF will also have a new `stream_id` if different topics are combined.
If branches of the same SDF are combined, the `stream_id` remains the same.

### Docs
The docs will be added later to this PR.


ghost:
Closes #487

quixstreams/app.py
```diff
@@ -1036,9 +1032,14 @@ def _on_revoke(self, _, topic_partitions: List[TopicPartition]):
...
-                self._state_manager.on_partition_revoke(
-                    topic=tp.topic, partition=tp.partition
+                stream_ids = self._dataframe_registry.get_stream_ids(
+                    topic_name=tp.topic
                 )
+                # Revoke store partitions for the given stream ids
+                for stream_id in stream_ids:
+                    self._state_manager.on_partition_revoke(
+                        stream_id=stream_id, partition=tp.partition
+                    )
```
gwaramadze: this whole bit
```python
        non_changelog_topics = self._topic_manager.non_changelog_topics
        non_changelog_tps = [
            tp for tp in topic_partitions if tp.topic in non_changelog_topics
        ]
        for tp in non_changelog_tps:
                stream_ids = self._dataframe_registry.get_stream_ids(
                    topic_name=tp.topic
                )
                for stream_id in stream_ids:
                    self._state_manager.on_partition_revoke(
                        stream_id=stream_id, partition=tp.partition
                    )
            self._processing_context.on_partition_revoke(
                topic=tp.topic, partition=tp.partition
            )
```

is identical in `_on_revoke` and `_on_lost`.

`_on_lost` is virtually a subset of `_on_revoke` not counting the debug log. Can you please make it DRY? I think we can move it to a separate helper method like `_drop_partitions` or something similar.

quixstreams/dataframe/dataframe.py
```diff
@@ -126,19 +131,24 @@ class StreamingDataFrame:
...
+        registry: DataFrameRegistry,
         processing_context: ProcessingContext,
         stream: Optional[Stream] = None,
     ):
+        if not topics:
+            raise ValueError("At least one Topic must be passed")
+
         self._stream: Stream = stream or Stream()
-        self._topic = topic
+        self._topics = set(topics)
```
gwaramadze: EDIT: Please don't apply this comment right away! I think the only place that requires sorted topic names is `TopicManager.stream_id_from_topics`, I left a comment there as well.

---------------------------------

~~Since `_topics` are immutable, we can sort them once.~~
```suggestion
        self._topics = sorted(set(topics), key=lambda t: t.name)
```

~~Additional reason: ensure deterministic behavior now and in the future. Eg. we need topics for `self._registry.register_stream_id` call that constructs two dicts which will have deterministically ordered keys. It's not a problem now, but we can future-proof it for no cost.~~

gwaramadze: If you don't apply the other suggestion about this line, alternatively, let's use a frozenset to indicate that we don't allow mutating topics after sdf instantiation.
```suggestion
        self._topics = frozenset(topics)
```

quixstreams/dataframe/dataframe.py
```diff
@@ -149,8 +159,24 @@ def stream(self) -> Stream:
...
+
+        StreamingDataFrames with different `stream_id` cannot access the same state stores.
+
+        By default, a topic name or a combination of topic names are used as `stream_id`.
+        """
+        return self._topic_manager.stream_id_from_topics(*self.topics)
+
+    @property
+    def topics(self) -> list[Topic]:
+        return sorted(self._topics, key=lambda t: t.name)
```
gwaramadze: Related to another comment:
```suggestion
        return self._topics
```

quixstreams/dataframe/dataframe.py
```diff
@@ -1467,6 +1502,39 @@ def _sink_callback(
...
+        unrelated to the original dataframes and topics.
+        The same is true for the repartition topics created by `.group_by()`.
+
+        :param other: other StreamingDataFrame
+        :return: a new StreamingDataFrame
+
+        """
+        merged_stream = self.stream.merge(other.stream)
+        topics = set(self.topics) | set(other.topics)
+        return self.__dataframe_clone__(*topics, stream=merged_stream)
```
gwaramadze: Technically, since the `StreamingDataFrame.__init__` is converting `topics` to set, here we could skip it. This will work out of the box:
```suggestion
        stream = self.stream.merge(other.stream)
        return self.__dataframe_clone__(*self.topics, *other.topics, stream=stream)
```

quixstreams/dataframe/registry.py
```diff
@@ -36,18 +36,26 @@ def consumer_topics(self) -> List[Topic]:
...
+        dataframe: "StreamingDataFrame",
     ):
         """
         Register a "root" SDF, or the start of a topic's processing.
-        :param new_sdf: the new SDF.
+        :param dataframe: the new SDF.
         """
-        if (topic := new_sdf.topic).name in self._registry:
+        topics = dataframe.topics
+        # TODO: Test
```
gwaramadze: tested? check if this comment is still relevant

quixstreams/dataframe/registry.py
```diff
@@ -89,3 +96,39 @@ def compose_all(
...
         return executors
+
+    def register_stream_id(self, stream_id: str, topic_names: list[str]):
+        """
+        Register a mapping between the stream_id and topic names.
+        This mapping is later used to match topics to state stores
+        during assignment and commits.
+
+        The same stream id can be registered multiple times.
+         # TODO: Test
```
gwaramadze: tested? check if this comment is still relevant

quixstreams/dataframe/registry.py
```diff
@@ -89,3 +96,39 @@ def compose_all(
...
+        :param topic_names: list of topics to map the stream id with
+        """
+        for topic_name in topic_names:
+            self._topics_to_stream_ids.setdefault(topic_name, set()).add(stream_id)
+            self._stream_ids_to_topics.setdefault(stream_id, set()).add(topic_name)
+
+    def get_stream_ids(self, topic_name: str) -> list[str]:
+        """
+        Get a list of stream ids for the given topic name
+        # TODO: Test
```
gwaramadze: tested? check if this comment is still relevant

quixstreams/dataframe/registry.py
```diff
@@ -89,3 +96,39 @@ def compose_all(
...
+        :param topic_name: a name of the topic
+        :return: a list of stream ids
+        """
+        return list(self._topics_to_stream_ids[topic_name])
+
+    def get_topics_for_stream_id(self, stream_id: str) -> list[str]:
+        """
+        Get a list of topics for the given stream id.
+
+        # TODO: Test
```
gwaramadze: tested? check if this comment is still relevant

quixstreams/models/topics/manager.py
```diff
@@ -201,147 +202,144 @@ def register(self, topic: Topic) -> Topic:
...
+        )
+
+    def stream_id_from_topics(self, *topics: Topic) -> str:
+        """
+        Generate a stream_id by combining names of the provided topics.
+        """
+        if not topics:
+            raise ValueError("At least one Topic must be passed")
+
+        return "--".join(t.name for t in topics)
```
gwaramadze: I left a comment about sorting the topics in `StreamingDataFrame.__init__`. But on the other hand, this looks like the only place that requires topics to be sorted, so maybe let it do it here.

And look, you already have a list of names, no need to pass `key` to `sorted`
```suggestion
        return "--".join(sorted(t.name for t in topics))
```

gwaramadze (commented):
Round 1

quixstreams/models/topics/manager.py
```diff
@@ -201,147 +202,144 @@ def register(self, topic: Topic) -> Topic:
...
+        return TopicConfig(
+            num_partitions=num_partitions,
+            replication_factor=replication_factor,
+            extra_config={
+                "retention.bytes": str(retention_bytes),
+                "retention.ms": str(retention_ms),
+            },
+        )
+
+    def stream_id_from_topics(self, *topics: Topic) -> str:
```
gwaramadze: I wonder if there is a reason for packing positional arguments versus accepting a single argument with precise typing.
```suggestion
    def stream_id_from_topics(self, topics: FrozenSet[Topic]) -> str:
```

quixstreams/dataframe/dataframe.py
```diff
@@ -149,8 +159,24 @@ def stream(self) -> Stream:
...
+
+        It is used as a common prefix for state stores and group-by topics.
+        A new `stream_id` is set when StreamingDataFrames are merged via `.merge()`
+        or grouped via `.group_by()`.
+
+        StreamingDataFrames with different `stream_id` cannot access the same state stores.
+
+        By default, a topic name or a combination of topic names are used as `stream_id`.
+        """
+        return self._topic_manager.stream_id_from_topics(*self.topics)
```
gwaramadze: Followup on a comment on the `stream_id_from_topics` method signature.
```suggestion
        return self._topic_manager.stream_id_from_topics(self._topics)
```

quixstreams/models/topics/manager.py
```diff
@@ -201,147 +202,144 @@ def register(self, topic: Topic) -> Topic:
...
-                    raise TopicConfigurationMismatch(
-                        f'changelog topic "{changelog_topic.name}" replication factor '
-                        f'does not match its source topic "{source_topic.name}": '
-                        f"expected {source_config.replication_factor}, "
-                        f'got {changelog_config.replication_factor}"'
-                    )
-
-        logger.info("Kafka topics validation complete")
+    @classmethod
+    def derive_topic_config(cls, *topics: Topic) -> TopicConfig:
```
gwaramadze: Same here, maybe just accept a single `topics` argument.
```suggestion
    def derive_topic_config(cls, topics: FrozenSet[Topic]) -> TopicConfig:
```

quixstreams/dataframe/dataframe.py
```diff
@@ -1512,19 +1585,23 @@ def _groupby_key(
 
     def __dataframe_clone__(
         self,
+        *topics: Topic,
         stream: Optional[Stream] = None,
-        topic: Optional[Topic] = None,
+        stream_id: Optional[str] = None,
```
gwaramadze: This argument is not used 🤔

quixstreams/dataframe/dataframe.py
```diff
@@ -1512,19 +1585,23 @@ def _groupby_key(
...
 
+        :param topics: one or more `Topic` objects
         :param stream: instance of `Stream`, optional.
-        :param topic: instance of `Topic`, optional.
+        :param stream_id: an identifier of the data stream, optional
         :return: a new `StreamingDataFrame`.
         """
+
         clone = self.__class__(
+            *topics or self._topics,
```
gwaramadze: This would make it slightly more clear
```suggestion
            *(topics or self._topics),
```

quixstreams/dataframe/registry.py
```diff
@@ -13,19 +13,19 @@
...
-    `SDF`s are registered by storing their topic and current Stream.
     """
 
     def __init__(self) -> None:
-        self._registry: Dict[str, Stream] = {}
-        self._topics: List[Topic] = []
-        # {repartition_topic_name: source_topic_name}
-        self._repartition_origins: Dict[str, str] = {}
+        self._registry: dict[str, Stream] = {}
+        self._topics: list[Topic] = []
```
gwaramadze, line 25: We are left with a single usage of `Dict` and `List` in this module. Let's finish the conversion and remove imports.

quixstreams/dataframe/registry.py
```diff
@@ -36,18 +36,26 @@ def consumer_topics(self) -> List[Topic]:
 
     def register_root(
         self,
-        new_sdf: "StreamingDataFrame",
+        dataframe: "StreamingDataFrame",
     ):
         """
         Register a "root" SDF, or the start of a topic's processing.
```
gwaramadze: Not a part of PR, but what the heck does it mean `or the start of a topic's processing`? 🤔 Do you think you could clarify this BTW? Maybe it's best to remove this part?

quixstreams/dataframe/registry.py
```diff
@@ -89,3 +96,39 @@ def compose_all(
...
+        """
+        Get a list of stream ids for the given topic name
+        # TODO: Test
+
+        :param topic_name: a name of the topic
+        :return: a list of stream ids
+        """
+        return list(self._topics_to_stream_ids[topic_name])
+
+    def get_topics_for_stream_id(self, stream_id: str) -> list[str]:
```
gwaramadze, line 126: Having `stream_id` as an argument is self-explanatory.
```suggestion
    def get_topics(self, stream_id: str) -> list[str]:
```

gwaramadze (commented):
Round 2

quixstreams/state/manager.py
```diff
@@ -162,10 +164,10 @@ def _setup_changelogs(
 
     def register_store(
         self,
-        topic_name: Optional[str],
+        stream_id: Optional[str],
         store_name: str = DEFAULT_STATE_STORE_NAME,
+        changelog_config: Optional[TopicConfig] = None,
```
gwaramadze: Did you mistakenly reorder the arguments? I think `store_type` should go after `store_name` and before `changelog_config`.

If not, then you need to re-order the docstring.

quixstreams/state/recovery.py
```diff
@@ -340,19 +340,19 @@ def recovering(self) -> bool:
...
-        topic_config: Optional[TopicConfig] = None,
+        topic_config: TopicConfig,
     ) -> Topic:
         """
         Register a changelog Topic with the TopicManager.
 
-        :param topic_name: source topic name
+        :param stream_id: stream id
         :param store_name: name of the store
         :param topic_config: a TopicConfig to use, optional.
```
gwaramadze:
```suggestion
        :param topic_config: a TopicConfig to use
```

tests/test_quixstreams/test_app.py
```diff
@@ -1264,9 +1234,11 @@ def count_and_fail(_, state: State):
         )
         state_manager.register_store(topic_in.name, "default")
```
gwaramadze: It works, but I think it will boost clarity if we use stream_id. let's replace 4 occurrences of `topic_in.name` with `sdf.stream_id`, WDYT?

tests/test_quixstreams/test_app.py
```diff
@@ -1373,11 +1351,13 @@ def test_clear_state(
         with state_manager:
             state_manager.register_store(topic_in_name, "default")
```
gwaramadze, line 1352: Similarly, here, for this test, a constant string like `stream_id = str(uuid.uuid4())` will be enough. And replace occurrences of `topic_in_name` with `stream_id`. This will render the `topic_factory` fixture useless in this test, so it may be removed. Additionally, the fixtures `producer`  and `executor` seem to be unused too, and they don't seem to set up any side effects because the test is green without them.

tests/test_quixstreams/test_app.py
```diff
@@ -1478,12 +1460,12 @@ def validate_state(stores):
                     state_manager.register_store(topic.name, store_name)
                     for p_num, count in partition_msg_count.items():
                         state_manager.on_partition_assign(
-                            topic=topic.name,
+                            stream_id=topic.name,
```
gwaramadze: Generally, could you give the tests another pass and check where it keeps passing topic instead of stream. These look like easy tweaks and will contribute to clarity for the future.

quixstreams/platforms/quix/topic_manager.py
```diff
@@ -59,7 +59,21 @@ def __init__(
...
+        if not topics:
+            raise ValueError("At least one Topic must be passed")
+        elif len(topics) == 1:
+            # If only one topic is passed, return its full name
+            # for backwards compatibility
+            return topics[0].name
+
+        # Use the "quix_name" to generate stream_id.
+        # In Quix Cloud, the "quix_name" can differ from the actual broker topic name
+        return "--".join(t.quix_name for t in topics)
```
gwaramadze: `name` vs `quix_name` is confusing.

`quix_name` is supposed to be prefixed with workspace id, right? So why do you call `name` as "full name"?

Looking at the `test_stream_id_from_topics_single_topic_prefixed_with_workspace` test I can see that in fact `Topic.name` is `"workspace_id-test1"`?? That is quite confusing; help me out here.

gwaramadze (commented):
Round 3

quixstreams/core/stream/stream.py
```diff
@@ -284,6 +284,27 @@ def add_transform(
         """
         return self._add(TransformFunction(func, expand=expand))  # type: ignore[call-overload]
 
+    def merge(self, other: Self) -> Self:
```
gwaramadze, line 287: For the public API, we have settled on `concat`. Should we rename this to concat, too, WDYT?

ghost: I'd keep `concat()` for DataFrames and `merge()` for Streams. 
As a name, `.concat()` makes sense only in DataFrames context.

quixstreams/dataframe/dataframe.py
ghost: Yep, I wanted to make stream_ids customizable, but later decided to postpone it.

quixstreams/dataframe/registry.py
ghost: Nope, forgot to resolve some todos

quixstreams/platforms/quix/topic_manager.py
ghost: I was also torn about that, but eventually decided to keep `quix_name` because this is the name returned by the Quix Platform, and I don't see how to abstract it away.
No other known platform does that. 
`quix_name` at least makes it clear where it comes from.

quixstreams/models/topics/manager.py
ghost: I agree with converting `*topics` to `topics`. This func is always called with the collection.
But I'd make the type to be `Iterable` since it's more than enough for this method (it doesn't really need to know that the set is being passed)

quixstreams/dataframe/dataframe.py
ghost: Yep, I like that

quixstreams/dataframe/dataframe.py
ghost: I'm making a few changes here:
1. The `self._topics` is now a tuple to guarantee immutability
2. `self._topics` is sorted in `StreamingDataFrame.__init__` mostly because I don't remember if there's anything else relying on it being sorted right now, and it's cheap.

quixstreams/state/manager.py
```diff
@@ -19,6 +19,7 @@
```
gwaramadze, line 19: In this file, you've got to update the docstring of `stores` property to say `{stream_id: {store_name: store}}`

quixstreams/dataframe/registry.py
ghost: I'd keep the more specific `get_topics` because the public property `consumer_topics` is in the same class.

quixstreams/dataframe/registry.py
ghost: Updated the docstring, although not sure if it makes it more clear

quixstreams/app.py
```diff
@@ -1030,37 +1026,32 @@ def _on_revoke(self, _, topic_partitions: List[TopicPartition]):
...
-        for tp in non_changelog_tps:
-            if self._state_manager.stores:
-                self._state_manager.on_partition_revoke(
-                    topic=tp.topic, partition=tp.partition
-                )
-            self._processing_context.on_partition_revoke(
-                topic=tp.topic, partition=tp.partition
-            )
+        self._revoke_state_partitions(topic_partitions=topic_partitions)
+        self._processing_context.on_partition_revoke()
```
gwaramadze, line 1030: This may be moved to `_revoke_state_partitions` too.

quixstreams/dataframe/dataframe.py
```diff
@@ -126,19 +132,27 @@ class StreamingDataFrame:
...
+        self._topics: tuple[Topic, ...] = tuple(
+            sorted({t.name: t for t in topics}.values(), key=attrgetter("name"))
+        )
         self._topic_manager = topic_manager
         self._registry = registry
         self._processing_context = processing_context
         self._producer = processing_context.producer
-        self._locked = False
+        self._registry.register_stream_id(
+            stream_id=self.stream_id, topic_names=[t.name for t in self._topics]
```
gwaramadze, line 154: I'm looking at the `self._topic_manager.stream_id_from_topics(self.topics)` call, which takes complete topics but only needs names. Same here, I think it will be slightly more readable if we pass full topics and let register_stream_id do the comprehension.

Or else, leave it be, but consider passing only names to `stream_id_from_topics`. (Reminder: topics are already sorted).

I mean, this is virtually the same case. I find it slightly dissonant that they are coded differently.

quixstreams/dataframe/dataframe.py
```diff
@@ -1467,6 +1506,44 @@ def _sink_callback(
...
+        on the new dataframe will create different state stores
+        unrelated to the original dataframes and topics.
+        The same is true for the repartition topics created by `.group_by()`.
+
+        :param other: other StreamingDataFrame
+        :return: a new StreamingDataFrame
+
+        """,
+    ) -> Self:
+        frame_
```
gwaramadze: matrix glitched 😉

quixstreams/dataframe/registry.py
gwaramadze: hey, unresolving because TODO is still here 🤔

quixstreams/dataframe/registry.py
gwaramadze: hey, unresolving because TODO is still here 🤔

quixstreams/dataframe/registry.py
gwaramadze: hey, unresolving because TODO is still here 🤔

quixstreams/models/topics/manager.py
```diff
@@ -201,147 +202,144 @@ def register(self, topic: Topic) -> Topic:
...
+        )
+
+    def stream_id_from_topics(self, topics: Sequence[Topic]) -> str:
+        """
+        Generate a stream_id by combining names of the provided topics.
+        """
+        if not topics:
+            raise ValueError("At least one Topic must be passed")
+
+        return "--".join(sorted(t.name for t in topics))
```
gwaramadze, line 343: The topics are already sorted. Do we want to be extra-cautious here?
```suggestion
        return "--".join(t.name for t in topics)
```

tests/test_quixstreams/test_dataframe/test_registry.py
```diff
@@ -0,0 +1,29 @@
...
+from quixstreams.models import Topic
+
+
+class TestDataFrameRegistry:
+    def test_register_root_multi_topic_sdf_fails(self):
+        registry = DataFrameRegistry()
+        sdf_mock = MagicMock(spec_set=StreamingDataFrame)
+        type(sdf_mock).topics = PropertyMock(
+            return_value=[Topic("test-1"), Topic("test-1")]
+        )
```
gwaramadze: This may be simplified
```suggestion
        sdf_mock = mock.Mock(topics=[Topic("test-1"), Topic("test-1")])
```

quixstreams/app.py
ghost: `self._processing_context.on_partition_revoke()` is not related to the state

quixstreams/dataframe/dataframe.py
ghost: I don't even know how it's happened :)

quixstreams/dataframe/dataframe.py
ghost: The logic behind this is that `register_stream_id` only needs names.
It doesn't need the entire Topic object or its properties, and passing `Topic` would create an additional dependency.

quixstreams/models/topics/manager.py
ghost: I'd rather be extra cautious to make this part more robust.

gwaramadze (approved):

PR #811, opened 2025-04-02
InfluxDB3Sink: improve timestamp type handling

tim-quix:
Changed some behavior around timestamp handling so when non-int timestamps are used, they could properly evaluate the min and max.

Also added further validation that helps with errors that are not super obvious to end user.

### DIFF
--- a/quixstreams/sinks/core/influxdb3.py
+++ b/quixstreams/sinks/core/influxdb3.py
@@ -297,7 +297,6 @@ class InfluxDB3Sink(BatchingSink):
                 else:
                     ts = value[time_key]
                     # Note: currently NOT validating the timestamp itself is valid
-                    # Also, int length must correspond to the passed `time_precision`
                     if not isinstance(ts, valid := (str, int, datetime)):
                         raise TypeError(
                             f'InfluxDB3 "time" field expects: {valid}, got {type(ts)}'
--- a/quixstreams/sinks/core/influxdb3.py
+++ b/quixstreams/sinks/core/influxdb3.py
@@ -28,6 +28,12 @@ logger = logging.getLogger(__name__)
 
 
 TimePrecision = Literal["ms", "ns", "us", "s"]
+TIME_PRECISION_LEN = {
+    "s": 10,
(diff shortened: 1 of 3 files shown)


quixstreams/sinks/core/influxdb3.py
```diff
@@ -265,6 +261,25 @@ def write(self, batch: SinkBatch):
...
                         tags[tag_key] = tag
 
+                if ts is None:
+                    ts = item.timestamp
+                else:
+                    # Note: currently NOT validating the timestamp itself is valid
+                    if not isinstance(ts, valid := (str, int, datetime)):
+                        raise TypeError(
+                            f'InfluxDB3 "time" field expects: {valid}, got {type(ts)}'
+                        )
```
gwaramadze:
```suggestion
                # Note: currently NOT validating the timestamp itself is valid
                elif not isinstance(ts, valid := (str, int, datetime)):
                    raise TypeError(
                        f'InfluxDB3 "time" field expects: {valid}, got {type(ts)}'
                    )
```

quixstreams/sinks/core/influxdb3.py
```diff
@@ -281,20 +296,19 @@ def write(self, batch: SinkBatch):
 
                 if self._convert_ints_to_floats:
                     fields = {
-                        k: float(v) if isinstance(v, int) else v
+                        k: float(v) if type(v) is int else v  # avoids bool matching
```
gwaramadze, line 298: can you explain this change to me?

tim-quix: Ah yes, welcome to one of the fun little oddities I stumbled upon:

```
>>> isinstance(True, int)
True
```

Booleans are a subclass of `int`...which makes sense when you think about it, but I would have never realized it without trying to do what I was originally doing here.

...and of course, I don't want to convert the booleans to floats haha

gwaramadze: image

gwaramadze (approved):

PR #836, opened 2025-04-14
Optimize group by for single partition topics

ovv:
Group by operations on topics with a single partition are now optimized to avoid creating a repartition topic. Instead, the messages are directly transformed to use the new key, as all messages go to the same partition.


quixstreams/dataframe/dataframe.py
```diff
@@ -135,15 +135,20 @@ def __init__(
         registry: DataFrameRegistry,
         processing_context: ProcessingContext,
         stream: Optional[Stream] = None,
+        stream_id: Optional[str] = None,
```
gwaramadze, line 138: Just a note. This proves that we need to rename it to `state_id`.
1. Passing both `stream` and `stream_id` is a rather strange API.
2. The `Stream` class does not feature any `id` itself - counterintuitive.

quixstreams/dataframe/registry.py
```diff
@@ -61,7 +61,10 @@ def register_root(
         self._registry[topic.name] = dataframe.stream
 
     def register_groupby(
-        self, source_sdf: "StreamingDataFrame", new_sdf: "StreamingDataFrame"
+        self,
+        source_sdf: "StreamingDataFrame",
+        new_sdf: "StreamingDataFrame",
+        register_new_root: bool = True,
```
gwaramadze, line 76: Add a docstring line for it

quixstreams/dataframe/registry.py
```diff
@@ -72,16 +75,26 @@ def register_groupby(
...
             )
+
         self._repartition_origins.add(new_sdf.stream_id)
 
+        if register_new_root:
+            try:
+                self.register_root(new_sdf)
+            except StreamingDataFrameDuplicate:
+                raise GroupByDuplicate(
+                    "A `SDF.group_by()` operation appears to be the same as another, "
```
gwaramadze:
```suggestion
                    "An `SDF.group_by()` operation appears to be the same as another, "
```

Same applies to line 81

tests/test_quixstreams/test_dataframe/test_dataframe.py
```diff
@@ -1676,8 +1683,11 @@ def test_group_by_column(
         sdf[col] = col_update
 
         groupby_topic = sdf.topics[0]
-        assert sdf_registry.consumer_topics == [topic, groupby_topic]
-        assert groupby_topic.name.startswith("repartition__")
+        if num_partitions == 1:
+            assert sdf_registry.consumer_topics == [topic]
```
gwaramadze, line 1687: You can move `groupby_topic = sdf.topics[0]` to else clause or add `assert groupby_topic == topic` to if clause.
```suggestion
            assert sdf_registry.consumer_topics == [topic]
            assert groupby_topic == topic
```

gwaramadze (dismissed):

quixstreams/dataframe/dataframe.py
```diff
@@ -606,6 +616,29 @@ def func(d: dict, state: State):
...
+                return value, key(value), timestamp, headers
+
+        stream = self.stream.add_transform(_callback, expand=False)
+
+        groupby_sdf = self.__dataframe_clone__(
+            stream=stream, stream_id=f"{self.stream_id}--groupby--{operation}"
+        )
+        self._registry.register_groupby(
+            source_sdf=self, new_sdf=groupby_sdf, register_new_root=False
+        )
```
ghost, line 638: I'm not sure we even need `.register_groupby()` here since the operation doesn't create a new topic.
We can treat it as a stateless operation, and keep the same `stream_id` too.

ghost (changes requested):
See the comments

quixstreams/dataframe/dataframe.py
ovv: > I'm not sure we even need .register_groupby() here since the operation doesn't create a new topic.

We do. It ensure nested group-by or multiple group-by on the same stream ID are disallowed. While not necessary for the single partition optimisation we want similar behaviour with the not optimised path to ensure the user code is valid no matter the number of partitions in use.

> We can treat it as a stateless operation, and keep the same stream_id too.

No, we need the state to be different after the group-by to keep the same behaviour

ghost (approved):

PR #843, opened 2025-04-15
CI: Bump Python version to 3.13

ovv:


gwaramadze:
AFAIK `requirements.txt` is fine, but `tests/requirements.txt` won't install.

ovv:
yeah, there is no wheel for the pyarrow version we need. I haven't look at why we are stuck on that version and can't use the latest

gwaramadze (approved):

PR #848, opened 2025-04-18
Fix key deserialization propagation in windows

gwaramadze:
Despite `key_deserializer` being set to `json`, after windowing, the deserialization is gone, and the key is served back as bytes. Affected windows:
* sliding
* tumbling and hopping in closing strategy "partition" mode

Tumbling and hopping windows with the closing strategy "key" are unaffected.

* The fix for the sliding window was rather trivial. All we need is to use the deserialized key coming directly from `process_window` instead of the serialized one coming from the store.
  * As a side effect, turns out nothing is utilizing the key from `state.expire_windows`, so it doesn't need to return it.
* The fix for the "partition" expiration mode is less elegant; refer to the docstring of `_deserialize_prefix` for a detailed explanation.

### DIFF
--- a/quixstreams/state/rocksdb/windowed/transaction.py
+++ b/quixstreams/state/rocksdb/windowed/transaction.py
@@ -9,11 +9,13 @@ from quixstreams.state.base.transaction import (
     PartitionTransactionStatus,
     validate_transaction_status,
 )
+from quixstreams.state.exceptions import StateSerializationError
 from quixstreams.state.metadata import DEFAULT_PREFIX, SEPARATOR
 from quixstreams.state.recovery import ChangelogProducer
 from quixstreams.state.serialization import (
     DumpsFunc,
     LoadsFunc,
+    deserialize,
     serialize,
 )
 from quixstreams.state.types import ExpiredWindowDetail, WindowDetail
@@ -360,7 +362,8 @@ class WindowedRocksDBPartitionTransaction(PartitionTransaction):
                             end=end,
                             prefix=prefix,
                         )
-                    yield (start, end), aggregated, collected, prefix
+                    deserialized_prefix = self._deserialize_prefix(prefix)
+                    yield (start, end), aggregated, collected, deserialized_prefix
 
         else:
             # If we don't have a saved last_expired value it means one of two cases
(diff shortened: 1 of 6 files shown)


quixstreams/state/rocksdb/windowed/transaction.py
```diff
@@ -394,6 +402,32 @@ def expire_all_windows(
...
+        Attempt to deserialize a window prefix.
+
+        Window prefixes can be provided either as raw bytes or as other types
+        (e.g., dict). The `as_state()` method conditionally serializes these
+        prefixes to bytes only if they are not already bytes before storing.
+
+        When retrieving a prefix during partition-level windows expiration, we
+        don't know its original type due to this conditional serialization.
+        Therefore, we must first *try* to deserialize it using the configured
+        `loads` function.
```
ovv, line 416: Could we store an indicator about the prefix type, if it's a byte or something else ?

For migration, existing windows missing that information can try to deserialize

gwaramadze: Yes, but it's considerably more code, not sure if it's worth it.

Let's see what @daniil-quix thinks. But even if we decide to do this, it may be done as a separate optimisation; in the meantime, this will fix the bug.

ovv: I'm mostly worried about the performance hit of trying to deserialise every keys

tests/test_quixstreams/test_state/test_rocksdb/test_windowed/test_transaction.py
```diff
@@ -328,9 +328,9 @@ def test_expire_windows_multiple_windows(self, windowed_rocksdb_store_factory):
             )
 
         assert len(expired) == 3
-        assert expired[0] == ((0, 10), 1, [], prefix)
-        assert expired[1] == ((10, 20), 1, [], prefix)
-        assert expired[2] == ((20, 30), 1, [], prefix)
```
ghost, line 333: This was kinda testing that the same prefix is returned :)
Let's return the key from `expire_windows()` to make the tests legit, even though we don't use it in the actual code.

quixstreams/state/rocksdb/windowed/transaction.py
ghost: I did some quick tests locally, and the performance hit actually depends on the value itself.

Two almost identical values `b'123123'` and `b'{123123` take very different amounts of time:

```
### Successful deserialization

%%timeit

try:
    orjson.loads(b'123123')
except Exception:
    ...

34.4 ns ± 0.182 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```


```
### Deserialization failed
%%timeit

try:
    orjson.loads(b'{123123')
except Exception:
    ...

765 ns ± 4.51 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

PR #853, opened 2025-04-22
Merge PausingManager into RowConsumer

ghost:
- The backpressure handling logic is moved directly to RowConsumer
- Replaced some usages of Consumer by RowConsumer in tests and fixtures


quixstreams/app.py
```diff
@@ -1029,7 +1025,8 @@ def _on_revoke(self, _, topic_partitions: List[TopicPartition]):
             self._processing_context.commit_checkpoint(force=True)
 
         self._revoke_state_partitions(topic_partitions=topic_partitions)
-        self._processing_context.on_partition_revoke()
+        self._consumer.reset_backpressure()
+        # self._processing_context.on_partition_revoke()
```
gwaramadze: should be removed?

quixstreams/app.py
```diff
@@ -1038,7 +1035,8 @@ def _on_lost(self, _, topic_partitions: List[TopicPartition]):
         logger.debug("Rebalancing: dropping lost partitions")
 
         self._revoke_state_partitions(topic_partitions=topic_partitions)
-        self._processing_context.on_partition_revoke()
+        # self._processing_context.on_partition_revoke()
```
gwaramadze: should be removed?

quixstreams/rowconsumer.py
```diff
@@ -149,4 +158,77 @@ def poll_row(self, timeout: Optional[float] = None) -> Union[Row, List[Row], Non
...
+        """
+        if self._backpressure_resume_at > monotonic():
+            return
+
+        # Resume the previously backpressured TPs
+        for tp in self._backpressurred_tps:
+            logger.debug(f'Resuming topic partition "{tp.topic}[{tp.partition}]"')
+            self.resume(
+                partitions=[TopicPartition(topic=tp.topic, partition=tp.partition)]
+            )
```
gwaramadze: Is there a reason we don't want to pass the existing tp object?
```suggestion
            self.resume(partitions=[tp])
```

quixstreams/rowconsumer.py
```diff
@@ -149,4 +158,77 @@ def poll_row(self, timeout: Optional[float] = None) -> Union[Row, List[Row], Non
...
+
+        # Pause only data TPs excluding changelog TPs
+        non_changelog_topics = {
+            k: v for k, v in self._topics.items() if not v.is_changelog
+        }
+        non_changelog_tps = [
+            tp for tp in self.assignment() if tp.topic in non_changelog_topics
+        ]
+
+        for tp in non_changelog_tps:
```
gwaramadze: Try this:
```suggestion
        changelog_topics = {k for k, v in self._topics.items() if v.is_changelog}
        for tp in self.assignment():
            # Pause only data TPs excluding changelog TPs
            if tp.topic in changelog_topics:
                continue
                
```

tests/test_quixstreams/fixtures.py
```diff
@@ -72,19 +79,12 @@ def factory(
         auto_commit_enable: bool = True,
         extra_config: dict = None,
```
gwaramadze: BTW, this seems wrong
```suggestion
        extra_config: Optional[dict] = None,
```

quixstreams/rowconsumer.py
ghost: Not really. 
The original `TopicPartition` object may have an additional attribute set (e.g. `offset`), but `.resume()` should ignore them.

I'll update that

gwaramadze (approved):

PR #857, opened 2025-04-28
Implement timestamp alignment by buffering the incoming Kafka messages

ghost:
This PR implements timestamp alignment to consume messages in-order across multiple partitions in the same number.

It is needed for `StreamingDataFrame.concat()` and upcoming Joins features.

## Why is it needed?
The usual Kafka consumer reads and returns messages in batches per partition. 
If you have two partitions TP1 and TP2 with 1000 messages each, the actual order of processing can look like this:
- Read and process 1000 messages from TP1
- Read and process 1000 messages from TP2

This works fine when these partitions are processed independently. 

But when they are joined, the messages are processed out-of-order, breaking things like windowed aggregations for joined or concatenated streams.

To solve that, this PR adds additional buffering of messages on the `RowConsumer` side.

### When buffering is used
**Note: currently, buffering kicks in only when `StreamingDataFrame.concat(...)` is used in the topology.**

There are two reasons for that:
1. For most simple use cases, when only one topic is consumed, buffering is not really needed, and even passthrough adds overhead
2. For use case of one topic + groupby, buffering leads to stuttering in the pipeline, because faster input topic needs to wait for inherently slower repartition topic, even though the timestamp order between these topics does not matter.

### Implementation

The new `InternalConsumerBuffer` class represents a buffer that groups incoming messages per partition numbers and allows reading them in the timestamp order across multiple TPs with the same number.

- The `RowConsumer` reads messages in batches from the broker
- Then, it feeds the batch to the internal buffer.
- The internal buffer communicates the state of the buffered TPs back to the `RowConsumer` via `pause_full()` and `resume_empty()` methods to pause or resume the individual partitions. 
This allows to balance reads across the assignment, and avoid processing one partition in the group before the others
- The consumer pops the next message from the buffer and returns it to the application to be processed.
   - When multiple topics are processed, the buffer picks the partition **with the smallest timestamp of the next message.**
      This way, messages are still processed in the offset order within the partition, but the partition with the smaller timestamps is processed before the others.
        - When one of the partitions in the group is empty, its high watermark is considered to determine whether the broker has more messages for it, or whether it's idle (no new messages to process). 
        - If the partition is idle, it is resumed, and the other partitions won't wait for it.
        - If the partition is empty but not idle, the buffer won't return a message and will wait for the empty buffer to fill up.

- The `max_partition_buffer_size` config parameter determines the size of the individual buffers. Default - `10000`

## Limitations
- Custom `timestamp_extractor` is not supported yet, and the buffering code always uses original Kafka timestamps. This will be implemented later.
- The in-order reading is not guaranteed for the messages that _are not been produced yet._ 
   If the producer sends more data to some empty partition later, it may be processed out-of-order, because the application moved on with other partitions (the empty TP was idle).

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -1611,9 +1611,11 @@ class StreamingDataFrame:
         """
 
         merged_stream = self.stream.merge(other.stream)
-        return self.__dataframe_clone__(
+        sdf = self.__dataframe_clone__(
             *self.topics, *other.topics, stream=merged_stream
         )
+        self._registry.require_copartitioning()
+        return sdf
 
     def ensure_topics_copartitioned(self):
         partitions_counts = set(t.broker_config.num_partitions for t in self._topics)
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -1611,11 +1611,10 @@ class StreamingDataFrame:
         """
 
         merged_stream = self.stream.merge(other.stream)
-        sdf = self.__dataframe_clone__(
+        self._registry.require_time_alignment()
+        return self.__dataframe_clone__(
             *self.topics, *other.topics, stream=merged_stream
         )
-        self._registry.require_time_alignment()
-        return sdf
 
     def ensure_topics_copartitioned(self):
         partitions_counts = set(t.broker_config.num_partitions for t in self._topics)
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -1614,7 +1614,7 @@ class StreamingDataFrame:
         sdf = self.__dataframe_clone__(
             *self.topics, *other.topics, stream=merged_stream
         )
-        self._registry.require_copartitioning()
+        self._registry.require_time_alignment()
         return sdf
 
     def ensure_topics_copartitioned(self):
--- a/quixstreams/rowconsumer/__init__.py
+++ b/quixstreams/rowconsumer/__init__.py
@@ -1,0 +1,1 @@
+from .consumer import RowConsumer as RowConsumer
--- a/quixstreams/rowconsumer/buffering.py
+++ b/quixstreams/rowconsumer/buffering.py
@@ -1,7 +1,8 @@
+import enum
 import logging
 from collections import deque
 from operator import attrgetter
-from typing import Iterable, Optional, Union
+from typing import Iterable, Optional
 
 from confluent_kafka import TopicPartition
 
@@ -12,6 +13,12 @@ logger = logging.getLogger(__name__)
 __all__ = ("InternalConsumerBuffer",)
 
 _next_timestamp_getter = attrgetter("next_timestamp")
+
+
+class Idleness(enum.Enum):
+    IDLE = 1  # The partition is idle and has no new messages to be consumed
+    ACTIVE = 2  # The partition has more messages to be consumed
+    UNKNOWN = 3  # The idleness is unknown
 
 
 class PartitionBuffer:
@@ -46,7 +53,7 @@ class PartitionBuffer:
         """
         self.high_watermark = offset
 
-    def idle(self) -> Union[bool, None]:
+    def idleness(self) -> Idleness:
         """
         Check if the partition is idle or has more data to be consumed from the broker.
 
@@ -58,9 +65,11 @@ class PartitionBuffer:
         high_watermark = self.high_watermark
         if high_watermark < 0:
             # There's no valid highwater for this partition yet, the idleness is unknown
-            return None
+            return Idleness.UNKNOWN
 
-        return self.max_offset + 1 >= high_watermark
+        return (
+            Idleness.IDLE if self.max_offset + 1 >= high_watermark else Idleness.ACTIVE
+        )
 
     def append(self, message: SuccessfulConfluentKafkaMessageProto):
         """
@@ -224,7 +233,7 @@ class PartitionBufferGroup:
             # but not idle.
             # Wait until new messages are fetched or the highwater is set.
             for buffer in buffers:
-                if buffer.empty() and buffer.idle() != True:
+                if buffer.empty() and buffer.idleness() != Idleness.IDLE:
                     return None
 
             buffer = min(buffers, key=_next_timestamp_getter)
@@ -247,7 +256,11 @@ class PartitionBufferGroup:
         for buffer in self._partition_buffers_values:
             # Pause consuming new messages for this partition
             # when the buffer is paused, full and not idle
-            if not buffer.paused and buffer.full() and buffer.idle() == False:
+            if (
+                not buffer.paused
+                and buffer.full()
+                and buffer.idleness() == Idleness.ACTIVE
+            ):
                 buffer.pause()
                 tps.append((buffer.topic, buffer.partition))
         return tps
@@ -262,7 +275,9 @@ class PartitionBufferGroup:
         for buffer in self._partition_buffers_values:
             # Resume consuming new messages for this partition
             # when the buffer is paused and either idle or not full
-            if buffer.paused and (buffer.idle() != False or not buffer.full()):
+            if buffer.paused and (
+                buffer.idleness() != Idleness.ACTIVE or not buffer.full()
+            ):
                 buffer.resume()
                 tps.append((buffer.topic, buffer.partition))
 
--- a/quixstreams/rowconsumer/buffering.py
+++ b/quixstreams/rowconsumer/buffering.py
@@ -304,7 +304,7 @@ class InternalConsumerBuffer:
         self._partition_groups: dict[int, PartitionBufferGroup] = {}
         self._max_partition_buffer_size = max_partition_buffer_size
 
-    def on_assign(self, topic_partitions: list[TopicPartition]):
+    def assign_partitions(self, topic_partitions: list[TopicPartition]):
         """
         Assign new partitions to the buffer.
 
@@ -322,7 +322,7 @@ class InternalConsumerBuffer:
             )
             partition_group.assign_partition(topic=tp.topic)
 
-    def on_revoke(self, topic_partitions: list[TopicPartition]):
+    def revoke_partitions(self, topic_partitions: list[TopicPartition]):
         """
         Drop the partitions from the buffer.
 
--- a/quixstreams/rowconsumer/buffering.py
+++ b/quixstreams/rowconsumer/buffering.py
@@ -270,8 +270,7 @@ class PartitionBufferGroup:
         return tps
 
     def clear(self, topic: str):
-        buffer = self._partition_buffers.get(topic)
-        if buffer is not None:
+        if buffer := self._partition_buffers.get(topic):
             buffer.clear()
 
 
--- a/quixstreams/rowconsumer/buffering.py
+++ b/quixstreams/rowconsumer/buffering.py
@@ -176,10 +176,9 @@ class PartitionBufferGroup:
         if topic in self._partition_buffers:
             return
         # New partition, need to create a new buffer and add it to the heap
-        partition_buffer = PartitionBuffer(
+        self._partition_buffers[topic] = PartitionBuffer(
             partition=self._partition, topic=topic, max_size=self._max_size
         )
-        self._partition_buffers[topic] = partition_buffer
 
     def revoke_partition(self, topic: str):
         """
--- a/quixstreams/rowconsumer/buffering.py
+++ b/quixstreams/rowconsumer/buffering.py
@@ -308,14 +308,18 @@ class InternalConsumerBuffer:
         How it works:
 
         - The buffer gets partitions assigned.
-        - The Consumer feeds messages to the buffer along with the high watermarks for all assigned partitions.
+        - The Consumer feeds messages to the buffer along with the high watermarks for all
+            assigned partitions.
         - The Consumer calls `.pop()` to get the next message to be processed.
-            If multiple partitions with the same number are assigned, the message will be popped from the partition with the smallest next timestamp, providing in-order reads.
-        - The Consumer calls `.pause_full()` and `.resume_empty()` methods to balance the reads across all partitions.
+            If multiple partitions with the same number are assigned, the message will be
+            popped from the partition with the smallest next timestamp, providing in-order reads.
+        - The Consumer calls `.pause_full()` and `.resume_empty()` methods to balance the reads
+            across all partitions.
 
-        :param max_partition_buffer_size: the maximum size of the individual topic partition buffer when the buffer is considered full.
-            It is a soft limit, and it may be exceeded in some cases.
-            When individual buffer exceeds this limit, its TP can be paused to let other partitions to be consumed too.
+        :param max_partition_buffer_size: the maximum size of the individual topic partition
+            buffer when the buffer is considered full. It is a soft limit, and it may be exceeded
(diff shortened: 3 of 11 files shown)


quixstreams/rowconsumer/__init__.py
```diff
@@ -0,0 +1 @@
+from .consumer import RowConsumer as RowConsumer
```
gwaramadze, line 1:
```suggestion
from .consumer import RowConsumer
```

quixstreams/rowconsumer/consumer.py
```diff
@@ -1,38 +1,58 @@
...
+            to_suppress = on_error(exc, message, logger)
+            if to_suppress:
+                continue
+            raise
 
 
 class RowConsumer(BaseConsumer):
     _backpressure_resume_at: float
 
+    # TODO: Rename it to "InternalConsumer"? Or "InternalBufferedConsumer"?
```
gwaramadze, line 43: For the sake of simplicity, we don't want to deal with multiple consumers, so just "internal" is a good umbrella description for a single consumer class we don't want to advertise to the user. ~I'd do one step further and rename the `BaseConsumer` to `Consumer` as it is (almost) the same thing, and then do `InternalConsumer(Consumer)`.~

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        - The Consumer calls `.pause_full()` and `.resume_empty()` methods to balance the reads across all partitions.
+
+        :param max_partition_buffer_size: the maximum size of the individual topic partition buffer when the buffer is considered full.
+            It is a soft limit, and it may be exceeded in some cases.
+            When individual buffer exceeds this limit, its TP can be paused to let other partitions to be consumed too.
+        """
+        self._partition_groups: dict[int, PartitionBufferGroup] = {}
+        self._max_partition_buffer_size = max_partition_buffer_size
+
+    def on_assign(self, topic_partitions: list[TopicPartition]):
```
gwaramadze: I don't like callback-like names for methods that are technically not a callback. I'd be more specific here.
```suggestion
    def assign_partitions(self, topic_partitions: list[TopicPartition]):
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        for tp in topic_partitions:
+            partition_group = self._partition_groups.setdefault(
+                tp.partition,
+                PartitionBufferGroup(
+                    partition=tp.partition, max_size=self._max_partition_buffer_size
+                ),
+            )
+            partition_group.assign_partition(topic=tp.topic)
+
+    def on_revoke(self, topic_partitions: list[TopicPartition]):
```
gwaramadze: Same here, especially given that this is called from callbacks `on_revoke` AND `on_lost`. A more generic name will be easier to follow.
```suggestion
    def revoke_partitions(self, topic_partitions: list[TopicPartition]):
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+    def __init__(self, max_partition_buffer_size: int = 10000):
+        """
+        A buffer to align messages across different topics by timestamps and consume them
+        in-order across partitions.
+
+        Under the hood, this class groups buffered messages by partition and
+        provides API to get the message with the smallest timestamp across all assigned
+        topics with the same partition number.
+
+
```
gwaramadze:
```suggestion
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        - The buffer gets partitions assigned.
+        - The Consumer feeds messages to the buffer along with the high watermarks for all assigned partitions.
+        - The Consumer calls `.pop()` to get the next message to be processed.
+            If multiple partitions with the same number are assigned, the message will be popped from the partition with the smallest next timestamp, providing in-order reads.
+        - The Consumer calls `.pause_full()` and `.resume_empty()` methods to balance the reads across all partitions.
+
+        :param max_partition_buffer_size: the maximum size of the individual topic partition buffer when the buffer is considered full.
+            It is a soft limit, and it may be exceeded in some cases.
+            When individual buffer exceeds this limit, its TP can be paused to let other partitions to be consumed too.
```
gwaramadze: Lines a bit too long
```suggestion
        - The Consumer feeds messages to the buffer along with the high watermarks for all
            assigned partitions.
        - The Consumer calls `.pop()` to get the next message to be processed.
            If multiple partitions with the same number are assigned, the message will be
            popped from the partition with the smallest next timestamp, providing in-order reads.
        - The Consumer calls `.pause_full()` and `.resume_empty()` methods to balance the reads
            across all partitions.

        :param max_partition_buffer_size: the maximum size of the individual topic partition
            buffer when the buffer is considered full. It is a soft limit, and it may be exceeded
            in some cases. When individual buffer exceeds this limit, its TP can be paused
            to let other partitions to be consumed too.
```

quixstreams/rowconsumer/consumer.py
```diff
@@ -225,5 +283,68 @@ def resume_backpressured(self):
...
+                raise
+            messages = []
+
+        # Get the recent cached high watermarks
+        high_watermarks: dict[tuple[str, int], int] = {}
+        for tp in self.assignment():
+            topic_obj = self._topics[tp.topic]
+            if not topic_obj.is_changelog:
+                _, high = self.get_watermark_offsets(partition=tp, cached=True)
+                high_watermarks[(tp.topic, tp.partition)] = high
```
gwaramadze, line 349: Not sure how this would affect the performance, but from a readability perspective, this may be done as `TopicPartition` is a hashable object.
```suggestion
                high_watermarks[tp] = high
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        for message in messages:
+            partition_group = self._partition_groups[message.partition()]
+            partition_group.append(message=message)
+
+        for partition_group in self._partition_groups.values():
+            group_watermarks = {
+                topic: watermark
+                for (topic, partition), watermark in high_watermarks.items()
+                if partition == partition_group.partition
```
gwaramadze, line 375: If you accept the suggestion in `_feed_buffer`, then this is an accompanying change
```suggestion
                tp.topic: watermark
                for tp, watermark in high_watermarks.items()
                if tp.partition == partition_group.partition
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        if len(buffers) == 1:
+            # Use iter() here to avoid creating a new list each time
+            buffer = next(iter(buffers))
+            return buffer.popleft()
+        elif len(buffers) > 1:
+            # There's more than one partition in the group and one of them is empty
+            # but not idle.
+            # Wait until new messages are fetched or the highwater is set.
+            for buffer in buffers:
+                if buffer.empty() and buffer.idle() != True:
```
gwaramadze: slightly more idiomatic
```suggestion
                if buffer.empty() and buffer.idle() is not True:
```

gwaramadze (commented):
Let me drop this batch. I will keep on digging for a while longer.

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+
+from confluent_kafka import TopicPartition
+
+from quixstreams.models import SuccessfulConfluentKafkaMessageProto
+
+logger = logging.getLogger(__name__)
+
+__all__ = ("InternalConsumerBuffer",)
+
+
```
gwaramadze:
```suggestion
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        self.next_timestamp = float("inf")
+        self._messages: deque[SuccessfulConfluentKafkaMessageProto] = deque()
+
+    def set_high_watermark(self, offset: int):
+        """
+        Set high watermark offset of topic partition.
+
+        :param offset: high watermark offset.
+        """
+        self.high_watermark = offset
```
gwaramadze: if `PartitionBuffer.high_watermark` is not private then hiding a setter behind a function call is a redundancy. Delete the whole method or rename `high_watermark` to `_high_watermark`. The former will be more performant since we ditch the middleman (function call).

PS. same goes for `pause` and `resume` since those are a single set operation methods too.

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+        :param max_size: the maximum size of the buffer when the buffer is considered full.
+            It is a soft limit, and it may be exceeded in some cases
+        """
+        self.partition = partition
+        self.topic = topic
+        self.max_size = max_size
+        self.high_watermark = -1001
+        self.max_offset = -1
+        self.paused = False
+        self.next_timestamp = float("inf")
```
gwaramadze, line 41: Why can't the undefined default be just None?
```suggestion
        self.next_timestamp: Optional[float] = None
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+
+        :param partition: partition number.
+        :param topic: topic name.
+        :param max_size: the maximum size of the buffer when the buffer is considered full.
+            It is a soft limit, and it may be exceeded in some cases
+        """
+        self.partition = partition
+        self.topic = topic
+        self.max_size = max_size
+        self.high_watermark = -1001
```
gwaramadze: Please add a comment explaining why it is -1001 as it's not a usual default value. Also why can't the undefined default be just None?
```suggestion
        self.high_watermark: Optional[int] = None
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+
+        :param topic: topic name.
+        """
+        if topic in self._partition_buffers:
+            return
+        # New partition, need to create a new buffer and add it to the heap
+        partition_buffer = PartitionBuffer(
+            partition=self._partition, topic=topic, max_size=self._max_size
+        )
+        self._partition_buffers[topic] = partition_buffer
```
gwaramadze:
```suggestion
        self._partition_buffers[topic] = PartitionBuffer(
            partition=self._partition, topic=topic, max_size=self._max_size
        )
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+            # when the buffer is paused and either idle or not full
+            if buffer.paused and (buffer.idle() != False or not buffer.full()):
+                buffer.resume()
+                tps.append((buffer.topic, buffer.partition))
+
+        return tps
+
+    def clear(self, topic: str):
+        buffer = self._partition_buffers.get(topic)
+        if buffer is not None:
```
gwaramadze:
```suggestion
        if buffer := self._partition_buffers.get(topic):
```

quixstreams/rowconsumer/buffering.py
```diff
@@ -0,0 +1,403 @@
...
+            partition_group.set_high_watermarks(offsets=group_watermarks)
+
+    def pop(self) -> Optional[SuccessfulConfluentKafkaMessageProto]:
+        """
+        Pop the next message from the buffer in the timestamp order.
+
+        :returns: `None` if all the buffers are empty or if the
+        """
+        # Iterate over group buffers and return the first messages we find
+        for partition_group in self._partition_groups.values():
```
gwaramadze, line 386: Do you want to make a shortcut `self._partition_groups_values` like you did with `_partition_buffers_values`?

quixstreams/rowconsumer/consumer.py
```diff
@@ -1,38 +1,58 @@
...
 
-_MAX_FLOAT = sys.float_info.max
+
+def _validate_message_batch(
+    messages: list[RawConfluentKafkaMessageProto], on_error: ConsumerErrorCallback
+) -> Iterator[SuccessfulConfluentKafkaMessageProto]:
+    for message in messages:
+        try:
+            message = raise_for_msg_error(message)
+            yield message
```
gwaramadze, line 32:
```suggestion
            yield raise_for_msg_error(message)
```

quixstreams/rowconsumer/__init__.py
ghost: `ruff` barks if it's not done this way.

```
quixstreams/rowconsumer/__init__.py:1:23: F401 `.consumer.RowConsumer` imported but unused; consider removing, adding to `__all__`, or using a redundant alias
  |
1 | from .consumer import RowConsumer
  |                       ^^^^^^^^^^^ F401
  |
  = help: Use an explicit re-export: `RowConsumer as RowConsumer
```

quixstreams/rowconsumer/consumer.py
ghost: Yep, it'll be just `InternalConsumer`

quixstreams/dataframe/dataframe.py
```diff
@@ -1611,9 +1611,11 @@ def concat(self, other: "StreamingDataFrame") -> "StreamingDataFrame":
         """
 
         merged_stream = self.stream.merge(other.stream)
-        return self.__dataframe_clone__(
+        sdf = self.__dataframe_clone__(
             *self.topics, *other.topics, stream=merged_stream
         )
+        self._registry.require_time_alignment()
+        return sdf
```
gwaramadze: Is there a gotcha in tweaking the registry before sdf clone?
```suggestion
        self._registry.require_time_alignment()
        return self.__dataframe_clone__(
            *self.topics, *other.topics, stream=merged_stream
        )
```

quixstreams/rowconsumer/consumer.py
ghost: I like the current way better because it separates "yield" and "raise_***", feels easier to follow

quixstreams/rowconsumer/__init__.py
gwaramadze: oh this, then we should add `__all__`

quixstreams/rowconsumer/buffering.py
ghost: Replaced `True/False/None`  with a proper enum

quixstreams/rowconsumer/buffering.py
ghost: Agree

quixstreams/rowconsumer/buffering.py
ghost: Because it is used when sorting partitions to pick the one with the smallest timestamp, and `inf` is more comfortable here.
Otherwise, you'd need to do `is None` checks, because you can't sort `Nones` and `floats`

quixstreams/rowconsumer/buffering.py
ghost: The consumer returns the high watermark this way if there's no message in the partition yet.
I think it's better to keep the same type than to use `None` here

quixstreams/rowconsumer/consumer.py
ghost: I prefer to keep it this way to avoid using `confluent_kafka.TopicPartition` in the `PartitionBufferGroup` class too.

quixstreams/rowconsumer/buffering.py
ghost: Agree, updated

gwaramadze (approved):

PR #867, opened 2025-05-07
PostgreSQLSink: allow dynamic table name selection based on record

tim-quix:
Also allows setting a schema.

Requested by Tomas.

### DIFF
--- a/quixstreams/sinks/community/postgresql.py
+++ b/quixstreams/sinks/community/postgresql.py
@@ -1,7 +1,7 @@
 import logging
 from datetime import datetime
 from decimal import Decimal
-from typing import Any, Mapping, Optional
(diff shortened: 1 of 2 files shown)


quixstreams/sinks/community/postgresql.py
```diff
@@ -178,15 +183,15 @@ def _init_table(self):
             )
             """
         ).format(
-            table=sql.Identifier(self.table_name),
+            table=sql.Identifier(table_name),
```
tomas-quix: Please add here a support for schema. Expose this in contrstructor and also connect that to environment variable in template.

Example:
```python
sql.Identifier(self.schema_name, self.table_name)
```

gwaramadze (approved):

PR #888, opened 2025-05-22
Clear corrupted RocksDB

gwaramadze:


quixstreams/state/rocksdb/partition.py
```diff
@@ -350,11 +350,33 @@ def _open_rocksdict(self) -> Rdict:
...
+        try:
+            rdict = create_rdict()
+        except Exception as exc:
+            if not str(exc).startswith("Corruption"):
+                raise
+            elif not self._changelog_producer:
+                raise RocksDBCorruptedError(
+                    "State is corrupted and cannot be recovered from changelog: "
+                    "`use_changelog_topics` is set to False."
+                )
```
ghost: Let's add the path to the exception message so it's clear which db is corrupted

ghost (approved):

PR #896, opened 2025-05-26
Topics refactoring

ghost:
- Resolve serdes when the topic is initialized
- Re-use the SerializationContext objects
- Add default serdes to the Topic class

### DIFF
--- a/quixstreams/models/topics/topic.py
+++ b/quixstreams/models/topics/topic.py
@@ -278,9 +278,9 @@ class Topic:
 
         timestamp_type, timestamp_ms = message.timestamp()
         message_context = MessageContext(
-            topic=message.topic(),
-            partition=message.partition(),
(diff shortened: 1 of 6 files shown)


quixstreams/models/topics/topic.py
```diff
@@ -244,42 +241,38 @@ def row_deserialize(
...
-            )
-        if self._value_deserializer is None:
-            raise DeserializerIsNotProvidedError(
-                f'Value deserializer is not provided for topic "{self.name}"'
-            )
-
         headers = message.headers()
+        topic = message.topic()
+        partition = message.partition()
+        offset = message.offset()
```
gwaramadze, line 247: We can also reuse these variables on lines 281-283.

gwaramadze (approved):

PR #903, opened 2025-05-28
Add Solar Farm Enrichment tutorial

ghost:


ghost:
Related docs PR - https://github.com/quixio/quix-docs/pull/491

docs/tutorials/solar-farm-enrichment/tutorial.md
```diff
@@ -0,0 +1,295 @@
...
+)
+```
+
+Now we join the telemetry data with the forecasts in "as-of" manner, so each telemetry event is matched with the latest forecast effective at the time
+when the telemetry was produced. 
+
+In particular: 
+
+- We use `how="inner"` which means that the results are emitted only when the match is found.  
+In can also be set to `how="left"` to emit records even if there is no matching forecast for the telemetry event.
```
SteveRosam: It can also be

docs/tutorials/solar-farm-enrichment/tutorial.md
```diff
@@ -0,0 +1,295 @@
...
+Because "join" is a stateful operation, and it requires access to multiple topic partitions within the same process, 
+the dataframes must meet certain requirements to be joined:
+
+1. The underlying topics of the dataframes must have the same number of partitions.  
+Quix Streams validates that when the `join_asof` is called.  
+In this tutorial, both topics have one partition.
+<br>
+<br>
+
+2. The messages keys in these topics must be distributed across partitions using the same algorithm Messageshe same partitioner.
```
SteveRosam: Messageshe ??

gwaramadze (approved):

PR #906, opened 2025-05-29
Feature: Collect duplicates in join-related store

gwaramadze:
* Rename retention_ms to grace_ms
* Pass grace_ms to the store constructor instead of later to the method
* Move global counter logic to a separate base class `RocksDBPartitionTransaction`
* Use the new class for both:
  * WindowedRocksDBPartitionTransaction
  * TimestampedPartitionTransaction
* Reset global counter if it overflows the unsigned 64bit integer
* Add param `collect_duplicates` to instruct `TimestampedStore` to collect all values for the same timestamp.
* Refactor `TimestampsCache` into a generic`Cache` that may accept any values because we need to store bytes.

### DIFF
--- a/quixstreams/dataframe/joins/join_asof.py
+++ b/quixstreams/dataframe/joins/join_asof.py
@@ -53,7 +53,7 @@ class JoinAsOf:
                 f"a callable to merge records manually."
             )
 
-        self._retention_ms = ensure_milliseconds(grace_ms)
+        self._grace_ms = ensure_milliseconds(grace_ms)
         self._store_name = store_name or "join"
 
     def join(
@@ -72,6 +72,7 @@ class JoinAsOf:
         right.processing_context.state_manager.register_timestamped_store(
             stream_id=right.stream_id,
             store_name=self._store_name,
+            grace_ms=self._grace_ms,
             changelog_config=changelog_config,
         )
 
@@ -101,12 +102,7 @@ class JoinAsOf:
                     store_name=self._store_name,
                 ),
             )
-            tx.set_for_timestamp(
-                timestamp=timestamp,
-                value=value,
-                prefix=key,
-                retention_ms=self._retention_ms,
-            )
+            tx.set_for_timestamp(timestamp=timestamp, value=value, prefix=key)
 
         right = right.update(right_func, metadata=True).filter(lambda value: False)
         left = left.apply(left_func, metadata=True).filter(
--- a/quixstreams/dataframe/joins/join_asof.py
+++ b/quixstreams/dataframe/joins/join_asof.py
@@ -73,7 +73,7 @@ class JoinAsOf:
             stream_id=right.stream_id,
             store_name=self._store_name,
             grace_ms=self._grace_ms,
-            collect_duplicates=False,
+            keep_duplicates=False,
             changelog_config=changelog_config,
         )
 
--- a/quixstreams/dataframe/joins/join_asof.py
+++ b/quixstreams/dataframe/joins/join_asof.py
@@ -73,6 +73,7 @@ class JoinAsOf:
             stream_id=right.stream_id,
             store_name=self._store_name,
             grace_ms=self._grace_ms,
+            collect_duplicates=False,
             changelog_config=changelog_config,
         )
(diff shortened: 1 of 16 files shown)


quixstreams/dataframe/joins/join_asof.py
```diff
@@ -72,6 +72,8 @@ def join(
         right.processing_context.state_manager.register_timestamped_store(
             stream_id=right.stream_id,
             store_name=self._store_name,
+            grace_ms=self._grace_ms,
+            collect_duplicates=False,
```
ghost: I suggest renaming it to `keep_duplicates`.
`Collect_duplicates` sounds like a method name to me :)

quixstreams/state/rocksdb/timestamped.py
```diff
@@ -88,14 +95,15 @@ def get_latest(self, timestamp: int, prefix: Any) -> Optional[Any]:
         :return: The deserialized value if found, otherwise None.
         """
         prefix = self._ensure_bytes(prefix)
-        min_eligible_timestamp = self._get_min_eligible_timestamp(prefix)
+        lower_bound_bytes = self._get_min_eligible_timestamp(prefix)
+        # +1 because upper bound is exclusive
+        upper_bount_bytes = int_to_bytes(timestamp + 1)
```
ghost:
```suggestion
        upper_bound_bytes = int_to_bytes(timestamp + 1)
```

quixstreams/state/rocksdb/timestamped.py
```diff
@@ -88,14 +95,15 @@ def get_latest(self, timestamp: int, prefix: Any) -> Optional[Any]:
...
 
-        if timestamp < min_eligible_timestamp:
+        if upper_bount_bytes <= lower_bound_bytes:
```
ghost:
```suggestion
        if upper_bound_bytes <= lower_bound_bytes:
```

quixstreams/state/rocksdb/timestamped.py
```diff
@@ -230,15 +227,15 @@ def _get_min_eligible_timestamp(self, prefix: bytes) -> int:
         :return: The minimum eligible timestamp (int).
```
ghost: Update the type here as well

quixstreams/state/rocksdb/transaction.py
```diff
@@ -0,0 +1,63 @@
...
+            dumps=dumps,
+            loads=loads,
+            changelog_producer=changelog_producer,
+        )
+        self._global_counter = CounterCache(
+            key=GLOBAL_COUNTER_KEY,
+            cf_name=GLOBAL_COUNTER_CF_NAME,
+        )
+
+    def _get_next_count(self) -> int:
```
ghost: Maybe rename it to `_increment_global_counter()`?

It does more than just "getting" under the hood

quixstreams/state/rocksdb/transaction.py
```diff
@@ -0,0 +1,63 @@
...
+
+        cache.counter += 1
+
+        try:
+            value = self._serialize_value(cache.counter)
+        except StateSerializationError:
+            cache.counter = 0
+            value = self._serialize_value(cache.counter)
+
+        self.set_bytes(value=value, key=cache.key, prefix=b"", cf_name=cache.cf_name)
```
ghost: Optimization idea: what if we read the counter once when the Transaction is initialized, bump the number on each update, and call `set_bytes` once on `flush`?

This way, we eliminate the serdes on each update.
And since it's a single global number, we don't need a special structure like `Cache` to maintain it (the cache always holds only one key).

quixstreams/state/rocksdb/timestamped.py
gwaramadze:
```suggestion
        :return: The minimum eligible timestamp as bytes.
```

quixstreams/state/rocksdb/transaction.py
gwaramadze: This is implemented, resolving.

ghost (approved):

PR #924, opened 2025-06-09
Feature: Interval joins

gwaramadze:
### Implement interval join

This PR introduces a new `join_interval()` method to `StreamingDataFrame` for performing stream-to-stream interval joins.

#### Interval Join

A new interval join feature has been added, allowing records from two streams to be joined based on a user-defined time interval around their timestamps.

**Key Features:**

*   **Symmetric Interval:** Configure the join with `backward_ms` and `forward_ms` to define a flexible time window for matching records between streams.
*   **Join Types:** Supports both `inner` and `left` joins.
*   **Flexible Merging:** Offers built-in strategies (`raise`, `keep-left`, `keep-right`) and allows custom functions for merging matched records.
*   **Stateful Efficiency:** Leverages an enhanced `TimestampedStore` that can now store and query multiple values per key and timestamp, which is crucial for handling multiple matches within an interval.

**Example Usage:**

```python
# Join records from `left_sdf` with records from `right_sdf` that
# fall within a window of 1s before to 2s after the left record's timestamp.
joined_sdf = left_sdf.join_interval(
    right_sdf,
    how="inner",
    backward_ms=1000,
    forward_ms=2000,
)
```

#### State Store Enhancements

To support this feature, `TimestampedStore` has been updated with a `get_interval()` method to retrieve all records within a given time range.


quixstreams/dataframe/joins/join_interval.py
```diff
@@ -0,0 +1,72 @@
...
+
+        def left_func(value, key, timestamp, headers):
+            tx(left).set_for_timestamp(timestamp=timestamp, value=value, prefix=key)
+
+            if right_values := tx(right).get_interval(
+                start=timestamp - backward_ms,
+                end=timestamp + 1,  # +1 because end is exclusive
+                prefix=key,
+            ):
+                return [merger(value, right_value) for right_value in right_values]
```
gwaramadze, line 72: `merger` may mutate the value, so I think we should pass copies of `value` here and in the `right_func` respectively. Still scratching my head on this, WDYT @daniil-quix ?
```suggestion
                return [merger(value.copy(), right_value) for right_value in right_values]
```

quixstreams/dataframe/joins/base.py
```diff
@@ -0,0 +1,116 @@
...
+__all__ = ("Join", "JoinHow", "JoinHow_choices", "OnOverlap", "OnOverlap_choices")
+
+JoinHow = Literal["inner", "left", "right", "outer"]
+JoinHow_choices = get_args(JoinHow)
+
+OnOverlap = Literal["keep-left", "keep-right", "raise"]
+OnOverlap_choices = get_args(OnOverlap)
+
+
+class Join:
```
ghost: This should be a subclass of `ABC`

docs/joins.md
```diff
@@ -199,3 +198,92 @@ if __name__ == '__main__':
...
+- custom callback - use a custom function to merge records
+
+> **Warning:** Custom merge functions must not mutate the input values as this will lead to
+> unexpected exceptions or incorrect data in the joined stream. Always return a new object instead.
+
+### Limitations
+
+- Joining dataframes belonging to the same topics (aka "self-join") is not supported
+- The `backward_ms` must not be greater than the `grace_ms` to avoid losing data
+- Interval join preserves headers only for the left dataframe
```
ghost: Should this be updated since we also support `right` and `outer`?

gwaramadze: In this case, it's even more complicated. Since we emit from both sides, it will sometimes come with the headers from the left and sometimes from the right.

docs/joins.md
```diff
@@ -199,3 +197,93 @@ if __name__ == '__main__':
...
+
+> **Warning:** Custom merge functions must not mutate the input values as this will lead to
+> unexpected exceptions or incorrect data in the joined stream. Always return a new object instead.
+
+### Limitations
+
+- Joining dataframes belonging to the same topics (aka "self-join") is not supported
+- The `backward_ms` must not be greater than the `grace_ms` to avoid losing data
+- Interval join does not preserve any headers. If you need headers from any side, consider adding them to the value.
+- Performance of the join depends on the density of the data. If streams on both sides move very fast (a lot of messages) then the performance may drop significantly.
```
ghost:
```suggestion
- The performance of the interval join depends on the density of the data.   
If both streams have too many matching messages falling within the interval, the performance may drop significantly due to the large number of produced outputs.
```

docs/joins.md
```diff
@@ -199,3 +197,93 @@ if __name__ == '__main__':
...
+- `keep-right` - prefer keys from the right record in case of overlap
+- custom callback - use a custom function to merge records
+
+> **Warning:** Custom merge functions must not mutate the input values as this will lead to
+> unexpected exceptions or incorrect data in the joined stream. Always return a new object instead.
+
+### Limitations
+
+- Joining dataframes belonging to the same topics (aka "self-join") is not supported
+- The `backward_ms` must not be greater than the `grace_ms` to avoid losing data
```
ghost:
```suggestion
- Joining dataframes belonging to the same topic (aka "self-join") is not supported.
- The `backward_ms` must not be greater than the `grace_ms` to avoid losing data.
```

quixstreams/dataframe/dataframe.py
```diff
@@ -1728,10 +1730,115 @@ def join_asof(
...
         """
-        return JoinAsOf(
+        return AsOfJoin(
             how=how, on_merge=on_merge, grace_ms=grace_ms, store_name=name
         ).join(self, right)
 
+    def join_interval(
+        self,
+        right: "StreamingDataFrame",
+        how: Literal["inner", "left"] = "inner",
```
ghost: Shoud it be `JoinHow` instead?

quixstreams/dataframe/dataframe.py
```diff
@@ -1661,7 +1661,7 @@ def concat(self, other: "StreamingDataFrame") -> "StreamingDataFrame":
     def join_asof(
         self,
         right: "StreamingDataFrame",
-        how: JoinAsOfHow = "inner",
+        how: JoinHow = "inner",
```
ghost: Here it should be `Literal["inner", "left"]` or an alias, because as-of join supports only these strategies

quixstreams/dataframe/joins/join_asof.py
```diff
@@ -1,112 +1,49 @@
...
-OnOverlap = Literal["keep-left", "keep-right", "raise"]
-OnOverlap_choices = get_args(OnOverlap)
 
-
-class JoinAsOf:
+class AsOfJoin(Join):
     def __init__(
         self,
-        how: JoinAsOfHow,
+        how: JoinHow,
```
ghost: Also here

quixstreams/dataframe/dataframe.py
gwaramadze: I have added a check for this in `AsOfJoin.__init__`. I will add a note in the docs for it.

quixstreams/dataframe/dataframe.py
gwaramadze: yes! correcting

ghost (approved):

PR #925, opened 2025-06-09
StreamingDataFrame: retain a custom stream_id across operations

ghost:
# Problem
In https://github.com/quixio/quix-streams/pull/836, a custom `stream_id` parameter was introduced to the `StreamingDataFrame` class and its `__dataframe_clone__` method; however, calling `__dataframe_clone__` again reset the stream_id back to the default value obtained from the underlying topics.

The stream_id is used as part of the State stores' names, and it wasn't propagated correctly, leading to incorrect store names in some rare cases.
This PR corrects that, but the state stores created after `.filter()` or `.apply()` operations on the grouped DataFrame won't be accessible anymore.

:
```python
topic = app.topic('<some-topic-with-one-partition>')

sdf = StreamingDataFrame(topic)  # stream_id = "<some-topic-with-one-partition>" (correct)

sdf_grouped = sdf.group_by("column")  # stream_id = "column--groupby--<some-topic-with-one-partition>" (correct)

# The store gets registered for the stream_id "column--groupby--<some-topic-with-one-partition>" (correct)
sdf_windowed = sdf_grouped.tumbling_window(...).sum().final()

# But here the state was registered under the stream_id "<some-topic-with-one-partition>" (incorrect).
# The correct stream_id is  "column--groupby--<some-topic-with-one-partition>"
sdf_windowed.apply(..., stateful=True)
```

# Solution
- Pass `stream_id` to the cloned dataframes
- Update `StreamingDataFrame.concat()` to generate a new stream_id when concatenating branches with different stream_ids (possible when concatenating the group_by-ed dataframe with a one-partition topic

### DIFF
--- a/quixstreams/utils/stream_id.py
+++ b/quixstreams/utils/stream_id.py
@@ -1,0 +1,3 @@
+def stream_id_from_strings(*strings: str) -> str:
+    parts = sorted(set(strings))
+    return "--".join(parts)
--- a/tests/test_quixstreams/test_dataframe/test_dataframe.py
+++ b/tests/test_quixstreams/test_dataframe/test_dataframe.py
@@ -2728,6 +2728,7 @@ class TestStreamingDataFrameJoinAsOf:
     def topic_manager(self, topic_manager_factory):
         return topic_manager_factory()
 
+    # TODO: Check if we already have a fixture for that to avoid the pollution
     @pytest.fixture
     def create_topic(self, topic_manager):
         def _create_topic(num_partitions=1):
--- a/tests/test_quixstreams/test_dataframe/test_dataframe.py
+++ b/tests/test_quixstreams/test_dataframe/test_dataframe.py
@@ -550,7 +550,7 @@ class TestStreamingDataFrame:
         assert sdf.stream_id == "some-id"
 
         # Ensure that the previous stream_id is passed
-        assert sdf.__dataframe_clone__().stream_id == sdf.stream_id
+        assert sdf.__dataframe_clone__().stream_id == sdf.stream_id == "some-id"
 
         # Ensure that new stream_id is passed
         assert sdf.__dataframe_clone__(stream_id="new-id").stream_id == "new-id"
--- a/tests/test_quixstreams/test_dataframe/test_dataframe.py
+++ b/tests/test_quixstreams/test_dataframe/test_dataframe.py
@@ -556,7 +556,7 @@ class TestStreamingDataFrame:
         assert sdf.__dataframe_clone__(stream_id="new-id").stream_id == "new-id"
 
         # If the topic is passed to the __dataframe_clone__, the existing
-        # must be overridden
+        # stream_id must be overridden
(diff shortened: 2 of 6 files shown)


quixstreams/utils/stream_id.py
```diff
@@ -0,0 +1,3 @@
+def stream_id_from_strings(*strings: str) -> str:
+    parts = sorted(set(strings))
```
gwaramadze, line 2: I'm not sure we need the `set` that badly. And if we ever want to do self-joins, we will have to accept duplicate strings. However, we can modify it later if necessary.

tests/test_quixstreams/test_dataframe/test_dataframe.py
```diff
@@ -541,6 +543,30 @@ def test_drop_missing_columns_errors_ignore(self, dataframe_factory):
...
         ] == (expected, key, timestamp, headers)
 
+    def test__dataframe_clone__stream_id(
+        self, dataframe_factory, topic_manager_topic_factory
+    ):
+        sdf = dataframe_factory(stream_id="some-id")
+        assert sdf.stream_id == "some-id"
+
+        # Ensure that the previous stream_id is passed
+        assert sdf.__dataframe_clone__().stream_id == sdf.stream_id
```
gwaramadze: We can make sure that nothing has mutated the `sdf.stream_id`.
```suggestion
        assert sdf.__dataframe_clone__().stream_id == sdf.stream_id == "some-id"
```

tests/test_quixstreams/test_dataframe/test_dataframe.py
```diff
@@ -541,6 +543,30 @@ def test_drop_missing_columns_errors_ignore(self, dataframe_factory):
...
+
+        # Ensure that new stream_id is passed
+        assert sdf.__dataframe_clone__(stream_id="new-id").stream_id == "new-id"
+
+        # If the topic is passed to the __dataframe_clone__, the existing
+        # must be overridden
```
gwaramadze: I think you missed this
```suggestion
        # stream_id must be overridden
```

tests/test_quixstreams/test_dataframe/test_dataframe.py
```diff
@@ -2650,7 +2728,6 @@ class TestStreamingDataFrameJoinAsOf:
     def topic_manager(self, topic_manager_factory):
         return topic_manager_factory()
 
-    # TODO: Check if we already have a fixture for that to avoid the pollution
```
gwaramadze: Minor: I did the check and refactored this place on my branch. If you bring back the line, it will save me from the conflict later 😉

quixstreams/utils/stream_id.py
ghost: I'd keep the set to make sure the stream_ids are not duplicated for now.
When we add self-joins, we may need to relax it.

gwaramadze (approved):

PR #932, opened 2025-06-13
Update Application.run() behavior

ghost:
## Main changes

1. `Application.run()` with the `count` param now counts the **outputs** instead of **messages**.  
It makes the behavior easier to reason about for the users, and also simplifies the inner logic.   
For example, we no longer need to handle messages from the repartition topics separately.

2. `Application.run()` with `count` or `timeout` parameters can now accumulate and return the processed outputs for further debugging.  
This approach is less manual than using a `ListSink`.

3. `StreamingDataFrame.group_by()` now filters the data after sending it to the repartition topic to avoid double-counting


docs/debugging.md
```diff
@@ -91,72 +91,101 @@ sdf.print_table(title="Final Output")
...
 when the applicable condition is met:
 
-- `timeout`: maximum time to wait for a new message (default `0.0` == infinite)
-- `count`: number of messages to process from main SDF input topics (default `0` == infinite)
+- `timeout`: maximum time to wait for a new message to arrive (default `0.0` == infinite)
+- `count`: a number of outputs to process across all dataframes and input topics (default `0` == infinite)
+- `collect_mode`: how to collect the processed outputs to return them as a result of the `Application.run()` call.  
+  This setting is effective only when either `timeout` or `count` are passed.  
+  Possible values:
+  - `"values"` (default) - the output values are collected into a list of dictionaries.
```
gwaramadze: How will it behave if the values are not dicts? Maybe this wording will be more accurate.
```suggestion
  - `"values"` (default) - the output values are collected into a list.
```

docs/debugging.md
```diff
@@ -91,72 +91,101 @@ sdf.print_table(title="Final Output")
...
+# Process one output and without collecting (stops if no messages for 10s)
+result_empty = app.run(count=1, timeout=10, collect_mode="off")
+# >>> result_empty = []
+
+
+```
+
+#### Count Behavior
+
+There are a few of things to be aware of with `count`:
```
gwaramadze:
```suggestion
There are a few things to be aware of with `count`:
```

docs/debugging.md
```diff
@@ -91,72 +91,101 @@ sdf.print_table(title="Final Output")
...
+
+- It counts _outputs_ from processed by **all** `StreamingDataFrames`.
```
gwaramadze:
```suggestion
- It counts _outputs_ processed by **all** `StreamingDataFrames`.   
```

quixstreams/runtracker.py
```diff
@@ -1,102 +1,163 @@
...
+                "_partition": partition,
+                "_offset": offset,
+                **value,
+            }
+        )
+        self._count += 1
+
+    def add_value(self, value: Any):
+        if not isinstance(value, Mapping):
+            value = {"_value": value}
```
gwaramadze, line 51: do we have to do this?
```suggestion
```

quixstreams/runtracker.py
```diff
@@ -1,102 +1,163 @@
...
+        self._items.append(value)
+        self._count += 1
+
+    def increment_count(self):
+        self._count += 1
+
+    @property
+    def items(self) -> list[dict]:
+        # Make a copy to retain a list after the .clear() is called
```
gwaramadze: It's not clear what `clear` method is this referring to and what is its relationship with the `RunCollector`

docs/debugging.md
ghost: If the record's value is not a dict, it will be wrapped into a dict by the `RunCollector`.

quixstreams/runtracker.py
ghost: It's a leftover, I forgot to remove it.

quixstreams/runtracker.py
ghost: This code wraps a value into a dict if it's not a dict, so yes

gwaramadze (dismissed):

gwaramadze (approved):

PR #936, opened 2025-06-17
InfluxDB1Sink: new

tim-quix:
Added Influxdb v1 sink, which closely matches v3 in functionality.

Still need to add docs.

Tested against a local influxdb1 docker image, but still need to test against a productionized instance.


tomas-quix:
It is working against Influx Enteprise V1. @daniil-quix can we merge it and release it?

ghost (approved):

PR #954, opened 2025-06-26
Join Lookup: PostgreSQL

tim-quix:
An equivalent approach to the Sqlite version, using PostgreSQL.

May want to discuss how best to handle the dependencies and/or mocking out calls (the sqlite one used sqlite directly).

I came up with a solution that seemed to work reasonably enough.

### DIFF
--- a/quixstreams/dataframe/joins/lookups/postgresql.py
+++ b/quixstreams/dataframe/joins/lookups/postgresql.py
@@ -1,0 +1,446 @@
+import abc
+import copy
+import dataclasses
+import logging
+import re
+import time
+from collections import OrderedDict
+from typing import Any, Literal, Mapping, Tuple, Union
+
+import psycopg2
+from psycopg2.extensions import cursor as pg_cursor
+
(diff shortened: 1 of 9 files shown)


quixstreams/dataframe/joins/lookups/postgresql.py
```diff
@@ -0,0 +1,446 @@
...
+
+        :param field: The field definition (SQLiteLookupField or SQLiteLookupQueryField).
+        :param on: The key to use in the WHERE clause for lookup.
+        :param value: The message value, used to substitute parameters in the query.
+
+        :returns: The extracted data, either a single row or a list of rows.
+        """
+        query, parameters = field.build_query(on, value)
+        if self._debug:
+            print(f"QUERY: {query}")
```
daniil-quix, line 347: Do we need this print here?

ghost:
```suggestion
            logger.debug(f"QUERY: {query}")
```

ghost (approved):

PR #955, opened 2025-07-01
New Connector: MySql CDC Source

tim-quix:
Adds `MySqlCdcSource`, a community `StatefulSource` that streams row-level changes from a MySQL table via binary log replication, with an optional initial snapshot of the table's current contents.

Extracted from https://github.com/quixio/quix-samples/pull/620 and reworked to follow the usual Source pattern.

```python
from quixstreams.sources.community.mysql_cdc import MySqlCdcSource

source = MySqlCdcSource(
    host="localhost", port=3306,
    user="cdc_user", password="cdc_password",
    database="test_database", table="test_table",
    initial_snapshot=True,
)

app = Application(broker_address="localhost:9092", consumer_group="mysql-cdc")
sdf = app.dataframe(source=source).print(metadata=True)

if __name__ == "__main__":
    app.run()
```

Install with `pip install quixstreams[mysql]`.

## Delivery guarantee

**At-least-once, not exactly-once.** The binlog position is committed only after the messages it covers have been flushed to Kafka, so no change is lost — but a crash between the data flush and the position commit replays the last batch. Consumers must treat change events as idempotent and deduplicate on the primary-key columns in `columnvalues`/`oldkeys` together with `kind`. The docs page lists the three windows where duplicates occur.

Every message is keyed `"<database>.<table>"`, so all changes to one table stay on one partition in binlog order. The key is not the row identity.

## Review fixes in this branch

Addresses the blocking review on this PR, plus three defects found while fixing it:

- **Data loss on restart** — the position was persisted before the events it covered were produced, and the shutdown drain was gated behind a 500 ms flush interval, so a prompt SIGTERM dropped the buffer. Now: produce → flush → commit position, with an unconditional final drain.
- **Restart recovery ignored the state store** — resume was driven by a JSON file at a relative path that does not survive an ephemeral filesystem, so a restart silently jumped to the current master position and lost the whole downtime window. The file is gone; `self.state` is the only source of truth.
- **`server_id` hardcoded to 1** — collided with the server's own id and with any second instance. Now an optional argument defaulting to a deterministic value derived from `name`/`database`/`table`.
- **Snapshot was not production-safe** — `LIMIT/OFFSET` without `ORDER BY` duplicated and skipped rows under concurrent DML, the whole table was accumulated before a single flush, and the stream resolved its start position only after the snapshot had already run. Now: keyset pagination on the primary key (composite keys supported, no primary key fails at `setup()`), a generator flushed per page with progress checkpointed to state, and the start position captured before the first row is read.
- **Repo conventions** — `ImportError` guard pointing at `quixstreams[mysql]`, the full `StatefulSource` contract, class and `:param:` docs, and registration in `docs/build/build.py`.

Also fixed:

- `mysql-replication>=1.1` was unsatisfiable (PyPI tops out at 1.0.17), so `pip install quixstreams[mysql]` could never resolve.
- Fatal MySQL auth failures (1044/1045/1227) were retried behind the reconnect loop instead of propagating. 1236 is deliberately still retryable — MySQL reuses it for a transient `server_id` collision as well as a purged binlog.
- Change events shipped `columnnames` as `UNKNOWN_COL0..n` on any server whose binlog carries no column metadata — the MySQL 8.x default (`binlog_row_metadata=MINIMAL`) and always on 5.7 — while snapshot events carried real names, putting two incompatible schemas in one topic.

## Verification

Unit tests cover the pure logic (36). Behaviour was verified against live MySQL in Docker — 18 scenarios across MySQL 8.0 and 8.4, including a primary/replica pair — each with the pre-fix code run as a control so the failures are reproduced before the fixes are proven:

| Scenario | Fixed | Pre-fix |
|---|---|---|
| Snapshot under concurrent INSERT/UPDATE/DELETE | 31 522/31 522 rows, gaps 0 | gaps 34, extras 43, mismatches 30 |
| Per-page flush on a 50 000-row snapshot | 100 bursts / 100 pages | 1 burst |
| SIGKILL mid-stream | 0 lost | gaps 300 |
| SIGTERM drain | 1000/1000, exit 0 | produced 0/1000 |
| State directory wiped between runs | 2000/2000, gaps 0 | gaps 1000 |
| Snapshot from a lagging replica | gaps 0, 83 739 B lag window | gaps 300 |
| Column names on a `MINIMAL`-metadata server | 80/80 correct | 80/80 `UNKNOWN_COL` |

Plus: interrupted snapshot resuming from its checkpoint, two instances against one server, no-primary-key and misconfiguration failures exiting with actionable messages, and purged-binlog refusal after bounded retries.

## Known limits

- `RENAME COLUMN` under a running source keeps serving the old name — `pymysqlreplication` caches the column list per `schema.table` and never invalidates it. `ADD`/`DROP` degrade visibly to `UNKNOWN_COL` instead.
- Column-level `SELECT` grants are discarded as a partial column list; grant `SELECT` on the whole table.
- MySQL 5.7 is documented as supported but was not exercised end to end; the rig runs 8.0 and 8.4.
- No tests ship for this connector beyond the unit file, consistent with the other community sources. The Docker rig is not part of the repo.


quixstreams/sources/community/mysql_cdc/mysql_cdc.py
```diff
@@ -0,0 +1,251 @@
...
+            **{
+                "host": host,
+                "port": port,
+                "user": user,
+                "password": password,
+                "database": database,
+                "table": table,
+                "snapshot_host": snapshot_host or host,
+                "state_dir": state_dir,
+            }
```
gwaramadze, line 46: I think it doesn't need to be a dict
```suggestion
            host=host,
            port=port,
            user=user,
            password=password,
            database=database,
            table=table,
            snapshot_host=snapshot_host or host,
            state_dir=state_dir,
```

quixstreams/sources/community/mysql_cdc/mysql_cdc.py
```diff
@@ -0,0 +1,251 @@
...
+                logger.info(
+                    f"Initial snapshot already completed at {snapshot_info.get('timestamp', 'unknown time')} - skipping"
+                )
+                return
+        else:
+            logger.info(
+                "Initial snapshot is enabled and not yet completed - performing snapshot..."
+            )
+
+        if not self.is_snapshot_completed() or self.force_snapshot:
```
gwaramadze, line 134: There is a bit of redundant logic here because `is_snapshot_completed` also checks `self.force_snapshot`. If you ask me, the method should not check `force_snapshot`, it should only check actual completeness, not if the snapshot should be done forcefully.

LudvikQuix:
## Review: not mergeable as-is

I merged `main` into this branch locally to review it in context. The two conflicts (`pyproject.toml` extras and `conda/post-link.sh`) are additive dependency lists and resolve cleanly by union — `tdengine`, `mqtt` and `mysql` extras all coexist, and `pre-commit run --all-files` passes on the touched files. So the merge itself is mechanically clean; the blockers below are in the code.

### Blocking points

**1. Data loss is structural, not incidental.**
`MySqlHelper.get_changes()` persists the binlog position (`mysql_helper.py:238`) *before* the events it covers have been produced or flushed to Kafka. In parallel, `stop()` (`mysql_cdc.py:240`) routes the remaining buffer through `process_buffered_messages()`, which is gated on the 500 ms flush interval — so on a SIGTERM less than 500 ms after the last flush it produces nothing, logs "Processing N remaining buffered messages", and drops them. Since the position was already advanced past those rows, they are unrecoverable. Every restart can lose up to a flush interval of changes; an unclean kill loses more. The PR description and docs claim *"Exactly-Once Processing: No data loss during application restarts or failures"* — that claim is not met.

**2. Restart recovery does not use the state store.**
`StatefulSource` is the headline feature here, but `self.state` is effectively write-only: `save_binlog_position()` writes to it (`mysql_cdc.py:105`) and `get_binlog_position()` reads it only to emit a log line. The position that actually drives resume comes from `MySqlHelper.load_binlog_position()` (`mysql_helper.py:140`) — a JSON file at the *relative* path `state/binlog_position_<db>_<table>.json`, resolved against the source subprocess CWD and independent of the Application's configured `state_dir`. On any ephemeral-filesystem deployment (the normal Quix Cloud case) that file is gone on restart, `load_binlog_position()` returns `(None, None)`, and the stream silently restarts from the *current* master position — the entire downtime window is lost, even though the changelog-backed store holds the correct position. This needs a design fix (drive resume from `self.state`), not a patch.

**3. `server_id` is hardcoded to 1 and not configurable.**
`create_binlog_stream(self, server_id=1)` (`mysql_helper.py:132`) is always called with no argument, and `MySqlCdcSource.__init__` exposes no `server_id` parameter. The PR's own docs (`docs/connectors/sources/mysql-cdc-source.md`, Prerequisites) instruct users to set `server-id = 1` on the MySQL server itself. MySQL rejects or repeatedly disconnects a replication client whose `server_id` duplicates another registered replica, so the documented setup — and any second `MySqlCdcSource` instance — hits *"A slave with the same server_id is already connected"* in a reconnect loop, with no knob to fix it.

**4. The snapshot path is not production-safe.**
Three separate issues:
- Pagination uses `LIMIT {n} OFFSET {m}` with **no `ORDER BY`** (`mysql_helper.py:264`). MySQL gives no stable ordering guarantee across separate executions, so batches can overlap and skip — rows silently duplicated as `snapshot_insert` and others omitted, with `processed_rows` still reaching the expected total so nothing detects it. Concurrent DELETEs during the snapshot shift OFFSET and skip rows too.
- `perform_initial_snapshot()` accumulates every row into one list and returns it (`mysql_helper.py:292`/`:312`), and the caller produces all of them before a single `flush()` (`mysql_cdc.py:141-149`). That defeats the point of the batching and OOMs on a large table, contradicting the docs' *"Memory Efficient: Processes data in chunks to avoid memory issues"*. A generator yielding per batch, flushed per batch, is the fix.
- The snapshot reads from `snapshot_host` (documented as a read replica) while the binlog stream is positioned on the **primary**. Rows changed on the primary within the replica's lag window are snapshotted with stale values, and their `UpdateRowsEvent` sits *before* the stream's start position, so it is never streamed — those rows stay permanently stale. "Read Replica Support" needs the stream to start from the replica's executed position, or the snapshot taken under a consistent-position lock.

**5. Repo conventions for community connectors are not met.**
- No `ImportError` guard on the optional deps: `mysql_helper.py:7-13` imports `pymysql`/`pymysqlreplication` bare, whereas every peer (e.g. `quixstreams/sources/community/mqtt.py:12-17`) wraps it to raise *`Package "..." is missing: run pip install quixstreams[...] to fix it`*. A user without the extra gets a raw `ModuleNotFoundError` and no pointer to `quixstreams[mysql]`.
- The module is not registered in `docs/build/build.py` (the `sources.md` list, ~lines 150-170), so `MySqlCdcSource` will not appear in the generated `docs/api-reference/sources.md` at all. There is also no class docstring and no `:param:` docs on `__init__`, unlike `KinesisSource`/`MQTTSource`.
- `__init__` calls `super().__init__(name=name)` only, dropping the standard `on_client_connect_success` / `on_client_connect_failure` / `shutdown_timeout` contract every other source exposes; `setup()`'s own try/except + `logger.error` + re-raise then duplicates what `BaseSource._init_client` already does, so connection failures are logged twice.
- The docs page describes a template app — `MYSQL_HOST`/`INITIAL_SNAPSHOT`/`output` env vars, `python main.py`, `pip install -r requirements.txt` — none of which apply to this constructor-configured SDK source.

### Not blockers

Two things I checked deliberately: no community source in this repo ships tests, so their absence here is consistent with existing practice; and `quixstreams.sources.community.*` carries `ignore_errors = true` in the mypy overrides, so the untyped new module won't redden CI.

LudvikQuix:
## Review guide — what changed since the original submission

The original branch was extracted from quix-samples#620 and reviewed here on 2026-09-14. That review raised five blocking points; fixing them surfaced ten more, most of which are invisible without a live MySQL server. Fifteen defects total, all reproduced red before being fixed.

`main` has also been merged in, so the two dependency conflicts (`pyproject.toml` extras, `conda/post-link.sh`) are resolved by union.

### The five original points

| | |
|---|---|
| Data loss on restart | position was persisted before the events it covered were produced, and the shutdown drain sat behind a 500 ms interval gate. Now produce → flush → commit position, with an unconditional final drain. |
| Restart recovery ignored the state store | resume came from a JSON file at a relative path that does not survive an ephemeral filesystem. The file is gone; `self.state` is the only source of truth. |
| `server_id` hardcoded to `1` | now derived, and see below — deriving it was not enough. |
| Snapshot not production-safe | `LIMIT/OFFSET` without `ORDER BY` duplicated and skipped rows under concurrent DML. Now keyset pagination on the primary key, a generator flushed per page with progress checkpointed, and the start position captured before the first row is read. |
| Repo conventions | `ImportError` guard, full `StatefulSource` contract, `:param:` docs, `docs/build/build.py` registration. |

### Found while fixing them

- `mysql-replication>=1.1` was **unsatisfiable** — PyPI tops out at 1.0.17, so `pip install quixstreams[mysql]` could never resolve.
- Fatal MySQL auth failures (1044/1045/1227) were retried behind the reconnect loop. 1236 stays retryable — MySQL reuses it for a transient `server_id` collision as well as a purged binlog.
- **Row metadata.** MySQL writes column names, charsets, enum/set dictionaries and signedness into the binlog only when `binlog_row_metadata=FULL`, which is not the 8.x default. Without them: column names became `UNKNOWN_COL0..n`, ENUM and SET arrived as `None`, `INT UNSIGNED 4294967295` decoded as `-1`, and a binary column holding a non-UTF-8 byte raised `UnicodeDecodeError` and killed the process permanently, replaying the same row every restart. The snapshot path read the same columns correctly, so one column produced two answers.
- **Value contract.** JSON columns emitted a Python repr with bytes keys that `json.loads` could not parse; SET emitted a Python set with process-varying order. Neither is fixed by `FULL`. A new `values` module gives both paths one contract.
- **Replica identity.** `server_id` was derived from configuration alone, so every replica of one deployment announced the same id and they evicted each other through 1236 forever — and the reconnect counter reset on each success, so the attempt cap was never reached.
- `force_snapshot` never cleared the stored position, so the documented recovery re-published the whole table on every restart and still died on the same purged position.
- **No TLS surface at all.** Now on by default with a real `SSLContext`.
- `read_changes` had no stop check or time bound, so a large catch-up ignored `stop()` and lost its batch to SIGKILL.

### Behaviour changes worth a reviewer's attention

- **Delivery is documented as at-least-once, not exactly-once.** The position is committed only after the data flush returns, so a crash replays a batch rather than skipping one. Consumers must deduplicate.
- **`binlog_row_metadata=FULL` is required by default.** Servers that cannot supply it — including every MySQL 5.7 — run with `allow_minimal_row_metadata=True`, which logs exactly which column types corrupt.
- **The on-topic message format changed** for JSON, SET, BIT and FLOAT so the snapshot and binlog paths agree.

### Verification

58 unit tests. Behaviour was verified against live MySQL 8.0 and 8.4 in Docker, including a primary/replica pair and a `MINIMAL`-metadata server, with the pre-fix code run as a control on each scenario. Representative before/after:

| | pre-fix | fixed |
|---|---|---|
| snapshot under concurrent DML | gaps 34, extras 43, mismatches 30 | 31 522/31 522, gaps 0 |
| SIGKILL mid-stream | gaps 300 | 0 lost |
| state directory wiped | gaps 1000 | 2000/2000 |
| two replicas, one config | 2.0× duplicate events, neither exited | distinct ids, both stream |
| value fidelity, 13 column types | 20 cross-path mismatches, 4 unparseable JSON | 0 |
| connection churn | 13.8/s | 2.0/s |

### Known limits, documented in the connector docs

- With `allow_minimal_row_metadata` set, a binary column whose bytes happen to be valid UTF-8 is delivered as text where the snapshot delivers base64. The information needed to distinguish them is the metadata the server is not writing.
- `RENAME COLUMN` under a running source keeps serving the old name; `pymysqlreplication` caches the column list per table and never invalidates it. `ADD`/`DROP` degrade visibly instead.
- `FLOAT` is rounded to six significant figures on the binlog side — neither path can emit the other's native precision.
- The Docker rig used for the above is not part of the repo.

LudvikQuix:
Closing as superseded by #1152.

The MySQL CDC source ships as the streaming-only `MySqlCdcLiteSource` in #1152 rather than this branch. That is a scope decision, not a verdict on this code: the initial snapshot, the server-configuration repair, and the refusals for configurations that corrupt silently were all built and verified here, and the design that replaced them assumes a controlled environment where the operator owns server configuration and data quality.

What carried over is the part that is not defensive — produce → flush → commit ordering, the binlog position in the state store, prompt stop with a final drain, the JSON encoder, and the statement-boundary commit rules found in review of #1152. The six review rounds on this PR, including the two external ones, are what found most of those; that history stays here.

The branch `source/mysql-cdc` is left in place at fafbe869 for reference.

PR #966, opened 2025-07-08
PostgreSQLSink - Add primary keys + upserting capabilites

tim-quix:
Added ability to add/use primary keys (single or compound) and "upsert" based on those primary keys.
## Overview

If the primary keys do not exist, it will create/set them (but if you exclude existing primary keys, it will throw an error).

New args for `PostgreSQLSink` include:

`primary_keys`: An optional single (or list of) primary key(s).
    Can also provide a callable that accepts the message value as input, and returns a similar result (or nothing).
    Often paired with `upsert_on_primary_keys=True`.
    It must include all currently defined primary_keys on a given table.

`upsert_on_primary_keys`: Upsert based on the given `primary_keys`.
    If False, every message is treated as an independent entry, and any primary key collisions will consequently raise an exception.
    

## Usage

Example might look like:

```python
sink = PostgreSQLSink(
    host="localhost",
    port=5432,
    dbname="local",
    user="local",
    password="local",
    table_name="my_test_table",
    primary_keys=["host", "resource"],
    upsert_on_primary_key=True,
)
```

## TODO:

- More Tests
- Any remaining adjustments to go from `Community` to `Core`
- Updating documentation
- Potentially revisit logic to see if there's room to streamline anything once tests are fully in place (currently only partially done).

### DIFF
--- a/docs/connectors/sinks/postgresql-sink.md
+++ b/docs/connectors/sinks/postgresql-sink.md
@@ -91,7 +91,14 @@ PostgreSQLSink accepts the following configuration parameters:
 - `schema_name`: The schema name. Schemas are a way of organizing tables and 
   not related to the table data, referenced as `<schema_name>.<table_name>`.  
   PostrgeSQL uses "public" by default under the hood.
-- `schema_auto_update`: If True, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
+- `schema_auto_update`: If `True`, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
+- `primary_key`: A single or multiple (composite) primary key columns.
+  Can also provide a callable that accepts the message value as input, and returns a similar result (or nothing).
+  Often paired with `upsert_on_primary_key=True`.
+  It must include all currently defined primary_key columns on a given table.
+  Once set, no others can be added; to change, the primary key must be removed manually.
+- `upsert_on_primary_key`: if `True`, upsert based on the given `primary_key` argument.
+  If False, every message is treated as an independent entry, and any primary key collisions will consequently raise an exception.
 
 
 ## Testing Locally
--- a/docs/connectors/sinks/postgresql-sink.md
+++ b/docs/connectors/sinks/postgresql-sink.md
@@ -92,7 +92,7 @@ PostgreSQLSink accepts the following configuration parameters:
   not related to the table data, referenced as `<schema_name>.<table_name>`.  
   PostrgeSQL uses "public" by default under the hood.
 - `schema_auto_update`: If `True`, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
-- `primary_key`: A single or multiple (composite) primary key columns.
+- `primary_keys`: A single or multiple (composite) primary key columns.
   Can also provide a callable that accepts the message value as input, and returns a similar result (or nothing).
   Often paired with `upsert_on_primary_key=True`.
   It must include all currently defined primary_key columns on a given table.
--- a/docs/connectors/sinks/postgresql-sink.md
+++ b/docs/connectors/sinks/postgresql-sink.md
@@ -92,8 +92,8 @@ PostgreSQLSink accepts the following configuration parameters:
   not related to the table data, referenced as `<schema_name>.<table_name>`.  
   PostrgeSQL uses "public" by default under the hood.
 - `schema_auto_update`: If `True`, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
-- `primary_keys`: A single or multiple (composite) primary key columns.
-  Can also provide a callable that accepts the message value as input, and returns a similar result (or nothing).
+- `primary_key_columns`: A single or multiple (composite) primary key columns.
+  Can also provide a callable that accepts the message value as input, and returns a string or list of strings.
   Often paired with `upsert_on_primary_key=True`.
   It must include all currently defined primary_key columns on a given table.
   Once set, no others can be added; to change, the primary key must be removed manually.
--- a/quixstreams/sinks/community/postgresql.py
+++ b/quixstreams/sinks/community/postgresql.py
@@ -108,7 +108,7 @@ class PostgreSQLSink(BatchingSink):
             returns a string or list of strings.
             Often paired with `upsert_on_primary_key=True`.
             It must include all currently defined primary key columns on a given table.
-        :param upsert_on_primary_key: Upsert based on the given `primary_keys`.
+        :param upsert_on_primary_key: Upsert based on the given `primary_key_columns`.
             If False, every message is treated as an independent entry, and any
             primary key collisions will consequently raise an exception.
         :param on_client_connect_success: An optional callback made after successful
--- a/quixstreams/sinks/community/postgresql.py
+++ b/quixstreams/sinks/community/postgresql.py
@@ -29,10 +29,10 @@ __all__ = ("PostgreSQLSink", "PostgreSQLSinkException")
 logger = logging.getLogger(__name__)
 
 # A column name for the records keys
-_KEY_COLUMN_NAME = "__key"
+KEY_COLUMN_NAME = "__key"
 
 # A column name for the records timestamps
-_TIMESTAMP_COLUMN_NAME = "timestamp"
+TIMESTAMP_COLUMN_NAME = "timestamp"
 
 # A mapping of Python types to PostgreSQL column types for schema updates
 _POSTGRES_TYPES_MAP: dict[type, str] = {
@@ -48,6 +48,7 @@ _POSTGRES_TYPES_MAP: dict[type, str] = {
     bool: "BOOLEAN",
 }
 
+TableName = Union[Callable[[SinkItem], str], str]
 PrimaryKeySetter = Callable[[SinkItem], Optional[Union[str, list[str]]]]
 PrimaryKeys = Union[str, list[str], PrimaryKeySetter]
 
@@ -58,6 +59,9 @@ class PostgreSQLSinkException(QuixException): ...
 class PostgresSQLSinkMissingExistingPK(QuixException): ...
 
 
+class PostgresSQLSinkInvalidPK(QuixException): ...
+
+
 class PostgreSQLSink(BatchingSink):
     def __init__(
         self,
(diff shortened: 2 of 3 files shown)


tim-quix:
There is still a bit I'd like to do here to bring it up to "core" quality. First I wanted to get some tests in place.

Since Tomas wants to get this new functionality available ASAP, I focused on adding some tests to around the primary key joining.

I can work on cleaning it up more and adding more tests next week.

quixstreams/sinks/community/postgresql.py
```diff
@@ -60,11 +70,13 @@ def __init__(
         dbname: str,
         user: str,
         password: str,
-        table_name: Union[Callable[[SinkItem], str], str],
+        table_name: TableName,
         schema_name: str = "public",
         schema_auto_update: bool = True,
         connection_timeout_seconds: int = 30,
         statement_timeout_seconds: int = 30,
+        primary_keys: Optional[PrimaryKeys] = None,
```
gwaramadze: This should be singular "primary_key". Even if the key is compound, it is the "key", not "keys".

quixstreams/sinks/community/postgresql.py
```diff
@@ -86,6 +98,14 @@ def __init__(
         :param connection_timeout_seconds: Timeout for connection.
         :param statement_timeout_seconds: Timeout for DDL operations such as table
             creation or schema updates.
+        :param primary_keys: An optional single or multiple (composite) primary key column(s).
+            Can also provide a callable that accepts the message value as input, and returns
+            a similar result (or nothing).
```
gwaramadze: This is a bit vague. What do you mean by "similar result". Similar to what?

quixstreams/sinks/community/postgresql.py
```diff
@@ -107,6 +127,17 @@ def __init__(
         options = kwargs.pop("options", "")
         if "statement_timeout" not in options:
             options = f"{options} -c statement_timeout={statement_timeout_seconds}s"
+
+        if primary_keys and not upsert_on_primary_key:
+            logger.warning(
+                "NOTE: `primary_keys` was set but `upsert_on_primary_key` is False; "
+                "upserting is commonly desired when using primary keys."
+            )
+        self._primary_keys = (
```
gwaramadze: Perhaps the name of this attribute should hint that it is a function, such as explicitly `self._primary_key_setter`.

quixstreams/sinks/community/postgresql.py
```diff
@@ -48,10 +48,20 @@
     bool: "BOOLEAN",
 }
 
+TableName = Union[Callable[[SinkItem], str], str]
+PrimaryKeySetter = Callable[[SinkItem], Optional[Union[str, list[str]]]]
```
gwaramadze: Why return type may be `None`? Do we need this flexibility, or rather let's be more strict?

quixstreams/sinks/community/postgresql.py
```diff
@@ -125,33 +156,50 @@ def setup(self):
...
-                table["cols_types"].setdefault(_KEY_COLUMN_NAME, key_type)
-                row[_KEY_COLUMN_NAME] = item.key
-
+                table["cols_types"].setdefault(KEY_COLUMN_NAME, key_type)
+                row[KEY_COLUMN_NAME] = item.key
             for key, value in item.value.items():
-                if value is not None:
+                if key in primary_keys:
+                    table["pks_types"].setdefault(key, type(key))
+                    row[key] = value
```
gwaramadze, line 182: Just to double check: it is fine for value to be None here, right?

quixstreams/sinks/community/postgresql.py
```diff
@@ -125,33 +156,50 @@ def setup(self):
...
         except psycopg2.Error as e:
             self._client.rollback()
+            if "duplicate key value violates unique constraint" in str(e):
+                raise PostgreSQLSinkException(
+                    f"Failed to write batch: attempted insert of new row using an "
+                    f"existing primary key. Try `upsert_on_primary_key=True` for "
+                    f"upsert functionality. PostgreSQL error info: {str(e)}"
+                ) from e
             raise PostgreSQLSinkException(f"Failed to write batch: {str(e)}") from e
         table_counts = {table: len(values["rows"]) for table, values in tables.items()}
```
gwaramadze: Maybe there is a way to avoid iterating over `tables.items()` twice? You can initialize the `table_counts`  dict before the previous for loop and update it on the go.

quixstreams/sinks/community/postgresql.py
```diff
@@ -238,20 +286,103 @@ def _add_new_columns(self, table_name: str, columns: dict[str, type]) -> None:
...
+                AND tc.table_name = {table}
+                AND tc.table_schema = {schema};
+            """
+        ).format(
+            schema=sql.Literal(self._schema_name),
+            table=sql.Literal(table),
+        )
+        with self._client.cursor() as cursor:
+            cursor.execute(query)
+            pks = list(sorted(row[0] for row in cursor.fetchall()))
```
gwaramadze, line 312: tiny redundancy, `sorted` returns a list
```suggestion
            pks = sorted(row[0] for row in cursor.fetchall())
```

quixstreams/sinks/community/postgresql.py
```diff
@@ -238,20 +286,103 @@ def _add_new_columns(self, table_name: str, columns: dict[str, type]) -> None:
...
+        return pks
+
+    def _add_new_primary_keys(self, table_name: str, pks: dict[str, type]) -> None:
+        current_pks = self.get_current_primary_keys(table_name)
+        _pks = list(sorted(pks.keys()))
```
gwaramadze: tiny redundancy, `sorted` returns a list
```suggestion
        _pks = sorted(pks.keys())
```

quixstreams/sinks/community/postgresql.py
gwaramadze: tiny redundancy, sorted returns a list
```suggestion
            pks = sorted(row[0] for row in cursor.fetchall())
```

quixstreams/sinks/community/postgresql.py
```diff
@@ -238,20 +286,103 @@ def _add_new_columns(self, table_name: str, columns: dict[str, type]) -> None:
...
         columns = list(rows[0].keys())
-        # Handle missing keys gracefully
-        values = [[row.get(col, None) for col in columns] for row in rows]
 
         query = sql.SQL("INSERT INTO {table} ({columns}) VALUES %s").format(
             table=sql.Identifier(self._schema_name, table_name),
             columns=sql.SQL(", ").join(map(sql.Identifier, columns)),
         )
+        if pks and self._do_upsert:
+            # we must de-deplicate based on primary keys else an exception is thrown.
```
gwaramadze:
```suggestion
            # We must de-duplicate based on primary keys, else an exception is thrown.
```

tests/test_quixstreams/test_sinks/test_community/test_postgres_sink.py
```diff
@@ -0,0 +1,590 @@
...
+        host=postgres_container.get_container_host_ip(),
+        port=postgres_container.get_exposed_port(postgres_container.port),
+        user=postgres_container.username,
+        password=postgres_container.password,
+        options=f"-c statement_timeout={20}s",
+    ) as conn:
+        yield conn
+
+
+@pytest.fixture()
```
gwaramadze: Maybe `autouse=True` without inner func?

quixstreams/sinks/community/postgresql.py
tim-quix: Agreed, though I was wanting to convey the fact that you could provide a list of columns. 

I'll change it to `primary_key_columns` to convey that better.

quixstreams/sinks/community/postgresql.py
tim-quix: Was referring to the single or list of key columns. I'll word it better =)

quixstreams/sinks/community/postgresql.py
tim-quix: It could be that you don't have a primary key for a given table, but I think you follow that much =). 

I thought allowing `None` was maybe more obvious in terms of lack of a PK compared to say, an empty list or string.

If you think it would otherwise read better, I can change it. I'd also probably change `Optional[PrimaryKey]=None` to just be an empty iterable by default then.

quixstreams/sinks/community/postgresql.py
tim-quix: If it's a part of a primary key, it cannot be None and would error out when you try to push it IIRC.

I could do an extra check here and more explicitly raise an error, though I think the error was pretty easy to follow.

quixstreams/sinks/community/postgresql.py
tim-quix: IDK why I thought it returned a 1-time iterator....thanks!

quixstreams/sinks/community/postgresql.py
tim-quix: Went ahead and changed it and make an explicit error message.

quixstreams/sinks/community/postgresql.py
tim-quix: Simplified this to no longer use Optional/None

docs/connectors/sinks/postgresql-sink.md
```diff
@@ -91,7 +91,14 @@ PostgreSQLSink accepts the following configuration parameters:
 - `schema_name`: The schema name. Schemas are a way of organizing tables and 
   not related to the table data, referenced as `<schema_name>.<table_name>`.  
   PostrgeSQL uses "public" by default under the hood.
-- `schema_auto_update`: If True, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
+- `schema_auto_update`: If `True`, the sink will automatically update the schema by adding new columns when new fields are detected. Default: True.
+- `primary_keys`: A single or multiple (composite) primary key columns.
```
gwaramadze:
```suggestion
- `primary_key_columns`: A single or multiple (composite) primary key columns.
```

quixstreams/sinks/community/postgresql.py
```diff
@@ -86,6 +101,16 @@ def __init__(
...
         :param statement_timeout_seconds: Timeout for DDL operations such as table
             creation or schema updates.
+        :param primary_key_columns: An optional single (string) or list of primary key
+            column(s); len>1 is a composite key, a non-empty str or len==1 is a primary
+            key, and len<1 or empty string means no primary key.
+            Can instead provide a callable, which uses the message value as input and
+            returns a string or list of strings.
+            Often paired with `upsert_on_primary_key=True`.
+            It must include all currently defined primary key columns on a given table.
+        :param upsert_on_primary_key: Upsert based on the given `primary_keys`.
```
gwaramadze:
```suggestion
        :param upsert_on_primary_key: Upsert based on the given `primary_key_columns`.
```

gwaramadze (commented):
docs to correct

docs/connectors/sinks/postgresql-sink.md
```diff
@@ -91,7 +91,14 @@ PostgreSQLSink accepts the following configuration parameters:
...
+  Can also provide a callable that accepts the message value as input, and returns a similar result (or nothing).
```
gwaramadze:
```suggestion
  Can also provide a callable that accepts the message value as input, and returns a string or list of strings.
```

quixstreams/sinks/community/postgresql.py
gwaramadze: That's true, I just realized that tables are not forced to have PK at all. It's a weird setup but allowed.

gwaramadze (approved):

PR #971, opened 2025-07-14
feat(lookup.quix_configuration_service): add support for binary and JSON fields

ovv:
…SON fields

- Introduce BytesField and JSONField for the Quix Configuration Lookup
- Update Quix Configuration Lookup to provide json_field and bytes_field helpers
- Refactor tests to use new field types and helpers

This enables extraction of both JSON and binary content from configuration services, improving flexibility for downstream consumers.


gwaramadze (approved):
LGTM!

PR #976, opened 2025-07-17
Feature: dynamic target topic routing

gwaramadze:
- Enhanced `StreamingDataFrame.to_topic()` to accept a callable that returns a `Topic` object based on message content (value, key, timestamp, headers).
- Enables dynamic message routing to different topics based on message content.
- Maintains backward compatibility with existing `Topic` object usage
- Includes performance guidance to prevent accidental topic proliferation

### DIFF
--- a/quixstreams/dataframe/dataframe.py
+++ b/quixstreams/dataframe/dataframe.py
@@ -669,7 +669,7 @@ class StreamingDataFrame:
 
     def to_topic(
         self,
-        topic: Union[Topic, Callable[[Any], Topic]],
+        topic: Union[Topic, Callable[[Any, Any, int, Any], Topic]],
         key: Optional[Callable[[Any], Any]] = None,
     ) -> "StreamingDataFrame":
         """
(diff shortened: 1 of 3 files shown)


quixstreams/dataframe/dataframe.py
```diff
@@ -703,7 +705,7 @@ def to_topic(
         """
         return self._add_update(
             lambda value, orig_key, timestamp, headers: self._produce(
-                topic=topic,
+                topic=topic if isinstance(topic, Topic) else topic(value),
```
gwaramadze: We could get rid of the `if` by duplicating code.

gwaramadze: @daniil-quix WDYT?

quixstreams/dataframe/dataframe.py
```diff
@@ -668,7 +668,9 @@ def callback(value, *_):
         return StreamingSeries.from_apply_callback(callback, sdf_id=id(self))
 
     def to_topic(
-        self, topic: Topic, key: Optional[Callable[[Any], Any]] = None
+        self,
+        topic: Union[Topic, Callable[[Any], Topic]],
```
ghost: Let's make it a `metadata=True` callback so it accepts also key, timestamp and headers.

quixstreams/dataframe/dataframe.py
```diff
@@ -668,7 +668,9 @@ def callback(value, *_):
         return StreamingSeries.from_apply_callback(callback, sdf_id=id(self))
 
     def to_topic(
-        self, topic: Topic, key: Optional[Callable[[Any], Any]] = None
+        self,
+        topic: Union[Topic, Callable[[Any, Any, int, Any], Topic]],
+        key: Optional[Callable[[Any], Any]] = None,
```
ghost: Let's also add the `@overload` here similar to other methods

ghost (approved):

PR #979, opened 2025-07-19
Improvement: TDengine sink adds empty value checking and supports string-formatted timestamps

huskar-t:
This pull request introduces enhancements and refactoring to the `TDengineSink` implementation, focusing on improved timestamp handling, stricter validation, and better error reporting.

* Added a new `time_setter` parameter to `TDengineSink`, which can accept a column name or a callable to determine the timestamp. It supports `int`, `string` (RFC3339 format), or `datetime` values. A `ValueError` is raised if both `time_key` and `time_setter` are set.
* Added checks to ensure that supertable names are non-empty and fields are present. Raises `ValueError` if these conditions are not met.
* Enhanced error reporting for overlapping keys between fields and tags, with detailed error messages. 
* Updated the documentation to reflect the new `time_setter` parameter and its behavior. Clarified the behavior of `time_key` and added notes about potential exceptions.


quixstreams/sinks/community/tdengine/sink.py
```diff
@@ -328,29 +349,72 @@ def write(self, batch: SinkBatch):
...
-                )
+                # check if fields is empty
+                if not fields:
+                    raise ValueError(
+                        f'No fields found in the record for supertable "{_measurement}" '
+                        f"and subtable name '{_subtable_name}', "
+                        f"record value: {copied_value}"
+                    )
+                
+                tags["__subtable"] = _subtable_name
```
zitsen, line 350: Can subtable_name be None/null/"" ?

huskar-t: > Can subtable_name be None/null/"" ?

If subtable_name is None or an empty string, it will be ignored when generating the line protocol.

quixstreams/sinks/community/tdengine/sink.py
```diff
@@ -47,6 +57,7 @@ def __init__(
         fields_keys: FieldsSetter = (),
         tags_keys: TagsSetter = (),
         time_key: Optional[str] = None,
+        time_setter: Optional[TimeSetter] = None,
```
tim-quix, line 59: if backwards compatibility is the only reason we're keeping `time_key` around, I'd be happy to just ditch it and just use `time_setter`...overall TDEngineSink is still pretty new and so I don't think we need to concern ourselves with backwards compatibility just yet, especially with something like this were I think having both settings as options could be confusing (especially since other connectors only have 1).

quixstreams/sinks/community/tdengine/sink.py
```diff
@@ -288,20 +299,31 @@ def write(self, batch: SinkBatch):
...
             records = []
 
-            min_timestamp = sys.maxsize
-            max_timestamp = -1
+            min_timestamp = None
+            max_timestamp = None
 
             for item in write_batch:
                 value = item.value
+                copied_value = value.copy()  # Copy to print the original value in case of errors
```
tim-quix: Because the only way we are manifesting errors is `SinkBackpressureError`, we don't need to do any data cloning as `SinkBackpressureError` re-consumes the Kafka messages (and thus uses the original data).

quixstreams/sinks/community/tdengine/sink.py
```diff
@@ -328,29 +349,72 @@ def write(self, batch: SinkBatch):
...
+                    f"record value: {copied_value}"
+                )
                 continue
             _start = time.monotonic()
             l: list[bytes] = [b""] * len(records)
             for i, point in enumerate(records):
                 p = Point.from_dict(point, self._write_precision)
                 l[i] = p.to_line_protocol().encode("utf-8")
-            body = b"\n".join(l)
+            body = b"\n".join([item for item in l if item])
```
tim-quix, line 399: Seems like this check is maybe unnecessary?

From what I can tell, you've already ensured that any given included record won't be empty during this check:

```
              # check if fields is empty
              if not fields:
                  raise ValueError(
                      f'No fields found in the record for supertable "{_measurement}" '
                      f"and subtable name '{_subtable_name}', "
                      f"record value: {copied_value}"
                  )
```
So we should be able to leave the original code, I believe?

```body = b"\n".join(l)```

tim-quix (changes requested):

quixstreams/sinks/community/tdengine/sink.py
huskar-t: I think copy is meaningful—for example, when checking if fields is empty, it can display all the values.

quixstreams/sinks/community/tdengine/sink.py
huskar-t: p.to_line_protocol may return an empty string, which could result in an empty body being generated.

quixstreams/sinks/community/tdengine/sink.py
huskar-t: done

quixstreams/sinks/community/tdengine/sink.py
tim-quix: It can be useful for sure, but just want to be mindful of the cost/implications:

1. `dictionary.copy` is just a shallow copy, so if there's any nested objects that you augment (list, dict), it will also affect the copy.
2. copying is generally not a computationally cheap operation in python, so we try to avoid it where possible.

So really this is just to emphasize that unless we really feel like we need it I recommend removing it (at least until we run into situations where it's unquestionably necessary). You might know the use cases better so I'll defer to your judgment here in terms of that =)

quixstreams/sinks/community/tdengine/sink.py
tim-quix: Oh interesting! Do you mean just in general, or specifically even when we had a non-empty `fields`? If the latter, would you mind sharing an example of it?

quixstreams/sinks/community/tdengine/sink.py
tim-quix: awesome, thanks! =D

quixstreams/sinks/community/tdengine/sink.py
huskar-t: removed `copied_value`

quixstreams/sinks/community/tdengine/sink.py
huskar-t: If a field's value is None or is finite , the field will be skipped

huskar-t:
Added retry mechanism

tim-quix (dismissed):

tim-quix:
Linters need to be run/applied. Might be easier to do ourselves post-merge?

huskar-t:
Fixed the formatting issues.

ghost (approved):

PR #1004, opened 2025-08-13
Connector: FileSink refactor

tim-quix:
Refactored FileSink to be more similar to how FileSource is structured (which was originally changed per Tomas request to simplify it's use). 

Now it no longer requires passing in separate "Destination" objects to a FileSink.


quixstreams/sinks/community/file/azure.py
```diff
@@ -105,4 +130,8 @@ def write(self, data: bytes, batch: SinkBatch) -> None:
             self._container,
             file_name,
         )
-        self._client.get_blob_client(file_name).upload_blob(data)
+        try:
+            self._client.get_blob_client(file_name).upload_blob(data)
+        except Exception as e:
+            print(e.__class__, str(e))
```
gwaramadze: This print looks like a leftover from testing. I think we can remove the `try/except` here.

quixstreams/sinks/community/file/base.py
```diff
@@ -0,0 +1,137 @@
...
+        Write a batch of data using the configured format.
+
+        :param batch: The batch of data to write.
+        """
+        data = self._format.serialize(batch)
+
+        try:
+            self._write(data, batch)
+        except Exception as e:
+            logger.error(f"Sink write attempt encountered an error: {e}")
```
gwaramadze: Do we need to log the error if it is going to be raised anyway?

quixstreams/sinks/community/file/azure.py
tim-quix: yeah, was left over, thanks for the catch =)

quixstreams/sinks/community/file/base.py
tim-quix: overall, prob not...it's prob clear enough if an error is manifesting why/how it happened. I'll remove it :+1:

gwaramadze (approved):

PR #1012, opened 2025-08-22
Connectors: fix FileSink/Source-based dependencies

tim-quix:
Fixed issue around optional dependencies with FileSink not caught during refactor.

This also fixes some related import issues in FileSource as well.

Also some other minor cleanup.


quixstreams/sinks/community/file/formats/parquet.py
```diff
@@ -42,6 +39,25 @@ def __init__(
...
+    def _get_writer(self):
+        try:
+            import pyarrow as pa
+            import pyarrow.parquet as pq
+        except ImportError as exc:
+            raise ImportError(
+                f"Package {exc.name} is missing: "
+                'run "pip install quixstreams[parquet]" to use the Parquet file format '
+                '(Note: options can be installed together i.e. "quixstreams[s3,parquet]")'
+            ) from exc
```
gwaramadze, line 53: What is the reason for this here? I believe this should live in the imports section at the top of the module.

quixstreams/sources/community/file/formats/parquet.py
```diff
@@ -18,7 +18,14 @@ def __init__(
...
             )
-        import pyarrow.parquet as pq
+        try:
+            import pyarrow.parquet as pq
+        except ImportError as exc:
+            raise ImportError(
+                f"Package {exc.name} is missing: "
+                'run "pip install quixstreams[parquet]" to use the Parquet file format '
+                '(Note: options can be installed together i.e. "quixstreams[s3,parquet]")'
+            ) from exc
```
gwaramadze, line 28: Should we move it to the top of the module too?

quixstreams/sinks/community/file/formats/parquet.py
tim-quix: IIRC, I believe this was to avoid import errors when not using Parquet.

gwaramadze (approved):

PR #1044, opened 2025-10-09
Feature: Early Window Expiration with Triggers

gwaramadze:
## Early Window Expiration with Triggers

This PR adds support for early window expiration via callback triggers for tumbling and hopping windows.

### Feature Overview

Windows can now be expired early based on custom logic by providing `before_update` or `after_update` callbacks when defining a window. When these callbacks return `True`, the window is immediately expired rather than waiting for its natural expiration time.

### API

Two new optional parameters added to `tumbling_window()` and `hopping_window()`:

- **`before_update`**: Callback invoked before updating the window state
  - Receives the current aggregated value (or collection) before the new value is applied
  - Useful for preventing a value from being included if a threshold would be exceeded

- **`after_update`**: Callback invoked after updating the window state  
  - Receives the updated aggregated value (or collection) after the new value is applied
  - Useful for expiring a window once a threshold is reached

Both callbacks receive: `(aggregated, value, key, timestamp, headers) -> bool`

### Examples

**After Update - Expire when sum exceeds threshold:**
```python
def trigger_on_sum(aggregated, value, key, timestamp, headers):
    return aggregated > 100  # Expire when sum exceeds 100

sdf = app.dataframe(topic)
sdf = sdf.tumbling_window(duration_ms=60000, after_update=trigger_on_sum).sum().final()
```

**Before Update - Prevent value from exceeding threshold:**
```python
def trigger_before_exceeding(aggregated, value, key, timestamp, headers):
    return (aggregated + value) > 100  # Expire before sum would exceed 100

sdf = app.dataframe(topic)
sdf = sdf.tumbling_window(duration_ms=60000, before_update=trigger_before_exceeding).sum().final()
```

**With collect operations:**
```python
def trigger_on_count(aggregated, value, key, timestamp, headers):
    return len(aggregated) >= 10  # Expire when 10 items collected

sdf = app.dataframe(topic)
sdf = sdf.tumbling_window(duration_ms=60000, after_update=trigger_on_count).collect().final()
```

### Implementation Details

- Triggered windows are moved to the expired windows list and deleted from state
- Collection values are not immediately deleted; normal expiration handles cleanup
- This means triggered windows can be "resurrected" if new data arrives before natural expiration
- Works with both aggregation operations (`.sum()`, `.mean()`, etc.) and `.collect()`
- Fully compatible with overlapping hopping windows

### DIFF
--- a/docs/windowing.md
+++ b/docs/windowing.md
@@ -596,19 +596,21 @@ if __name__ == '__main__':
 ### Early window expiration with triggers
 !!! info New in v3.24.0
 
-To expire windows before their natural expiration time based on custom conditions, you can pass the `on_update` callback to `.tumbling_window()` and `.hopping_window()` methods.
+To expire windows before their natural expiration time based on custom conditions, you can pass `before_update` or `after_update` callbacks to `.tumbling_window()` and `.hopping_window()` methods.
 
 This is useful when you want to emit results as soon as certain conditions are met, rather than waiting for the window to close naturally.
 
 **How it works**:
 
-- The `on_update` callback is invoked every time a window is updated with a new value.
-- It receives `old_value` and `new_value` - the raw aggregated values before and after the update.
-- For `collect()` operations, it receives lists of collected values.
-- If the callback returns `True`, the window is immediately expired and emitted downstream.
-- The expired window is removed from state, but may be "resurrected" if new data arrives within its time range before natural expiration.
+- The `before_update` callback is invoked before the window aggregation is updated with a new value.
+- The `after_update` callback is invoked after the window aggregation has been updated with a new value.
+- Both callbacks receive: `aggregated` (current or updated aggregated value), `value` (incoming value), `key`, `timestamp`, and `headers`.
+- For `collect()` operations without aggregation, `aggregated` contains the list of collected values.
+- If either callback returns `True`, the window is immediately expired and emitted downstream.
+- The window metadata is deleted from state, but collected values (if using `.collect()`) remain until natural expiration.
+- This means a triggered window can be "resurrected" if new data arrives within its time range - a new window will be created with the previously collected values still present.
 
-**Example**:
+**Example with after_update**:
 
 ```python
 from typing import Any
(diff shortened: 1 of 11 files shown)


docs/windowing.md
```diff
@@ -593,6 +593,53 @@ if __name__ == '__main__':
...
+
+This is useful when you want to emit results as soon as certain conditions are met, rather than waiting for the window to close naturally.
+
+**How it works**:
+
+- The `on_update` callback is invoked every time a window is updated with a new value.
+- It receives `old_value` and `new_value` - the raw aggregated values before and after the update.
+- For `collect()` operations, it receives lists of collected values.
+- If the callback returns `True`, the window is immediately expired and emitted downstream.
+- The expired window is removed from state, but may be "resurrected" if new data arrives within its time range before natural expiration.
```
ghost: Can you please clarify what "resurrection" means here? How does it work in practice?

quixstreams/dataframe/windows/definitions.py
```diff
@@ -379,13 +389,20 @@ def __init__(
         dataframe: "StreamingDataFrame",
         name: Optional[str] = None,
         on_late: Optional[WindowOnLateCallback] = None,
+        on_update: Optional[WindowOnUpdateCallback] = None,
     ):
+        if on_update is not None:
+            raise ValueError(
+                "Sliding windows do not support the 'on_update' trigger callback. "
+                "Use tumbling or hopping windows instead."
```
ghost, line 408:
```suggestion
                "Only tumbling and hopping windows support it as of now."
```

quixstreams/dataframe/windows/time_based.py
```diff
@@ -198,7 +270,10 @@ def process_window(
                 key, state, max_expired_window_start, collect
             )
 
-        return updated_windows, expired_windows
+        # Combine triggered windows with time-expired windows
+        all_expired_windows = triggered_windows + list(expired_windows)
+
+        return updated_windows, iter(all_expired_windows)
```
ghost: Do we need `iter()` here when `all_expired_windows` is already a list?

gwaramadze: Nope, correcting

quixstreams/dataframe/windows/time_based.py
```diff
@@ -198,7 +270,10 @@ def process_window(
                 key, state, max_expired_window_start, collect
             )
 
-        return updated_windows, expired_windows
+        # Combine triggered windows with time-expired windows
+        all_expired_windows = triggered_windows + list(expired_windows)
```
```suggestion
        all_expired_windows = itertools.chain(expired_windows, triggered_windows)
```

Two things here:
1. `itertools.chain()` allows us to not evaluate the iterator straight away. (`import itertools` is needed at the top)
2. It's better to process the expired windows first and only then triggered to keep the track of time in order downstream.

ghost (approved):

PR #1046, opened 2025-10-14
WIP: Watermarking

ghost:


quixstreams/core/stream/functions/apply.py
```diff
@@ -47,12 +47,23 @@ def wrapper(
...
+                        value,
+                        key,
+                        timestamp,
+                        headers,
+                        True,
+                    )
+                else:
+                    result = func(value)
+                    for item in result:
+                        child_executor(item, key, timestamp, headers)
```
gwaramadze, line 65: That may not change the game entirely, but still.
```suggestion
                result = value if is_watermark else func(value)
                child_executor(result, key, timestamp, headers, is_watermark)
```
But with this `is_watermark` could also be left as a mandatory arg.

Issue #1072, opened 2026-01-15
InfluxDB3Sink TypeError multiple values for keyword argument 'timeout'


zenzxjul:
**Tell us about the bug**

The InfluxDB3Sink passes the `timeout` parameter as keyword argument to influxdb_client_3.InfluxDB3Client via `_client_args` even though that class now takes a new positional argument `write_timeout` causing multiple values for `timeout` argument of influxdb_client_3.write_client.InfluxDBClient

See [release 0.17.0](https://github.com/InfluxCommunity/influxdb3-python/releases/tag/v0.17.0) or more specifically [PR 158](https://github.com/InfluxCommunity/influxdb3-python/pull/158) for details.

**What did you expect to see?**
Correct use of influxdb_client_3.InfluxDB3Client initialization.

**What version of the library are you using?**
quixstreams[influxdb3]==3.23.1

**Workaround?**
Limit influxdb3<0.17.0

**Fix?**
Update `_client_args` to define `write_timeout` instead of `timeout`

PR #1073, opened 2026-01-15
Fix Influxdb3 sink extra timeout argument

### LINKED ISSUE #1071: InfluxDB3 sink crashes with influxdb3-python >= 0.17.0 due to extra timeout argument
**Tell us about the bug**  
When initializing the InfluxDB3 sink, an extra `timeout` argument is passed to the client initialization, which causes the sink to crash.

The InfluxDB3 sink should initialize without crashing.

This happens with any version of `quixstreams` if the Docker image is built after Jan 08, the release date of `influxdb3-python` `0.17.0`.

**Workaround**  
Add to `requirements.txt`:
```text
influxdb3-python==0.16.0
```

Since the sink wraps the client initialization, this should ideally be handled internally. If I’m using the sink, I shouldn’t need to care which version of `influxdb3-python` to use, as it should be managed by `quixstreams` dependencies.

rodas96:
Fixes #1071 and #1072.

The `WriteOptions` class in `influxdb3-python` v0.17+ added a `timeout` parameter. Previously, `timeout` was passed at the top level to `InfluxDBClient3`, but in v0.17 this causes a "got multiple values for keyword argument 'timeout'" error because:

1. The top-level `timeout` is passed to `_InfluxDBClient` via `**kwargs`
2. The `timeout` inside `write_client_options` is spread and passed to `_WriteApi`
3. Both clients now receive `timeout`, causing a conflict

This PR moves the `timeout` parameter into the `WriteOptions` object to be compatible with v0.17+.

**Breaking change:** This requires `influxdb3-python>=0.17.0`. Users who have pinned to v0.16 to quick fix this will need to update.

### DIFF
--- a/quixstreams/sinks/core/influxdb3.py
+++ b/quixstreams/sinks/core/influxdb3.py
@@ -2,7 +2,6 @@ import logging
 import sys
 import time
 from datetime import datetime, timezone
-from importlib import metadata
 from typing import Any, Callable, Iterable, Literal, Mapping, Optional, Union, get_args
 
 import httpx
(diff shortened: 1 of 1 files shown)


SteveRosam:
@rodas96 thank you for this. I have tested and can confirm it's working now. Will merge and release.

SteveRosam (approved):

PR #1079, opened 2026-02-11
Add quixdatalake sink

SteveRosam:


quixstreams/sinks/core/quix_lake_blob_storage.py
```diff
@@ -0,0 +1,643 @@
...
+# Timestamp column mappers for Hive partitioning
+TIMESTAMP_COL_MAPPER = {
+    "year": lambda col: col.dt.year.astype(str),
+    "month": lambda col: col.dt.month.astype(str).str.zfill(2),
+    "day": lambda col: col.dt.day.astype(str).str.zfill(2),
+    "hour": lambda col: col.dt.hour.astype(str).str.zfill(2),
+}
+
+
+class QuixLakeBlobStorageSink(BatchingSink):
```
tomas-quix: Should we name the file same? We forgot sink in filename.

quixstreams/sinks/core/_catalog_client.py
```diff
@@ -0,0 +1,71 @@
...
+
+Simple HTTP client for REST Catalog API interactions.
+"""
+
+from typing import Optional
+
+import requests
+
+
+class CatalogClient:
```
tomas-quix: Should we call this quix-ts-datalake-catalog-client.py and Class equally?

quixstreams/sinks/core/quix_lake_blob_storage.py
tomas-quix: I mean I would expect QuixTSDataLakeSink

tomas-quix: to get align with Emanuels renaming

SteveRosam: none of the other sinks have sink in the filename

tomas-quix: ok fine, but at least match the Emanuel new naming please

tomas-quix (approved):

PR #1085, opened 2026-02-24
Detect prolonged broker unavailability and crash for orchestrator restart

peter-quix:
Features
- Track per-broker connectivity via stats_cb on both Producer and BaseConsumer, logging state transitions (broker down/restored) for observability
- Detect prolonged total broker unavailability (_ALL_BROKERS_DOWN) with periodic active metadata probes, raising KafkaBrokerUnavailableError after a configurable timeout to allow orchestrator restarts
- New Application(broker_availability_timeout=...) parameter (default 120s, set 0 to disable) wired into the main processing loop, sources-only loop, and recovery loop

Refactors
- _ALL_BROKERS_DOWN no longer silently ignored in producer error callback — now logged at WARNING level and tracked for availability detection

Docs
- Updated API reference documentation for the new parameter and broker availability behaviour

### DIFF
--- a/quixstreams/kafka/consumer.py
+++ b/quixstreams/kafka/consumer.py
@@ -37,6 +37,8 @@ AutoOffsetReset = typing.Literal["earliest", "latest", "error"]
 
 logger = logging.getLogger(__name__)
 
+_BROKER_PROBE_INTERVAL = 30.0
+
 
 def _default_error_cb(error: KafkaError):
     error_code = error.code()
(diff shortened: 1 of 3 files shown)


peter-quix:
When a single broker fails and recovers, users will see:

Broker goes down (from librdkafka + our stats_cb):
```
ERROR quixstreams.kafka.producer: FAIL [rdkafka#producer-1] [thrd:localhost:9094/3]: localhost:9094/3: Disconnected (after 47149ms in state UP)
INFO  quixstreams.kafka.producer: Kafka producer: broker localhost:9094/3 (node 3) went down; 2 other broker(s) still available
```

Broker comes back (from our stats_cb):
```
INFO  quixstreams.kafka.producer: Kafka producer: broker localhost:9094/3 (node 3) is UP again (was INIT)
```

The same pattern applies for the consumer, with Kafka consumer: instead of Kafka producer:
What users won't see:
- No log when brokers first connect during startup
- No "went down" reassurance when all brokers are down (the _ALL_BROKERS_DOWN error path covers that)
- No "UP again" for brokers that were never UP before (e.g. INIT → UP during initial connection)
- No repeated messages — each transition logs exactly once

SteveRosam (approved):

PR #1092, opened 2026-04-08
Feature: InfluxDB3 sink verify_ssl param

cbradfordaws:
When connecting to an instance of InfluxDB3 that is using self signed certificates an error is received as the certificate cannot be verified. This feature adds a wrapper for the InfluxDB3 client to be able to pass the verify_ssl parameter. Default value is set to True.


tomas-quix (approved):
Amazing, thank you for your contribution!

Issue #1096, opened 2026-04-14
Redis Sink : Add TTL support to prevent memory exhaustion


terencebon1n:
**Is your feature request related to a problem? Please describe.**
Currently, the RedisSink.write method uses a standard pipe.set(key, value) call during the batching process. It does sinks data to Redis, but it creates keys with no expiration.

My keys are quite volatile, this leads to unbounded memory growth. Without a mechanism to expire old data, the Redis instance eventually hits its maxmemory limit.

**Describe the solution you'd like**
Modify the RedisSink class to accept an optional `ttl` parameter (in seconds) in the `__init__ `method. This TTL should then be applied during the write operation using the `ex` argument in the `set` command.

```
# Current
pipe.set(key, value)

# Proposed
pipe.set(key, value, ex=self._ttl)
```

terencebon1n:
For anyone with the same issue, I went with Custom Sink instead, closing that issue as no answer were provided after 2 months

PR #1097, opened 2026-04-14
Add TTL support to RedisSink

terencebon1n:
CF : #1096


terencebon1n:
For anyone with the same issue, I went with Custom Sink instead, closing that PR as no answer were provided after 2 months.

PR #1103, opened 2026-05-12
fix(quix-ts-datalake-sink): stop silently dropping rows with NULL par…

tomas-quix:
…tition values

`_write_batch` builds a DataFrame from the batch and groups by the configured `hive_columns` before writing one parquet per partition. `pandas.DataFrame.groupby` defaults to `dropna=True`, so any row whose partition key contains NaN (the partition field was `None` in the payload, or absent entirely) was silently excluded from iteration. The for-loop never produced a group for those rows, `_write_parquet_to_storage` was never called, and `_finalize_writes` returned early without logging. Yet `write()` still reported success using `batch.size` — so the sink logged "Wrote N rows to blob storage" while nothing reached storage.

In the worst case where every row of every batch was missing one partition column, the sink ran for hours producing only successful- looking logs and zero PUTs to Azure/S3.

Fix:
- Fill NaN partition values with `__HIVE_DEFAULT_PARTITION__` before grouping. This is the Hive/Iceberg/Spark convention; DuckDB's `hive_partitioning=true` reader, Spark, and Trino all materialise the column as SQL NULL when they encounter this sentinel.
- `_write_batch` now returns the rows actually grouped and written.
- `write()` logs that count instead of `batch.size`, so any future silent-drop regression is visible in the log instead of being papered over with the input count.

Adds two regression tests covering mixed-null and all-null batches.

### DIFF
--- a/quixstreams/sinks/core/quix_ts_datalake_sink.py
+++ b/quixstreams/sinks/core/quix_ts_datalake_sink.py
@@ -45,6 +45,12 @@ TIMESTAMP_COL_MAPPER = {
     "hour": lambda col: col.dt.hour.astype(str).str.zfill(2),
 }
 
+# Hive/Iceberg/Spark convention for partition values that are NULL. Readers
+# that respect the convention (DuckDB hive_partitioning, Spark, Trino…)
+# materialise the column as NULL when they encounter this sentinel in a
+# path segment.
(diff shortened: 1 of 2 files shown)


LudvikQuix (approved):
review ok

PR #1110, opened 2026-05-27
feat(lookup): grace period for late configuration, resolved at the deadline with or without traffic

LudvikQuix:
Gives `join_lookup` a grace period for configuration that has not arrived yet: a record that cannot resolve is held in durable state with its own deadline, released enriched when its config lands, and resolved through `on_timeout` when the deadline passes first — **whether or not any further traffic arrives**.

> **Reviewers: this PR changes checkpoint commit semantics.** See *Shared machinery* below. That part reaches every store and operator in the SDK, not just this feature, and is the piece most worth your attention.

## What it does

```python
from quixstreams.dataframe.joins.lookups import LookupBuffer

sdf = sdf.join_lookup(
    lookup,
    fields,
    buffer=LookupBuffer(grace_ms=2000, is_resolved=lambda v: v["threshold"] is not None),
)
```

`buffer=None` is the default and leaves the existing `update()` path byte-for-byte unchanged. This is additive: no existing pipeline changes behaviour.

`grace_ms` here is **wall-clock time from the record's arrival**, not the event-time retention that `grace_ms` means on `join_asof`, `join_interval` and windows. What is being waited for is a message on another topic, which has no event-time analogue. The docstring says so at the point of use.

## Why not the blocking design this PR originally proposed

It slept on the processing thread inside `join()`. The run loop is one thread draining one `poll_row()` across every assigned partition, so the stall was never scoped to a partition — it froze every key on the instance and delayed `commit_checkpoint()` with it. Nothing remembered unresolvable keys, so every record for a permanently unconfigured key paid the full window. Four keys across two partitions with one config missing — an ordinary partially-configured deployment — at `grace_ms=2000` and 10 msg/s on that key needs 20s of stall per wall-clock second: unbounded lag, then eviction past `max.poll.interval.ms`, rebalance, replay, stall again.

## Shared machinery

Two changes outside the feature were required, and both are proven against the full state suite (789 passed, baseline) and `test_checkpointing` + `test_internal_producer` (106 passed).

**`BaseCheckpoint.empty()` now also consults pending state changes.** It counted stored offsets only. A deadline pass consumes no message, so it stores no offset, so the checkpoint reported empty, was closed rather than committed, and its store transaction was discarded — *after* its emissions were queued on the producer. The same records would come due again every commit interval, forever, while idle. `commit()` now tolerates zero offsets, and under exactly-once a pass-only checkpoint commits while skipping `send_offsets_to_transaction` rather than aborting, which would livelock silently. **Any checkpoint carrying offsets behaves exactly as before**; state from a failed message still cannot be committed, because an unsuppressed exception never reaches the commit and `_on_revoke` drops the checkpoint when `_failed` is set.

**`Stream.add_function`** lets the buffer bind its resolved downstream executor at compose time. `expand=True` cannot serve this path — it expands an iterable returned from a transform applied to an *arriving* record, and a deadline pass has none. Each released record carries a `MessageContext` rebuilt from its envelope, so `to_topic` routes it and downstream operators still see it.

**`_get_items` gained an optional `limit`** that stops the iteration itself rather than slicing afterwards. With `limit=None` it is structurally identical to before.

## Behaviour worth knowing

- **Release is per lookup key.** A record whose own key has not resolved stays buffered at its original arrival time when a sibling sharing its message key resolves. The trade: a resolving record can overtake an older one waiting on a different lookup key. Order *within* a lookup key is preserved.
- **A value the store cannot hold is settled, not raised.** `bytes`, `tuple`, `set` and `frozenset` round-trip intact. A shape JSON cannot express — a dict with non-`str` keys, an integer outside 64 bits, an arbitrary object, a reference cycle — resolves immediately through `on_timeout` with a rate-limited warning naming the key and the refused path, rather than failing the transaction with the offset uncommitted.
- **Sub-second `grace_ms` warns.** Resolution granularity is the consumer poll timeout, ~1s.
- **The index format changed** between builds of this PR. Bounded to pre-release builds, but a deployment running an earlier commit of this branch orphans its buffered records on upgrade.

## Testing

`pre-commit run --all-files` green. Non-broker buffer and tick suites 135 passed / 2 xfailed. Full `test_dataframe/test_joins/` green including broker-backed suites. `test_state/` 789 passed, exact baseline.

Durability is tested, not argued: held records survive process death and a wiped state directory, recover from the changelog, and release enriched in arrival order with their own timestamps. One honest limit — the rebalance leg stops one app and starts another in the same consumer group, driving a real `_on_revoke`/`_on_assign` cycle, rather than two consumers rebalancing concurrently. Stated in that file's module docstring rather than absorbed.

## Reviews

Two external reviews found nine defects between them; all are fixed, each backed by a test that was red first. Sibling-key data loss through unescaped prefixes; a `bytes`-valued field crash-looping before its offset committed; a pending index scaling with waiting keys, then with *overdue* keys; `use_changelog_topics=False` silently discarding held records; unbounded bookkeeping logs; a null message key killing any unkeyed topic; a record leaving early when a sibling resolved a different lookup key; and the unserializable-value crash-loop above.

The prefix collision also exists in `TimestampedPartitionTransaction`, so it reaches released `join_asof`/`join_interval`. Fixing it there changes on-disk layout and needs a migration, so this PR escapes only its own buffer prefixes and leaves the shared store additive-only. **That is a live silent data-loss bug in two released features** — tracked as #1148 item 5, with the two tests proving it kept as `xfail(strict=True)` so they fail loudly the day it is fixed. It warrants a release note independently of this PR.

## Related

Five pre-existing defects found along the way are tracked in #1148. None is caused by this change and none blocks it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: 'CI'
 
 on:
   pull_request:
+  workflow_dispatch:
 
 
 # Run only one workflow per PR or commit
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,9 @@ site/
 memories/
 
 .claude/
+
+# scratch
+.tmp/
+
+# local planning artifacts (spec + architecture)
+dev-planning/
--- a/docs/api-reference/quixstreams.md
+++ b/docs/api-reference/quixstreams.md
@@ -6542,6 +6542,25 @@ Hive-partitioned Parquet files, with optional REST Catalog integration.
 
 Uses quixportal for unified blob storage access (Azure, AWS S3, GCP, MinIO, local).
 
+<a id="quixstreams.sinks.core.quix_ts_datalake_sink.silence_chatty_loggers"></a>
+
+#### silence\_chatty\_loggers
+
+```python
+def silence_chatty_loggers() -> None
+```
+
+[[VIEW SOURCE]](https://github.com/quixio/quix-streams/blob/main/quixstreams/sinks/core/quix_ts_datalake_sink.py#L74)
+
+Mute per-request HTTP logging from the cloud-storage SDKs used by
+this sink (Azure SDK + adlfs, botocore/boto3 + s3transfer).
+
+Safe to call from application code at any point. Levels are raised to
+WARNING, so anything actually noteworthy (auth failures, retries,
+throttling, server errors) still propagates. Call after configuring
+your own logging (e.g. after instantiating quixstreams.Application)
+so the framework's logging setup does not reset these levels.
+
 <a id="quixstreams.sinks.core.quix_ts_datalake_sink.QuixTSDataLakeSink"></a>
 
 ### QuixTSDataLakeSink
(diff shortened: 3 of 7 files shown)


LudvikQuix:
was covered by https://github.com/quixio/quix-streams/pull/1151

PR #1111, opened 2026-05-28
fix(quix-ts-datalake-sink): literal NULL partition + invariant ts_ms dtype    

tomas-quix:
## Summary                                                                                                                                                                                                                 
                                                                                                                                                                                                                             
  Two related correctness fixes in `QuixTSDataLakeSink`:                                                                                                                                                                     
                                                                                                                                                                                                                             
  1. **NULL partitions are now registered in the catalog as the literal `__None__` string** instead of being silently translated to SQL `NULL` (`quix_ts_datalake_sink.py:_register_files_in_manifest`).                     
  2. **`ts_ms` (and any user-configured `timestamp_column`) keeps its input dtype** when `HIVE_COLUMNS` includes `year` / `month` / `day` / `hour` (`quix_ts_datalake_sink.py:_add_timestamp_columns`).                      
                                              
  ## Why                                                                                                                                                                                                                     
                                                            
  ### NULL-partition leak (commit 1)                                                                                                                                                                                         
                                                            
  The sink wrote NULL-partition rows to disk under `col=__None__/` but POSTed the catalog manifest with `partition_<col>=NULL`. The on-disk sentinel and the catalog row disagreed for every file the sink registered. The   
  reading side of the lake now treats partition values as opaque strings end-to-end — the catalog, the on-disk path, and what DuckDB's `hive_partitioning=true` exposes at query time must all agree on the same literal. 
  With the old translation in place, the partition tree split one logical NULL bucket into two (files lived under `__None__/` but their catalog rows showed up under an empty bucket).

  Fix: send `str(val)` for every partition value, no NULL magic.                                                                                                                                                             
                                              
  ### Timestamp dtype mutation (commit 2)                                                                                                                                                                                    
                                                            
  `_add_timestamp_columns` overwrote `df[timestamp_column]` in place when converting numeric timestamps to `datetime64` for year/month/day/hour extraction. As a result:                                                     
                                              
  - `HIVE_COLUMNS` with no time columns → parquet stored `ts_ms` as **BIGINT**.                                                                                                                                              
  - `HIVE_COLUMNS` with any time column → parquet stored `ts_ms` as **TIMESTAMP**.
                                                                                                                                                                                                                             
  The same logical column landed in parquet as two incompatible types depending on partitioning config, so reconfiguring `HIVE_COLUMNS` (or running two sinks with different configs against one table) silently fractured 
  the table.                                                                                                                                                                                                                 
                                                                                                                                                                                                                             
  This violates the contract on `ts_ms` in particular: `ts_ms` is sink-injected from `item.timestamp` (Kafka message timestamp, always int64 ms). Its dtype is not a partitioning concern.                                   
                                                                                                                                                                                                                             
  Fix: derive year/month/day/hour from a local `datetime64` series; never write back into `df[timestamp_column]`.
                                                                                                                                                                                                                             
  ## Test plan                                                                                                                                                                                                               
   
  - [x] `tests/test_quixstreams/test_sinks/test_core/test_quix_ts_datalake_sink.py` — 58/58 pass.                                                                                                                            
  - [x] New `test_manifest_registers_null_partition_as_literal_sentinel` asserts the catalog payload carries `'__None__'` literal and the on-disk path matches.
  - [x] New `test_add_timestamp_columns_does_not_mutate_timestamp_column_dtype` asserts `ts_ms` keeps `int64` after extraction with `HIVE_COLUMNS=[year, month, day, hour]`.                                                 
  - [x] Adjacent stale assertion in `test_write_all_null_partition_values_still_writes` updated from `machine=None` to `machine=__None__` (was already expected on-disk, just an out-of-date string check).
                                                                                                                                                                                                                             
  ## Coordination with the lake                                                                                                                                                                                              
                                                                                                                                                                                                                             
  The lake side (`Quix.DataLake.Timeseries`) has been refactored to literal-passthrough handling of partition values and to expect `ts_ms` as int64. With both changes in place, the sink→catalog→tree→query path agrees     
  end-to-end and `WHERE test_run = '__None__'` returns the rows of the NULL bucket without any sentinel translation in between.                                                                                              
                                                                                                                                                                                                                             
  ## Migration note for users                               
                                                                                                                                                                                                                             
  - Files already on disk written by the old sink (catalog rows say `NULL`, on-disk path says `__None__`) appear as two distinct buckets in the partition tree until the table is refreshed against the new lake. A single 
  `/refresh` re-scans S3, finds the literal `__None__`, and collapses them into one bucket.                                                                                                                                  
  - Files written before the timestamp fix may have `ts_ms` as `TIMESTAMP` (if `HIVE_COLUMNS` included a time column at write time) and `BIGINT` otherwise. The lake's compaction already groups by type and unifies on read;
   running compaction after this lands will produce a uniformly-typed `ts_ms` column.

### DIFF
--- a/quixstreams/sinks/core/quix_ts_datalake_sink.py
+++ b/quixstreams/sinks/core/quix_ts_datalake_sink.py
@@ -497,14 +497,20 @@ class QuixTSDataLakeSink(BatchingSink):
         Add timestamp-based columns (year/month/day/hour) for time-based partitioning.
 
         This method extracts time components from the timestamp column and adds them
-        as separate columns that can be used for Hive partitioning.
+        as separate columns that can be used for Hive partitioning. The source
+        ``timestamp_column`` is **not** mutated — derivation happens on a local
+        datetime64 series. Preserving its dtype is part of the sink's contract
+        with readers (in particular, ``ts_ms`` is the system-injected Kafka
+        timestamp and must always land in parquet as int64 ms, regardless of
+        whether time-based hive partitioning is configured).
         """
-        # Convert to datetime if needed (handles numeric timestamps)
-        if not pd.api.types.is_datetime64_any_dtype(df[self.timestamp_column]):
+        # Build a datetime64 view of the timestamp column to extract from.
+        # Never write this back into ``df`` — that would change the dtype of
+        # the source column in the parquet output.
+        timestamp_col = df[self.timestamp_column]
+        if not pd.api.types.is_datetime64_any_dtype(timestamp_col):
             sample_value = float(
-                df[self.timestamp_column].iloc[0]
-                if not df[self.timestamp_column].empty
-                else 0
+                timestamp_col.iloc[0] if not timestamp_col.empty else 0
             )
 
             # Auto-detect timestamp unit by inspecting the magnitude of the value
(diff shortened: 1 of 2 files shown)


LudvikQuix (approved):
review OK

PR #1115, opened 2026-06-04
Recover state when source offset expires

### LINKED ISSUE #1114: Stateful recovery can get stuck when committed source offset is no longer retained
**Tell us about the bug**
A stateful Quix Streams application can get stuck when its committed source-topic offset is no longer retained by Kafka. librdkafka resets the fetch position according to auto.offset.reset, but state recovery still uses the stale committed offset as the changelog recovery boundary.

The app should not silently stay alive without processing or advancing commits. It should either recover from a valid retained source offset with clear destructive-recovery logging, or fail fast with an actionable error when automatic recovery is disabled.

Observed on quixstreams==3.23.6.

**Workaround?**
Reset the consumer group offset to a retained source offset and clear/rebuild the local state directory, or start with a new consumer group and state directory.

This affects stateful apps using changelog recovery, for example apps using join_asof or other stateful operations. The production case had committed offset 50456 while the broker low watermark was 75033, so the original committed source data was no longer recoverable.

peter-quix:
Features
- Add destructive state recovery for stateful apps whose committed source offset is below the broker low watermark, with configurable recovery reset policy `"earliest"`, `"latest"`, or `"match"`.

Docs
- Document automatic source-offset-out-of-range recovery, the fail-fast flag, and the new `state_recovery_offset_reset` behavior.


peter-quix:
One thing I want to call out for review is the default for `state_recovery_offset_reset`.

I set the default to `"earliest"` because, after we delete local state, rebuilding from the broker low watermark gives the most coherent remaining state from the data Kafka still retains. If we defaulted to matching `auto_offset_reset="latest"`, recovery would delete state and then skip all currently retained source records too, leaving the app with a fresh/empty state from the topic high watermark.

That said, this does slightly muddy the Kafka-like expectation that `auto.offset.reset=latest` means "go to the end" whenever the committed offset is out of range. The new `"match"` option exists for users who want that behavior, but we are intentionally not making it the default. I think that is the safer stateful-processing default, but I’d like reviewers to weigh in on whether `"earliest"` or `"match"` is the right default.

LudvikQuix (commented):
review ok.
Let test together with TTL

LudvikQuix (approved):
review OK

PR #1117, opened 2026-06-08
Prevent buffered consumers from stalling after restart

### LINKED ISSUE #1116: Stateful join app can stall after restart when one time-aligned input is already caught up
## Tell us about the bug

A stateful Quix Streams application using a join/time-aligned dataframe can get stuck after restart or reassignment when one subscribed non-changelog topic is already at its committed end offset.

In the observed case, the app consumed two non-changelog topics:

- `source-events`
- `lookup-state`

The application was stuck with lag on `source-events`:

```text
source-events current offset: 462
source-events log end:        909
lag:                          447
```

The other joined topic was already fully caught up:

```text
lookup-state current offset: 3
lookup-state log end:        3
lag:                         0
```

Kafka could serve the stuck records, and librdkafka did return the `source-events` backlog to Quix Streams. However, the time-alignment buffer did not release those records for processing because the already-caught-up joined topic had not produced any messages in the current process.

The buffer determines whether an empty topic partition is idle using its buffer-local max offset and the Kafka high watermark:

```text
max_seen_offset + 1 >= high_watermark
```

After restart, the buffer-local max offset for the already-caught-up topic starts at `-1`. If that partition is committed at offset `3` and high watermark is also `3`, the buffer still sees:

```text
-1 + 1 >= 3  -> false
```

So the partition is treated as not idle, even though the consumer group is caught up. This blocks release of records from the other joined topic.

Runtime diagnostics from the production case showed:

```text
consume returned count=447 offsets=[
  source-events 462,
  source-events 463,
  ...
]

assignment=[
  source-events position 909 high 909,
  lookup-state position -1001 high 3
]
```

The app stayed alive and kept polling, but no further records were processed or committed until the buffer was seeded with committed offsets.

## What did you expect to see?

The app should continue processing records from the lagging topic when another time-aligned topic partition is already caught up.

An assigned topic partition that has a committed offset equal to its high watermark should be treated as idle by the time-alignment buffer, even if the current process has not observed any messages from that partition.

The app should not silently stay alive with lag while withholding fetched records in the internal buffer.

## What version of the library are you using?

Observed on:

```text
quixstreams==3.23.6
confluent-kafka==2.11.1
librdkafka==2.11.1
```

## Workaround?

Temporary operational workarounds:

- Start with a fresh consumer group and state directory.
- Reset the affected joined topic offsets so every time-aligned input has new messages to consume after restart.
- Patch Quix Streams locally so the time-alignment buffer is initialized from committed offsets during assignment.

In the production case, mounting a patched Quix Streams module into the running deployment unblocked the app. The app processed the entire backlog:

```text
source-events current offset: 909
source-events log end:        909
lag:                          0

output-topic-a: 198 -> 423
output-topic-b: 264 -> 486
```

The output increase was exactly `447` records, matching the original lag.

## Anything else we should know?

This affects stateful/time-aligned apps, for example apps using `join_asof`, when multiple non-changelog topics share the same partition number and one of those topics is already caught up at startup/reassignment.

peter-quix:
Fixes
- Seed the time-alignment buffer from effective consumer positions, explicit assignment offsets, or committed offsets so empty co-partitioned topics can be recognised as idle after restart
- Preserve custom assignment offsets while still falling back to committed offsets when live positions are unknown

Tests
- Cover buffered restart recovery when one topic has lag and another topic is already committed at its high watermark
- Cover buffered replay from a custom assignment offset below the committed group offset

### DIFF
--- a/quixstreams/internal_consumer/buffering.py
+++ b/quixstreams/internal_consumer/buffering.py
@@ -57,10 +57,7 @@ class PartitionBuffer:
         """
         Set the current consumer position for this topic partition.
 
-        The position is the next offset to be returned by the Kafka consumer. When a
-        process starts from a committed offset, the buffer has not observed the
-        previously consumed messages, so its max offset must be initialized from
-        the position to determine idleness correctly.
+        The position is the next offset to be returned by the Kafka consumer.
 
         :param offset: the current consumer position.
         """
--- a/quixstreams/internal_consumer/buffering.py
+++ b/quixstreams/internal_consumer/buffering.py
@@ -53,6 +53,20 @@ class PartitionBuffer:
         """
         self._high_watermark = offset
 
+    def set_position(self, offset: int):
+        """
+        Set the current consumer position for this topic partition.
+
+        The position is the next offset to be returned by the Kafka consumer. When a
+        process starts from a committed offset, the buffer has not observed the
+        previously consumed messages, so its max offset must be initialized from
+        the position to determine idleness correctly.
+
+        :param offset: the current consumer position.
+        """
+        if offset > 0:
+            self._max_offset = max(self._max_offset, offset - 1)
+
     def idleness(self) -> Idleness:
         """
         Check if the partition is idle or has more data to be consumed from the broker.
@@ -202,6 +216,17 @@ class PartitionBufferGroup:
         for topic, watermark in offsets.items():
             buffer = self._partition_buffers[topic]
             buffer.set_high_watermark(watermark)
+
+    def set_positions(self, offsets: dict[str, int]):
+        """
+        Set consumer positions for assigned partitions.
+
+        :param offsets: a mapping of {<topic>: <offset>} with the current consumer
+            positions for this group.
+        """
+        for topic, offset in offsets.items():
+            buffer = self._partition_buffers[topic]
+            buffer.set_position(offset)
 
     def append(self, message: SuccessfulConfluentKafkaMessageProto):
         """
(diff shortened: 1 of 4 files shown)


peter-quix:
# Time-Alignment Buffer Restart Stall

## Problem

Buffered consumption aligns messages from co-partitioned input topics by timestamp. For a partition group with multiple topics, the buffer only releases a message when every other topic buffer is either non-empty or known to be idle.

The stall happens after restart when one topic still has lag, while another topic is already committed at its high watermark but has not produced any messages in the new process.

```text
Kafka after restart

topic A: events-valid
  committed/current: 462
  high watermark:     909
  records available:  462..908

topic B: state-period_windows_by_stream
  committed/current: 3
  high watermark:    3
  records available: none
```

Before this fix, the time-alignment buffer only knew about offsets it had observed in the current process:

```text
Buffer-local state before fix

topic A:
  observed max offset: 462
  high watermark:      909
  buffer:              [message at 462]

topic B:
  observed max offset: -1
  high watermark:      3
  buffer:              []

idleness(topic B) = observed max offset + 1 >= high watermark
                  = -1 + 1 >= 3
                  = false
```

Because topic B looked active instead of idle, the buffer withheld topic A's message even though Kafka could serve it.

```text
             same partition group

topic A buffer: [A462, A463, ...]  ready
topic B buffer: []                 incorrectly treated as active
                         |
                         v
                 buffer.pop() returns None
```

## Fix

Seed the buffer with effective consumer positions in addition to high watermarks.

The position is the next offset Kafka will return. If a topic partition is positioned at `3`, then offsets up to `2` have already been consumed or skipped for this process, so the buffer can initialize its local max offset to `2`.

```text
Buffer-local state after fix

topic B:
  effective position:  3
  observed max offset: 2
  high watermark:      3

idleness(topic B) = observed max offset + 1 >= high watermark
                  = 2 + 1 >= 3
                  = true
```

Now topic B is correctly recognized as idle, so topic A can be released:

```text
             same partition group

topic A buffer: [A462, A463, ...]  ready
topic B buffer: []                 idle
                         |
                         v
                 buffer.pop() returns A462
```

## Position Source Priority

Assignment-time seeding uses the most accurate position signal available:

```text
1. live consumer position after on_assign
2. explicit offsets set by the on_assign callback
3. committed offsets from the consumer group
```

The order matters because `on_assign` is the supported place for custom offsets. If an assignment callback replays a partition from an earlier offset, seeding from the old committed offset would make the buffer think replayed records had already been observed.

```text
Committed offset before assignment: 3
on_assign chooses replay offset:    0

wrong seed:
  max offset = committed - 1 = 2
  first replayed message offset = 0
  append fails as duplicate/old offset

correct seed:
  max offset = assignment offset - 1 = -1
  first replayed message offset = 0
  append succeeds
```

LudvikQuix (approved):
Review ok

PR #1123, opened 2026-06-16
fix: scope TimestampedStore expiry to the prefix namespace

### LINKED ISSUE #1122: State expiry in TimestampedStore can delete entries belonging to unrelated keys
**Tell us about the bug**

`TimestampedStore` stores each value under a RocksDB key of the form `<prefix> + SEPARATOR + <timestamp>`, where `prefix` is the message key and `SEPARATOR` is `b"|"` (`0x7c`).

While reviewing the timestamped-state expiry logic, I noticed that `TimestampedPartitionTransaction._expire()` builds the deletion range with the **bare prefix** as the lower bound (`quixstreams/state/rocksdb/timestamped.py`):

```python
key = self._serialize_key(min_eligible_timestamp, prefix)   # b"key|<ts>"  (upper bound)
stored = self._partition.iter_items(lower_bound=prefix,      # b"key"       (lower bound)
                                    upper_bound=key)
```

Because the lower bound omits the separator, the scanned range `[b"key", b"key|<ts>")` also covers keys belonging to **other prefixes** that begin with this prefix's bytes. Any sibling prefix whose next byte is `< 0x7c` — i.e. digits, letters, `-`, `.`, `_` — sorts inside the range and gets deleted. For example `b"key2|<ts>"` is `b"key" + b"2"(0x32) + b"|" + ...`, and since `0x32 < 0x7c` it sorts before `b"key|<ts>"`, so expiring `key` silently deletes `key2`'s data.

This affects very common key shapes (`key`/`key2`, `user_1`/`user_12`, `order`/`order_99`, numeric IDs), and the interval-join path, which uses message keys as state prefixes directly.

**Minimal reproduction:**

```python
import tempfile
from quixstreams.state.rocksdb.timestamped import TimestampedStore

store = TimestampedStore(stream_id="t", name="default",
                         base_dir=tempfile.mkdtemp(),
                         grace_ms=5, keep_duplicates=False)
store.assign_partition(0)

# write under prefix b"key2" at an old-but-recent timestamp
with store.start_partition_transaction(0) as tx:
    tx.set_for_timestamp(timestamp=3, value="key2val", prefix=b"key2")

# write under prefix b"key" at a high timestamp -> triggers expiry for b"key"
with store.start_partition_transaction(0) as tx:
    tx.set_for_timestamp(timestamp=100, value="keyval", prefix=b"key")

# b"key2" was never eligible for expiry under its own retention
with store.start_partition_transaction(0) as tx:
    print(tx.get_latest(timestamp=3, prefix=b"key2"))   # -> None   (expected: "key2val")
```

Expiring entries for one key must only delete that key's own data. `get_latest(prefix=b"key2")` should return `"key2val"`; instead it returns `None` because `key2`'s entry was swept up by `key`'s expiry.

3.24.0 (also present on `main`, commit `e96e28b`).

**Workaround?**

Avoid state keys where one key's byte sequence is a prefix of another (e.g. pad/hash keys to fixed length). Not generally practical since keys come from the data.

This was found by code review, not a production incident. Root cause is the unscoped lower bound; the fix is to scope it to the prefix's namespace the way every other range query in the file does:

```python
stored = self._partition.iter_items(lower_bound=prefix + SEPARATOR, upper_bound=key)
```

With `lower_bound=b"key|"`, sibling prefixes sort outside the range while all genuine `b"key|<ts>"` rows remain covered. I verified this locally: the repro above passes with the change and the full `tests/test_quixstreams/test_state/` suite stays green. The in-memory-cache half of `_expire` is already correctly prefix-scoped; only the RocksDB `iter_items` range is affected.

I have the one-line fix and a regression test ready and will open a PR shortly.

Vedant-Agarwal:
## What

`TimestampedPartitionTransaction._expire()` built its RocksDB deletion range with the **bare prefix** as the lower bound:

```python
stored = self._partition.iter_items(lower_bound=prefix, upper_bound=key)
```

State keys are stored as `<prefix> + SEPARATOR + <timestamp>` (`SEPARATOR = b"|"`, `0x7c`). Because the lower bound omits the separator, the range `[b"key", b"key|<ts>")` spills into **other prefixes** whose bytes start with this prefix and whose next byte is `< 0x7c` (digits, letters, `-`, `.`, `_`). For example `b"key2|<ts>"` sorts before `b"key|<ts>"`, so expiring `key` silently deletes `key2`'s data.

This hits common key shapes (`key`/`key2`, `user_1`/`user_12`, numeric IDs) and the interval-join path, which uses message keys as state prefixes.

## Fix

Scope the lower bound to the prefix's namespace, matching every other range query in the file:

```python
lower_bound = self._serialize_key(b"", prefix)   # prefix + SEPARATOR
stored = self._partition.iter_items(lower_bound=lower_bound, upper_bound=key)
```

Sibling prefixes now sort outside the range while all genuine `<prefix>|<ts>` rows remain covered. The in-memory-cache half of `_expire` was already correctly prefix-scoped; only the `iter_items` range needed fixing.

## Tests

Added `test_expire_does_not_affect_keys_sharing_byte_prefix` (byte-prefixed `key`/`key2` pair — the case the existing `test_expire_multiple_keys` misses since it uses equal-length, non-nested prefixes). It fails without this change and passes with it. The full `tests/test_quixstreams/test_state/` suite passes (323), ruff clean.

Fixes #1122

### DIFF
--- a/quixstreams/state/rocksdb/timestamped.py
+++ b/quixstreams/state/rocksdb/timestamped.py
@@ -203,7 +203,14 @@ class TimestampedPartitionTransaction(RocksDBPartitionTransaction):
                 if cached_key < key:
                     keys_to_delete.append((cached_key, prefix))
 
-            stored = self._partition.iter_items(lower_bound=prefix, upper_bound=key)
+            # Scope the lower bound to this prefix's namespace by appending the
+            # SEPARATOR. Without it, the bare prefix lets the range spill into
+            # other prefixes that share these bytes (e.g. b"key" vs b"key2"),
+            # deleting entries that belong to unrelated keys.
+            lower_bound = self._serialize_key(b"", prefix)
+            stored = self._partition.iter_items(
+                lower_bound=lower_bound, upper_bound=key
+            )
             for stored_key, _ in stored:
                 keys_to_delete.append((stored_key, prefix))
 
--- a/tests/test_quixstreams/test_state/test_rocksdb/test_timestamped.py
+++ b/tests/test_quixstreams/test_state/test_rocksdb/test_timestamped.py
@@ -175,6 +175,24 @@ class TestTimestampedPartitionTransaction:
             assert tx.get_latest(timestamp=30, prefix=b"key1") == "112"
             assert tx.get_latest(timestamp=30, prefix=b"key2") == "212"
 
+    def test_expire_does_not_affect_keys_sharing_byte_prefix(
+        self, transaction: TimestampedPartitionTransaction
+    ):
+        # `b"key2"` shares its leading bytes with `b"key"`. Since the SEPARATOR
+        # (b"|", 0x7c) sorts after digits/letters, `b"key2|..."` orders before
+        # `b"key|..."`, so expiring `b"key"` must not reach into `b"key2"`.
+        with transaction(grace_ms=5) as tx:
+            tx.set_for_timestamp(timestamp=3, value="key2val", prefix=b"key2")
+
+        # Writing `b"key"` at a high timestamp advances its expiration watermark
+        # and triggers a sweep. `b"key2"` was never eligible under its own grace.
+        with transaction(grace_ms=5) as tx:
+            tx.set_for_timestamp(timestamp=100, value="keyval", prefix=b"key")
+
+        with transaction(grace_ms=5) as tx:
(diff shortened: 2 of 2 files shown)


LudvikQuix (approved):
review OK

PR #1127, opened 2026-07-01
Feature/sc 73191/udpate state behavior of dedup feature to

LudvikQuix:
The motivation (populated legacy store couldn't adopt TTL without a destructive state reset, impossible in Quix Cloud), the migration engine (legacy_records_ttl, chunked/bounded, one-time, durable), recovery correctness (__ttl_stamped__ header, OP-4 cold-restore completion, memory-backend parity), migration-only semantics, the two-tier logging + QUIXSTREAMS_STATE_LOG_LEVEL env var, the docs, backward-compat guarantees, a new-config table, and the test inventory


LudvikQuix:
will be delivered with [PR1129](https://github.com/quixio/quix-streams/pull/1129)

PR #1128, opened 2026-07-01
fix: TTL sweep no longer deletes a key refreshed in the same flush

Vedant-Agarwal:
## What

The TTL sweep (`_run_sweep`, called from `RocksDBStorePartition.write()`) can **silently delete a key that was just refreshed with a fresh TTL in the same flush** — data loss.

## Why

`write()` stages the transaction's writes into the `WriteBatch` first, then runs `_run_sweep`. The sweep reads the main value via `main_cf.get(user_key)`, which sees **committed disk state, not the uncommitted batch**. So when a key with an expired stamp is refreshed (re-`set` with a new TTL) in the same flush:

1. the batch stages `put(user_key, fresh_value)`;
2. the sweep reads the **stale** committed value, finds its old stamp still matches the old index entry (`main_expires_at == idx_expires_at`), and stages `batch.delete(user_key)`;
3. RocksDB applies batch ops in insertion order → the delete lands **after** the put and wins → the fresh value is removed from disk.

This hits the most natural TTL pattern: `state.set(k, v, ttl=...)` to keep a key alive.

### Reproduction
t=1000 `set(k, v1, ttl=100ms)` (expires 1100); t=2000 `set(k, v2, ttl=100ms)` (expires 2100, still alive since high-water=2000). After the second flush, `get(k)` returns `None` — the freshly-written `v2` was swept. (Added as a regression test; it fails on `main`, passes with this fix.)

## Fix

Pass the set of default-CF keys re-written in this flush into `_run_sweep` and **skip deleting any of them** — drop only the stale index pointer (the fresh stamp's own index entry governs its future expiry). Default behaviour for non-refreshed keys is unchanged.

Related to the TTL feature added in #1107.

### DIFF
--- a/quixstreams/state/rocksdb/partition.py
+++ b/quixstreams/state/rocksdb/partition.py
@@ -270,6 +270,11 @@ class RocksDBStorePartition(StorePartition):
         # For unflipped partitions this commits the cache as-is — exactly the
         # v3.23.6 behavior. For flipped partitions the transaction layer has
         # already stamped values and emitted index-CF writes into the cache.
+        # Keys re-written into the default CF in this same flush. The TTL sweep
+        # reads committed disk state (not this uncommitted batch), so it must not
+        # delete a key the batch just refreshed — otherwise the stale-read delete
+        # clobbers the fresh write.
+        staged_default_keys: set[bytes] = set()
         for cf_name in column_families:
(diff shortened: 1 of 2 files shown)


peter-quix:
Thank you, super-seeded by #1129

PR #1132, opened 2026-07-08
Prevent RocksDB lock contention during Kafka rebalances

### LINKED ISSUE #1131: Stateful app livelocks on shared state volumes: sequential RocksDB revoke blows `max.poll.interval.ms`, triggering eviction and a lock-contention rebalance loop
## Tell us about the bug

A stateful Quix Streams application backed by RocksDB can get stuck in a continuous rebalance loop when multiple replicas share a state volume and a new instance joins the consumer group.

In the observed case, ran multiple replicas against a topic with **64 partitions** and **tens of millions of unique keys** in state, with replicas intentionally **sharing a state volume** (warm state handover to avoid hours of changelog replay). When a second instance joined, the two instances fought over RocksDB file locks for an extended period instead of performing the expected clean handover, logging in a tight loop:

```
Failed to open rocksdb partition, cannot acquire a lock. Retrying in 5.0sec.
```

Each store partition is a separate RocksDB database that holds an OS-level `flock` on its `LOCK` file for the lifetime of the owning process. Kafka's group protocol guarantees group-membership exclusivity, but it does **not** fence external resources like these lock files. The whole revoke sequence runs single-threaded and strictly sequential, *inside the rebalance callback* between two `poll()` calls, so it all counts against the `max.poll.interval.ms` clock (default 300s):

1. Force checkpoint commit — flush sinks, `prepare()` per partition, `producer.flush()`, synchronous offset commit, then `transaction.flush()` per partition **sequentially**.
2. Close each revoked store partition in a plain loop; each `db.close()` waits for that DB's background flush/compaction to wind down — seconds per partition on large DBs. The lock for partition *k* is released only after every partition before it has fully closed.

With 32–64 large partitions this compounds into minutes, and the failure cascades:

```
Instance A (revoking)                 Instance B (assigned)
---------------------                 ---------------------
close p0  (lock released)             open p0  -> lock held by A -> retry
close p1  ...                         open p0 ... (chases A's close pointer)
   |  cumulative close time           open runs inside B's _on_assign with a
   |  exceeds max.poll.interval.ms     blocking retry loop (open_max_retries x
   v                                   open_retry_backoff, sequential)
librdkafka evicts A mid-revoke  --->  B is assigned ALL partitions while A is
coordinator completes rebalance       still closing them front-to-back
```

Because A is evicted **mid-revocation**, it never finishes releasing its locks, yet the coordinator has already reassigned the partitions to B. B blocks acquiring each lock, blows *its own* `max.poll.interval.ms` (or exhausts retries and crashes), rejoins, and the cycle repeats. Externally this is a continuous multi-minute lock fight with no forward progress and no single buggy step — every component behaves as written.

Secondary defect: the open-retry `time.sleep` loop ignores the application stop flag, so an instance stuck waiting on a lock cannot be shut down cleanly.

## What did you expect to see?

- The revoke/handover sequence should complete fast enough to stay within `max.poll.interval.ms`, even with many large store partitions on a shared/slow volume.
- RocksDB state-directory locks should be released promptly on revoke so a reassigned partition's new owner can acquire the lock without an extended contention loop.
- Lock contention during handover should be absorbed without any member being evicted; a rebalance should converge to a stable assignment instead of livelocking through repeated evict → crash → rebalance cycles.
- An instance blocked waiting on a lock should still be promptly killable (honour the application stop signal).

## What version of the library are you using?

- `quixstreams` 3.23.x with RocksDB state (`rocksdict`).
- Environment: state directory on a shared/network-attached volume; consumer group with multiple stateful partitions per member (observed with a 64-partition topic and tens of millions of keys).
- Forced `range` (eager) assignment strategy (`quixstreams/kafka/consumer.py`).

## Workaround?

Operational mitigations (no code change):

- Size instance B's retry budget to outlast A's worst-case handover: `stores × partitions × open_max_retries × open_retry_backoff` must exceed A's checkpoint+close time while staying under B's `max.poll.interval.ms` — raise both together via `RocksDBOptions` and consumer config.
- Lower `commit_interval` so less data accumulates in the forced checkpoint at revoke time.
- Put bounded timeouts on all sink/external calls so `commit_checkpoint` cannot hang.
- For deploy-driven rebalances, stop the old replica before starting the new one (`Recreate` strategy) so graceful shutdown runs the revoke callback and releases all locks before the new instance joins.

## Anything else we should know?

This affects any stateful RocksDB app where revoke work scales with partition count and per-partition close is slow — shared/network volumes make it acute.

peter-quix:
Fixes
- Keep RocksDB rebalance handovers within Kafka poll deadlines by cancelling background work, skipping safe local state flushes, sharing a bounded revoke budget, and interrupting contended opens during shutdown.
- Preserve checkpoint correctness when revoke-time flushes or EOS transaction operations fail by aborting or purging pending writes without committing offsets.
- Resume recovery-paused data partitions when they are reassigned without a stateful partition, preventing consumers from remaining paused indefinitely.

Docs
- Document bounded checkpoint commits during partition revocation, including fast state handover and at-least-once failure semantics.

Closes #1131


LudvikQuix:
please fix codes to pass CI

LudvikQuix (approved):
review OK

Issue #1139, opened 2026-08-11
Bump confluent-kafka dependency to 2.13.2


ffailla:
## Description

Update the Kafka client dependency used by Quix Streams to confluent-kafka==2.13.2 in requirements.txt.

This is needed for compatibility with downstream services that require confluent-kafka 2.13.2. Aligning Quix Streams with that version will avoid dependency resolver conflicts and keep builds reproducible.

## Requested Change

Update the Quix Streams dependency declarations for confluent-kafka and confluent-kafka[avro] to support version 2.13.2.

PR #1140, opened 2026-08-11
updated confluent-kafka version to be <=2.13.2

ffailla:
## Description

Update the Kafka client dependency used by Quix Streams to confluent-kafka<=2.13.2 in requirements.txt.

This is needed for compatibility with downstream services that require confluent-kafka 2.13.2. Aligning Quix Streams with that version will avoid dependency resolver conflicts and keep builds reproducible.

## Requested Change
Update the Quix Streams dependency declarations for confluent-kafka and confluent-kafka[avro] to support version 2.13.2.

## Issue

https://github.com/quixio/quix-streams/issues/1139

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-confluent-kafka[avro,json,protobuf,schemaregistry]>=2.8.2,<2.12
+confluent-kafka[avro,json,protobuf,schemaregistry]>=2.8.2,<=2.13.2
 rocksdict>=0.3,<0.4
 typing_extensions>=4.8
 orjson>=3.9,<4


snarayans (approved):

peter-quix:
Hi,

Thanks for the submission, but this is something we'll need to test a bit more. There are two main issues that need handling:
1. Consumer batching regression in 2.13.x. Quix’s normal 1-second buffered polling can return after the first 200 ms chunk, producing much smaller batches and affecting time-aligned dataframe processing. Existing 0.1-second tests miss this path. Upstream source (https://github.com/confluentinc/confluent-kafka-python/blob/v2.13.2/src/confluent_kafka/src/Consumer.c#L1154-L1201), issue #2205 (https://github.com/confluentinc/confluent-kafka-python/issues/2205).

2. Producer backpressure latency in 2.13.0–2.13.2. When the queue is full, Quix polls for callbacks with a default 30-second timeout. The 2.13.x implementation may continue waiting after a callback has already freed capacity, causing an avoidable stall. quixstreams/kafka/producer.py:205, upstream source (https://github.com/confluentinc/confluent-kafka-python/blob/v2.13.2/src/confluent_kafka/src/Producer.c#L363-L405), issue #2215 (https://github.com/confluentinc/confluent-kafka-python/issues/2215).

Will need to investigate these and determine if it is actually safe to bump or wait for fixes (or bump to a different version)

LudvikQuix:
Checked the two upstream issues today. Both are still open and no released version fixes them, so 2.14.x and 2.15.x have the same problem as 2.13.x.

On confluentinc/confluent-kafka-python#2205, Confluent said on 26 Aug the fix would go into the next release, but 2.15.1 (10 Sep) doesn't mention it. confluentinc/confluent-kafka-python#2215 has had no maintainer response since March, and the PR someone offered in June went unanswered.

Keeping the cap at <2.13 until both are closed.

@peter-quix can you add a regression test for the 1s buffered-poll path, and one for the producer poll timeout? With those in the suite the next bump either goes green or tells us exactly what's still broken, instead of us re-reading upstream issues every time.

PR #1143, opened 2026-09-01
feat(sinks): virtual partitions, file statistics and sort column for QuixLake sink

SteveRosam:
## What

Three query-performance features for `QuixTSDataLakeSink`, plus the docs and tests for them.

**1. Virtual partition columns.** A `~`-prefixed entry in `hive_columns` becomes a *virtual* partition: it appears in the partition tree and is filterable, but gets no `key=value/` folder and does not split files. The column stays in the Parquet data so queries can still filter rows by it. Use it when a value should be navigable without fragmenting storage — high-cardinality identifiers such as a driver, device, or session, where a physical partition would produce one small file per value per batch.

Each data file gets a virtual-index sidecar in a `.vidx/` subfolder of its own Hive partition folder, holding one row per distinct virtual tuple with the file's physical partition values added as constant columns. A reader gets full partition tuples from the content alone, with no `hive_partitioning` needed. A `.vidx` folder carries no `key=value`, so partition discovery skips it and it never joins the data set.

**2. Min/max file statistics (zone maps).** Per-file `column_stats` are computed from the in-memory Arrow batch before serialization and sent to the REST Catalog with each file, so the query layer can skip files whose range cannot satisfy a `WHERE` or `ORDER BY`. The Parquet footer is never re-read from storage. `stats_columns` restricts the tracked set for very wide tables.

Numeric bounds are widened outward to the nearest representable float. That guarantees the stored range is a *superset* of the real one, so float rounding of large integers (nanosecond epochs beyond 2^53) can only cost some pruning and can never wrongly skip a file holding matching rows.

**3. Sort column.** `sort_column` is recorded as `properties.sort_column` for lakehouse compaction to order files by, so `ORDER BY` and time-range queries can skip files and stream rather than sorting the whole table. Falls back to `timestamp_column` when unset. Table metadata only — the sink does not reorder rows within a file.

## Testing

81 tests in `test_quix_ts_datalake_sink.py`, all passing. Sink coverage **96%** (360 statements, 16 missed).

The last three commits close two gaps where a feature's own guarantee had no test behind it:

- **The min-side float widening had never executed.** The existing test used `2**53+1`, which rounds *down*, so only `_safe_float_max` widened. Now parametrised with `2**53+3`, which rounds *up*, plus a strict-bracket assertion — without it the test would still pass with a widening loop deleted.
- **Both sidecar failure handlers had no test.** `_write_virtual_sidecar` promises *"Never raises (the index is a hint, not the data)"* and `_await_sidecar_uploads` promises a failed sidecar never blocks the data. There is now one test per handler asserting the data file still lands and the manifest call still happens. Since `write()` retries a batch 3x, asserting *exactly one* data upload also proves no retry was triggered. Verified load-bearing: narrowing both handlers to `except ValueError` fails both tests.

## Pre-existing test failures — not from this branch

13 tests fail on this branch. **The identical 13 fail on `main`** (`13 failed, 12 passed` on both), so they are pre-existing:

- 8 in `test_utils/test_printing.py` — a `rich` version difference in the default table box style (expects heavy `┏━━┓`, gets light `┌───┐`)
- 5 in `test_sources/test_base/test_manager.py` — the `-9` / `cleanup` parametrisations depend on POSIX process-signal semantics; run on Windows

Tracked as [sc-74770](https://app.shortcut.com/quix/story/74770) and [sc-74771](https://app.shortcut.com/quix/story/74771). Both are expected to be green on Linux CI, which would confirm them as local environment drift.

The full suite was not run to completion locally: it reached 4% in 27 minutes against 1897 tests (Kafka testcontainers on Windows Docker, no `pytest-xdist`), projecting to ~10 hours. Docker was verified up and healthy first — 0 `DockerException`, 0 failures across the tests that did run. **CI is the full-suite verification here**, and it runs on the platform sc-74771 actually depends on. This branch changes 3 files, so the Kafka half cannot be affected by the diff.

## Known follow-ups

Filed rather than fixed here, to keep this PR's outstanding work test-only. Both are tracked under the epic below (stories 74765 and 74766) — this PR does **not** implement either:

- **`~` on a time column (`~hour`) is a silent no-op.** `_ts_hive_columns` derives time parts for physical columns only, and making it work would mean the sink inventing a column the records never carried. Documented as a caveat here; a `ValueError` is proposed but would be new behaviour.
- **A virtual column absent from the payload yields an empty tree level, silently.** Benign (no data loss) but worth a once-per-column warning.

Epic: [QuixLake Sink](https://app.shortcut.com/quix/epic/74763)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/connectors/sinks/quix-ts-datalake-sink.md
+++ b/docs/connectors/sinks/quix-ts-datalake-sink.md
@@ -59,6 +59,108 @@ Records must be dictionaries. If your values are not dicts, convert them before 
 
 Blob credentials are read automatically from the `Quix__BlobStorage__Connection__Json` environment variable when running on Quix Cloud; for local runs, the filesystem is inferred from the `quixportal` configuration.
 
+## Partition Columns
+
+`hive_columns` accepts two kinds of entry.
+
+A plain entry is a **physical** partition. It becomes a real `key=value/` folder in storage, it groups the batch (each distinct value gets its own file), and the column is dropped from the Parquet data because its value is already in the path.
+
+An entry prefixed with `~` is a **virtual** partition. It appears in the partition tree and is filterable, but it gets no folder and does not split files — one file keeps every value of it. The column stays in the Parquet data so queries can still filter rows by it.
+
+```python
+sink = QuixTSDataLakeSink(
+    s3_prefix="data-lake/time-series",
+    table_name="telemetry",
+    hive_columns=["year", "month", "~driver"],
+    timestamp_column="ts_ms",
+)
+```
+
+That folders by `year=YYYY/month=MM/` and exposes `driver` as a third, virtual level:
+
+```
+data-lake/time-series/telemetry/year=2024/month=01/data_<uuid>.parquet        # every driver, one file
+data-lake/time-series/telemetry/year=2024/month=01/.vidx/data_<uuid>.parquet  # its virtual index
+```
+
+Use a virtual column when you want a value to be navigable and filterable but do not want it to fragment storage — high-cardinality identifiers such as a driver, device, or session are the typical case. A physical partition on the same column would produce one small file per value per batch.
+
+### The `.vidx` sidecar index
+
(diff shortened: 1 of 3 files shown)


LudvikQuix (approved):
review OK

PR #1144, opened 2026-09-04
Fix TTL write with Kafka NO_TIMESTAMP being stored expired in 1970 (#1141)

### LINKED ISSUE #1141: TTL write with Kafka NO_TIMESTAMP can be stored expired in 1970 and swept immediately
## Summary

A `state.set(..., ttl=...)` write whose record carries Kafka's `NO_TIMESTAMP`
(`-1`) can be stored with an expiry in **1970**, so the record is already expired
the moment it lands and the next TTL sweep deletes it.

It needs a batch that contains **both** an un-timestamped and a normally
timestamped `ttl=` write, on a store that has not yet flipped into TTL mode.

## Why the expiry lands in 1970

`_compute_stamp` derives the expiry as `timestamp + ttl`. It rejects a `None`
timestamp and a non-positive expiry, but **not** a negative timestamp that still
yields a positive expiry:

```
timestamp = -1  (Kafka NO_TIMESTAMP)
ttl       = 5s
expiry    = -1 + 5000 = 4999 ms after the epoch  ->  1 Jan 1970
```

That stamp is recorded in `_pending_stamps`. On an unflipped store the write
itself does not trigger the flip — `advance_high_water` ignores negative
timestamps, so such a write cannot anchor a migration and is deliberately
excluded from setting `_batch_has_ttl_writes`. But the **pending stamp is still
recorded**, so if any *other* write in the same batch flips the store,
`_restamp_default_cf_cache_for_flip` applies it and the record is written with
the 1970 expiry.

Sites (all four have the same shape):

- `quixstreams/state/rocksdb/transaction.py:295` (`set`) and `:377` (`set_bytes`)
- `quixstreams/state/memory/partition.py:1718` and `:1786`

```python
if timestamp is not None and timestamp >= 0:
    self._batch_has_ttl_writes = True          # flip trigger correctly withheld
self._track_batch_ttl_ms(ttl)
key_serialized = self._serialize_key(key, prefix=prefix)
self._pending_stamps[(prefix, key_serialized)] = stamp   # <-- recorded anyway
```

## Impact

Silent data loss for that key: it is written, immediately considered expired
against any real event-time high-water, and swept on the next flush. Nothing is
logged.

Reachability is narrow — it needs an un-timestamped record *and* a timestamped
one in the same checkpoint, on a store that flips in that same batch — but
`NO_TIMESTAMP` arrives from any producer that does not set a timestamp, and the
flip happens exactly once per store, during migration.

Note this is a **regression introduced while fixing a crash** on the same path:
previously such a write triggered the flip with no high-water and raised
`IncompatibleStateStoreError` out of the flush (a crash loop). Withholding the
flip trigger removed the crash but left the stamp behind, which moved the failure
from "loud crash" to "silent delete".

## Suggested fix

Record the pending stamp under the same condition as the flip trigger:

```python
key_serialized = self._serialize_key(key, prefix=prefix)
if timestamp is not None and timestamp >= 0:
    self._batch_has_ttl_writes = True
    self._pending_stamps[(prefix, key_serialized)] = stamp
else:
    logger.warning(
        "state.set(..., ttl=...) carried timestamp=%s, which cannot anchor an "
        "expiry; storing the record without one.", timestamp,
    )
self._track_batch_ttl_ms(ttl)
```

This relies on an existing contract — keys present in `_update_cache` but absent
from `_pending_stamps` are treated as `SENTINEL_NEVER` at flip time — so the
record becomes never-expiring instead of born-expired. That trades silent
deletion for silent non-expiry, which is the safer direction for a dedup store,
and the warning makes it visible.

`_track_batch_ttl_ms(ttl)` should stay outside the guard: it feeds the implicit
legacy TTL used when a populated store flips without `legacy_records_ttl`, and
the *duration* is still valid information even when the anchor is not.

Two alternatives considered and rejected:

- **Reject the write** with `ValueError` — breaks
  `TestNegativeEventTimeGuard::test_negative_first_timestamp_does_not_crash_high_water`,
  which deliberately requires such writes to be accepted.
- **Anchor on the batch's high-water** (`high_water + ttl`) — gives the record a
  real TTL, but invents an event-time anchor for a record that has none and makes
  the result depend on write order within the batch.

## Test gap

`tests/test_quixstreams/test_state/test_rocksdb/test_no_timestamp_flip.py::
test_a_later_good_timestamp_in_the_same_batch_still_flips` already constructs
exactly this batch and asserts only that the store flipped. It should also assert
the un-anchorable key's stamp is `SENTINEL_NEVER` rather than a 1970 expiry —
that assertion is red on current code.

nirajkumarbarot:
Closes #1141

## Summary

A `state.set(..., ttl=...)` write whose record carries Kafka's `NO_TIMESTAMP` (`-1`) could be stored with an expiry in 1970, making the record already expired the moment it lands — deleted silently on the very next TTL sweep.

## Root cause

In `set()` / `set_bytes()` (both `RocksDBPartitionTransaction` and `MemoryPartitionTransaction`), an un-timestamped write correctly avoided setting `_batch_has_ttl_writes` (so it couldn't wrongly trigger the store flip), but its computed stamp - a bogus `NO_TIMESTAMP (-1) + ttl` expiry was still recorded in `_pending_stamps` unconditionally:

```python
if timestamp is not None and timestamp >= 0:
    self._batch_has_ttl_writes = True   # flip trigger correctly withheld
self._pending_stamps[(prefix, key_serialized)] = stamp   # <-- recorded anyway
```

If any *other* write in the same batch flipped the store, the stale stamp was applied to this record via `_restamp_default_cf_cache_for_flip`, producing a born-expired record with nothing logged.

## Fix

Record the pending stamp under the same condition as the flip trigger:

```python
key_serialized = self._serialize_key(key, prefix=prefix)
if timestamp is not None and timestamp >= 0:
    self._batch_has_ttl_writes = True
    self._pending_stamps[(prefix, key_serialized)] = stamp
else:
    logger.warning(
        "state.set(..., ttl=...) carried timestamp=%s, which cannot "
        "anchor an expiry; storing the record without one.", timestamp,
    )
self._track_batch_ttl_ms(ttl)
```

This relies on the existing contract that a key present in the update cache but absent from `_pending_stamps` falls back to `SENTINEL_NEVER` (never-expiring) at flip time - trading silent deletion for silent non-expiry, which is the safer default for a dedup store, with a warning now logged for visibility.

`_track_batch_ttl_ms(ttl)` stays outside the guard: it feeds the implicit legacy TTL used when a populated store flips without `legacy_records_ttl`, and the *duration* is still valid information even when the write's timestamp can't anchor an expiry.

Applied identically to all four call sites (RocksDB `set`/`set_bytes` and Memory `set`/`set_bytes` on the unflipped-partition path).

## Alternatives considered (and why rejected — per the issue)

- **Reject the write with `ValueError`** — breaks
  `TestNegativeEventTimeGuard::test_negative_first_timestamp_does_not_crash_high_water`,  which deliberately requires such writes to be accepted.

- **Anchor on the batch's high-water** (`high_water + ttl`) - gives the   record a real TTL but invents an event-time anchor for a record that has none, and makes the result depend on write order within the batch.

## Test gap closed

`test_a_later_good_timestamp_in_the_same_batch_still_flips` already constructed the exact batch that reproduces this bug but only asserted the store flipped. Added an assertion that the un-anchorable key's stamp decodes to `SENTINEL_NEVER` rather than a 1970 expiry - verified red on the original code, green after the fix.

## Testing

- `pytest tests/test_quixstreams/test_state/test_rocksdb/test_no_timestamp_flip.py -v` - all 4 pass
- Full `tests/test_quixstreams/test_state/` suite — 717 passed (remaining failures are pre-existing Docker-fixture errors unrelated to this change)
- Confirmed `TestNegativeEventTimeGuard::test_negative_first_timestamp_does_not_crash_high_water` still passes (the write-acceptance contract is preserved)

### DIFF
--- a/quixstreams/state/memory/partition.py
+++ b/quixstreams/state/memory/partition.py
@@ -1756,11 +1756,17 @@ class MemoryPartitionTransaction(PartitionTransaction[bytes, Any]):
             # write would mark the batch as flip-triggering while leaving the
             # high-water unset, and the backfill would then raise
             # IncompatibleStateStoreError out of the flush.
+            key_serialized = self._serialize_key(key, prefix=prefix)
             if timestamp is not None and timestamp >= 0:
                 self._batch_has_ttl_writes = True
+                self._pending_stamps[(prefix, key_serialized)] = stamp
+            else:
+                logger.warning(
+                    "state.set(..., ttl=...) carried timestamp=%s, which cannot "
+                    "anchor an expiry; storing the record without one.",
+                    timestamp,
+                )
             self._track_batch_ttl_ms(ttl)
-            key_serialized = self._serialize_key(key, prefix=prefix)
-            self._pending_stamps[(prefix, key_serialized)] = stamp
             if timestamp is not None:
                 self._partition.advance_high_water(timestamp)
         elif self._pending_stamps:
@@ -1824,11 +1830,17 @@ class MemoryPartitionTransaction(PartitionTransaction[bytes, Any]):
             # write would mark the batch as flip-triggering while leaving the
             # high-water unset, and the backfill would then raise
             # IncompatibleStateStoreError out of the flush.
+            key_serialized = self._serialize_key(key, prefix=prefix)
             if timestamp is not None and timestamp >= 0:
                 self._batch_has_ttl_writes = True
+                self._pending_stamps[(prefix, key_serialized)] = stamp
+            else:
+                logger.warning(
+                    "state.set(..., ttl=...) carried timestamp=%s, which cannot "
+                    "anchor an expiry; storing the record without one.",
+                    timestamp,
+                )
             self._track_batch_ttl_ms(ttl)
-            key_serialized = self._serialize_key(key, prefix=prefix)
-            self._pending_stamps[(prefix, key_serialized)] = stamp
             if timestamp is not None:
                 self._partition.advance_high_water(timestamp)
         elif self._pending_stamps:
--- a/quixstreams/state/rocksdb/transaction.py
+++ b/quixstreams/state/rocksdb/transaction.py
@@ -292,11 +292,17 @@ class RocksDBPartitionTransaction(PartitionTransaction[bytes, Any]):
             # itself still succeeds and is still re-stamped if some OTHER write
             # in this batch carries a real timestamp and flips the store; only
             # the flip trigger is withheld.
+            key_serialized = self._serialize_key(key, prefix=prefix)
             if timestamp is not None and timestamp >= 0:
                 self._batch_has_ttl_writes = True
+                self._pending_stamps[(prefix, key_serialized)] = stamp
+            else:
+                logger.warning(
+                    "state.set(..., ttl=...) carried timestamp=%s, which cannot "
+                    "anchor an expiry; storing the record without one.",
+                    timestamp,
+                )
             self._track_batch_ttl_ms(ttl)
-            key_serialized = self._serialize_key(key, prefix=prefix)
-            self._pending_stamps[(prefix, key_serialized)] = stamp
             # Advance the high-water on TTL writes only — we don't want a
             # plain ``state.set(k, v)`` on an unflipped store to start
             # writing the TTL high-water marker (legacy stores must stay
@@ -374,11 +380,17 @@ class RocksDBPartitionTransaction(PartitionTransaction[bytes, Any]):
             # itself still succeeds and is still re-stamped if some OTHER write
             # in this batch carries a real timestamp and flips the store; only
             # the flip trigger is withheld.
+            key_serialized = self._serialize_key(key, prefix=prefix)
             if timestamp is not None and timestamp >= 0:
                 self._batch_has_ttl_writes = True
+                self._pending_stamps[(prefix, key_serialized)] = stamp
+            else:
+                logger.warning(
+                    "state.set(..., ttl=...) carried timestamp=%s, which cannot "
+                    "anchor an expiry; storing the record without one.",
+                    timestamp,
+                )
             self._track_batch_ttl_ms(ttl)
-            key_serialized = self._serialize_key(key, prefix=prefix)
-            self._pending_stamps[(prefix, key_serialized)] = stamp
             if timestamp is not None:
                 self._partition.advance_high_water(timestamp)
         elif self._pending_stamps:
--- a/tests/test_quixstreams/test_state/test_rocksdb/test_no_timestamp_flip.py
+++ b/tests/test_quixstreams/test_state/test_rocksdb/test_no_timestamp_flip.py
@@ -25,6 +25,7 @@ import pytest
 
 from quixstreams.state.exceptions import IncompatibleStateStoreError
 from quixstreams.state.rocksdb import RocksDBOptions
+from quixstreams.state.rocksdb.ttl_codec import SENTINEL_NEVER, decode_ttl_value
 
 # Kafka's NO_TIMESTAMP. Messages produced without a timestamp arrive with this.
 NO_TIMESTAMP = -1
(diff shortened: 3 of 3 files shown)


nirajkumarbarot:
Hi @SteveRosam, @LudvikQuix 
Whenever you get a chance, could you please share your review comment or an approval on this PR please. Sincerely appreciate your time.

LudvikQuix:
Two gaps, both reproducible on this branch.

**1. Flipped stores aren't covered.** `_pending_stamps` only exists on the unflipped path. Once a store takes its first TTL write it flips, and from then on `set()` returns early into `_set_default_cf_stamped` (`rocksdb/transaction.py:478` / `:522`, `memory/partition.py:1901` / `:1937`), which stamps inline and never goes near `_pending_stamps`. So the same NO_TIMESTAMP write still lands in 1970:

```
default CF:  b'pfx|"kbad"'   -> 0000000000001387 = 4999   (1 Jan 1970)
             b'pfx|"kgood"'  -> 000000e8d9cb6c00          (correct)
read back:   kgood -> 'vgood'      kbad -> None
```

Since a store is flipped from its second TTL batch onward, that's the common case. The same fallback needs applying in those four functions, where the stamp is written directly.

**2. The warning isn't always true.** It says "storing the record without one", but if the key already has a pending stamp from an earlier write in the same batch, you leave it in place and the record inherits that expiry. Writing `kbad` with a real timestamp and then again with NO_TIMESTAMP gives `payload=b'"v2"'` stamped `1000000005000` — a finite expiry the last write never asked for. The `elif self._pending_stamps:` branch just below yours already pops for exactly this reason; the same `pop` in your `else` closes it.

Two tests for your `test_no_timestamp_flip.py`, red on this branch and green once both gaps are closed:

```python

    def test_no_timestamp_write_on_an_already_flipped_store(
        self, store_partition_factory
    ):
        """
        The same un-anchorable write, but on a store that has already flipped --
        which is every TTL store from its second batch onward.

        ``set()`` returns early into ``_set_default_cf_stamped()``, which stamps
        inline and never touches ``_pending_stamps``, so the pending-stamp guard
        cannot reach it.
        """
        partition = store_partition_factory(name="db", options=RocksDBOptions())
        try:
            # An ordinary ttl= write flips the store and sets the frontier.
            with partition.begin() as tx:
                tx.set(
                    key="kgood",
                    value="vgood",
                    prefix=b"pfx",
                    timestamp=1_000_000_000_000,
                    ttl=timedelta(days=1),
                )
            assert partition.uses_ttl_stamps

            # A later batch carries the un-anchorable write.
            with partition.begin() as tx:
                tx.set(
                    key="kbad",
                    value="vbad",
                    prefix=b"pfx",
                    timestamp=NO_TIMESTAMP,
                    ttl=timedelta(seconds=5),
                )

            cf = partition.get_or_create_column_family("default")
            with partition.begin() as tx:
                raw_key = tx._serialize_key("kbad", prefix=b"pfx")
                # Control: the read path itself works.
                assert tx.get(key="kgood", prefix=b"pfx") == "vgood"
                readback = tx.get(key="kbad", prefix=b"pfx")
            stamp, payload = decode_ttl_value(cf[raw_key])

            assert stamp == SENTINEL_NEVER, (
                "an un-anchorable NO_TIMESTAMP ttl= write must be stored as "
                "never-expiring on a flipped store too, not with a bogus expiry "
                f"derived from NO_TIMESTAMP + ttl (got {stamp})"
            )
            assert payload == b'"vbad"'
            assert readback == "vbad", "the record is unreadable the moment it lands"
        finally:
            partition.close()

    def test_no_timestamp_write_clears_an_earlier_stamp_for_the_same_key(
        self, store_partition_factory
    ):
        """
        Last-write-wins. If a key is written twice in one unflipped batch -- first
        with a real timestamp and a ttl, then with NO_TIMESTAMP -- the second write
        must not inherit the first write's expiry.

        The warning this fix logs says "storing the record without one", so the
        record must end up never-expiring.
        """
        partition = store_partition_factory(name="db", options=RocksDBOptions())
        try:
            with partition.begin() as tx:
                tx.set(
                    key="kbad",
                    value="v1",
                    prefix=b"pfx",
                    timestamp=1_000_000_000_000,
                    ttl=timedelta(seconds=5),
                )
                tx.set(
                    key="kbad",
                    value="v2",
                    prefix=b"pfx",
                    timestamp=NO_TIMESTAMP,
                    ttl=timedelta(seconds=5),
                )
                # A sibling write flips the store.
                tx.set(
                    key="kgood",
                    value="vgood",
                    prefix=b"pfx",
                    timestamp=1_000_000_000_000,
                    ttl=timedelta(days=1),
                )

            cf = partition.get_or_create_column_family("default")
            with partition.begin() as tx:
                raw_key = tx._serialize_key("kbad", prefix=b"pfx")
            stamp, payload = decode_ttl_value(cf[raw_key])

            assert payload == b'"v2"'
            assert stamp == SENTINEL_NEVER, (
                "the last write for this key could not anchor an expiry, so the "
                "record must be never-expiring; it kept the earlier write's "
                f"expiry instead (got {stamp})"
            )
        finally:
            partition.close()
```

**On duplicate work:** #1163 is the same fix from another author, on a branch with the same name. It's further along — newer base (CI on 3.11/3.13 rather than 3.9), a dedicated test file, and it already clears the stale stamp, so it doesn't have gap 2. It does have gap 1, and it dropped the `logger.warning` that #1141 asked for and you kept. Probably worth the maintainers picking one to carry rather than both of you fixing gap 1 separately.

LudvikQuix (changes requested):
see comment above please

PR #1147, opened 2026-09-11
feat: Column stats, virtual partitions, sort column and row-group sizing for QuixTSDataLakeSink

tomas-quix:
## Summary

Adds four query-performance features to `QuixTSDataLakeSink` so the lakehouse can prune files and navigate partitions without reading Parquet footers.

- **Per-file column statistics.** Min/max/null_count zone maps are computed from the in-memory Arrow batch for every numeric and timestamp column and sent to the REST Catalog as `column_stats` on each add-files entry. Numeric bounds are widened to the nearest representable float so pruning can only lose precision, never skip a matching file. `stats_columns` restricts the tracked set on wide tables.
- **Virtual partition columns.** A `~` prefix in `hive_columns` (e.g. `["year", "month", "~driver"]`) marks a column that appears in the partition tree and is filterable, but is not written as a `key=value/` folder and does not split files. The column stays in the Parquet data. Each data file gets a `.vidx/` sidecar Parquet next to it holding the distinct virtual tuples plus the file's physical partition values, so navigation can glob the sidecars with DuckDB. Sidecar writes are best-effort and never block data.
- **`sort_column`.** Recorded on the table as `properties.sort_column` (with `timestamp_column` as the fallback) so compaction can write ordered, skippable files.
- **Fixed Parquet row-group size.** New `row_group_rows` parameter (default 1,000,000, matching the lakehouse compaction) caps rows per row group so large flushes are split and small flushes remain one group.

Also fixes a spurious "Partition strategy mismatch" on restart against an existing virtual-partitioned table: catalog validation now compares against the full tree order rather than physical columns only.

## Catalog contract changes

- `PUT table`: `partition_spec` is sent up front when virtual columns are configured, plus `properties.virtual_partitions`, `properties.sort_column`, `properties.timestamp_column`.
- `add-files`: entries may carry `column_stats`. The key is omitted when empty so older catalogs ignore it. Nothing virtual is sent to the catalog any more; the virtual index lives only in blob sidecars.

### DIFF
--- a/docs/connectors/sinks/quix-ts-datalake-sink.md
+++ b/docs/connectors/sinks/quix-ts-datalake-sink.md
@@ -59,6 +59,23 @@ Records must be dictionaries. If your values are not dicts, convert them before 
 
 Blob credentials are read automatically from the `Quix__BlobStorage__Connection__Json` environment variable when running on Quix Cloud; for local runs, the filesystem is inferred from the `quixportal` configuration.
 
+### Parquet row groups
+
+Each file is written with row groups of about `row_group_mb` (default 32 MB) — the same
+target the lakehouse compaction uses. A reader needs roughly four storage range requests
+per row group, so many small groups make every query pay a round-trip storm on a
+high-latency storage path, while one enormous group forfeits intra-file skipping and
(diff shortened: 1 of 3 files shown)


LudvikQuix (approved):
review ok

